PR #19, opened 2019-09-21
make heliospring the default fc

bubi-007:
rewrite the emuboost

### DIFF
--- a/src/main/flight/pid.c
+++ b/src/main/flight/pid.c
@@ -136,10 +136,21 @@ typedef float (*pidControllerFn)(const pidProfile_t *pidProfile, int axis, float
 float featheredPids(const pidProfile_t *pidProfile, int axis, float errorRate, float dynCi, float currentPidSetpoint);
 float classicPids(const pidProfile_t *pidProfile, int axis, float errorRate, float dynCi, float currentPidSetpoint);
 
+float errorBoostLimit = 1.0f;
+float errorMultiplier = 1.0f;
+
+float featheredPIDboost( float errorRate)
+{
+	float boostedErrorFactor = (errorRate * errorRate) * errorMultiplier;
+	boostedErrorFactor += 1.0f;
+	boostedErrorFactor = MIN(boostedErrorFactor, errorBoostLimit);
+
+	return boostedErrorFactor;
+}
 
 void resetPidProfile(pidProfile_t *pidProfile)
 {
-	r_weight = 0.67;
+	r_weight = 0.33;
 
     RESET_CONFIG(pidProfile_t, pidProfile,
         .pid = {
@@ -164,7 +175,7 @@ void resetPidProfile(pidProfile_t *pidProfile)
         .feedForwardTransition = 0,
         .feathered_pids = USE_FEATHERED_PIDS,
         .i_decay = 4,
-        .r_weight = 67,
+        .r_weight = 33,
         .errorBoost = 15,
         .errorBoostLimit = 20,
         .yawRateAccelLimit = 100,
@@ -272,6 +283,8 @@ void pidInitFilters(const pidProfile_t *pidProfile)
     const uint32_t pidFrequencyNyquist = pidFrequency / 2; // No rounding needed
 
     r_weight = (float) pidProfile->r_weight / 100.0f;
+	errorMultiplier = (float)pidProfile->errorBoost * (float)pidProfile->errorBoost * 1e-9f;
+	errorBoostLimit = (float)pidProfile->errorBoostLimit / 100.0f + 1.0f;
 
     uint16_t dTermNotchHz;
     if (pidProfile->dterm_notch_hz <= pidFrequencyNyquist) {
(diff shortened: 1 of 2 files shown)


src/main/flight/pid.c
```diff
@@ -136,10 +136,21 @@ typedef float (*pidControllerFn)(const pidProfile_t *pidProfile, int axis, float
 float featheredPids(const pidProfile_t *pidProfile, int axis, float errorRate, float dynCi, float currentPidSetpoint);
 float classicPids(const pidProfile_t *pidProfile, int axis, float errorRate, float dynCi, float currentPidSetpoint);
 
+float errorBoostLimit = 1.0f;
+float errorMultiplier = 1.0f;
+
+float featheredPIDboost( float errorRate)
```
Quick-Flash: Is this mean to only work in feathered pids? Some people may still want to try this in conjunction with feedforward and will need to use classic pids.

src/main/flight/pid.c
```diff
@@ -136,10 +136,21 @@ typedef float (*pidControllerFn)(const pidProfile_t *pidProfile, int axis, float
...
+	boostedErrorFactor += 1.0f;
+	boostedErrorFactor = MIN(boostedErrorFactor, errorBoostLimit);
+
+	return boostedErrorFactor;
+}
 
 void resetPidProfile(pidProfile_t *pidProfile)
 {
-	r_weight = 0.67;
+	r_weight = 0.33;
```
Quick-Flash: Why change the r_weight?

src/main/flight/pid.c
```diff
@@ -136,10 +136,21 @@ typedef float (*pidControllerFn)(const pidProfile_t *pidProfile, int axis, float
 float featheredPids(const pidProfile_t *pidProfile, int axis, float errorRate, float dynCi, float currentPidSetpoint);
 float classicPids(const pidProfile_t *pidProfile, int axis, float errorRate, float dynCi, float currentPidSetpoint);
 
+float errorBoostLimit = 1.0f;
+float errorMultiplier = 1.0f;
+
+float featheredPIDboost( float errorRate)
+{
+	float boostedErrorFactor = (errorRate * errorRate) * errorMultiplier;
```
Quick-Flash: How does this deal with the error being a negative value? May it is, just verifying.

src/main/flight/pid.c
```diff
@@ -865,28 +878,13 @@ FAST_CODE float featheredPids(const pidProfile_t *pidProfile, int axis, float er
 {
     (void)(currentPidSetpoint);
 
-//Add EmuBoost to the code (non linear boost to errorRate)
-float errorMultiplier = (pidProfile->errorBoost * pidProfile->errorBoost / 1000000) * 0.003;
```
Quick-Flash: This line was added to the code to allow for a nicer scaling of the emuboost values. It looks like nothing has been added to account for this. You think that we could keep this same scaling with the update?

src/main/flight/pid.c
```diff
@@ -272,6 +283,8 @@ void pidInitFilters(const pidProfile_t *pidProfile)
     const uint32_t pidFrequencyNyquist = pidFrequency / 2; // No rounding needed
 
     r_weight = (float) pidProfile->r_weight / 100.0f;
+	errorMultiplier = (float)pidProfile->errorBoost * (float)pidProfile->errorBoost * 1e-9f;
```
Quick-Flash: If we change this line to be 

((float)pidProfile->errorBoost * (float)pidProfile->errorBoost * 1e-9f) * 0.003

then the emuboost will retain the same scaling as before.

src/main/flight/pid.c
```diff
@@ -865,28 +878,13 @@ FAST_CODE float featheredPids(const pidProfile_t *pidProfile, int axis, float er
...
-  {
-    boostedErrorRate = (0 - errorRate * errorRate) * errorMultiplier;
-  } else {
-    boostedErrorRate = errorRate * pidProfile->errorBoostLimit / 100;
-  }
-}
-
     // -----calculate P component
-    pidData[axis].P = pidCoefficient[axis].Kp * (boostedErrorRate + errorRate);
+    pidData[axis].P = pidCoefficient[axis].Kp * errorRate;
```
Quick-Flash: Does this change the error to always be a boost? I think that it is still good to have some linearity to the error. I could just be overlooking something here though.

Quick-Flash:
I've taken a look at the changes you have made and they definitely do simplify the code. I have just a few questions and I'm sure that they can be easily resolved.
Thanks for the work Marinus!

src/main/flight/pid.c
bubi-007: works for any pid controller

src/main/flight/pid.c
bubi-007: that line can be removed.
33 is the default imu-f value.

src/main/flight/pid.c
bubi-007: multiplication.
the boostfactor is positive and the error is positive or negative.
the sign is not a problem.

src/main/flight/pid.c
bubi-007: do the math.

src/main/flight/pid.c
bubi-007: div 1000000 * .003 is how much?
1e-9 i think - six plus three would be nine.

src/main/flight/pid.c
Quick-Flash: ok I didn't realize that.
I'll get some people testing their rweight at .33 and see what they think of it.
I think we should wait for now before doing that change though, but that's good info to have.

src/main/flight/pid.c
Quick-Flash: I just did that math and got 3e-9
so if we change that 1e-9 everything should be right.

src/main/flight/pid.c
Quick-Flash: aww my bad i didn't notice that at first. I shouldn't look at code this early in the morning.

src/main/flight/pid.c
```diff
@@ -1102,6 +1086,10 @@ void pidController(const pidProfile_t *pidProfile, const rollAndPitchTrims_t *an
 
         // -----calculate error rate
         errorRate = currentPidSetpoint - gyro.gyroADCf[axis]; // r - y
+        {
+        	const float nonLinearBoost = featheredPIDboost(errorRate);
+        	errorRate *= nonLinearBoost;
```
Quick-Flash: Does this multiply the error by our boost? Would we not want to use a += here instead?

Aside from this now that you have commented everything is looking good :)

PR #20, opened 2019-09-22
add new racerstar and airbot targets (F7 variant)

bubi-007:
adding the AIRBOTF7 and AIRF7 targets.


Quick-Flash:
Thanks Marinus!

PR #40, opened 2019-11-02
Doud fpv changes to the kalman filtering

Quick-Flash:
Much needed changes to the kalman filter :)

thanks doud


DoudFPV:
Update Kalman filter with the IMU-F Kalman code.
The main different is covariance if compute between axes in IMU-F

PR #57, opened 2019-11-24
Battery fix

loutwice:


Quick-Flash:
already merged into the current code. Thanks loutwice!

PR #65, opened 2020-01-03
increment msp api for v0.2.12+ [skip ci]

nerdCopter:
increment msp api to 1.44 for v0.2.12+


loutwice:
Why increment api ? As there is no new changes ?

nerdCopter:
to remove dynamic LPF's from gui, it needed to match some api number

PR #78, opened 2020-02-01
fix firmware to show v2 font crosshair [skip ci]

nerdCopter:
requires v2 fonts uploaded to FC


nerdCopter:
closed/unmerged, want to check other options first.

PR #80, opened 2020-02-03
Use v2font msp 1.44+ [skip ci]

nerdCopter:
Requires v2 fonts uploaded to flight controller.  GUI will trigger v2 font usage with msp 1.44+
Not only fixes crosshair, but also includes some GPS, temperature, and imperial/metric character fixes.


nerdCopter:
https://youtu.be/4Cow4hZbRoU

Quick-Flash:
Looks good thanks for doing this :)

PR #93, opened 2020-03-04
feat: vbat compensation

tylercorleone:
The purpose of **battery compensation** is to provide constant flight performance during flight time, by making it independent from battery voltage. It aims to provide a constant throttle-thrust and error-correction relationship through the battery life.

A configurable voltage reference (vref) is the target of all the throttle and PIDs compensations, that is the battery voltage that the user wants to "feel" during the flight.

The compensation is obtained by applying a factor calculated based on vref, the instantaneous battery voltage (filtered) and the level of correction the user wants to apply, independent for throttle and PIDs.

The vref is the pivot of the compensation, that is, if the battery voltage equals vref, there is no compensation (factor = 1). When the battery is above vref the compensation will be negative (factor < 1) or positive otherwise (factor > 1).

Considering a linear relationship between the voltage and RPM, a quadratic relationship between RPM and thrust and an almost linear relationship between the throttle and voltage, the correction will be inversely proportional to the square of the relationship between vref and the battery voltage:

factor = (vref / vbat)^2

To take in account the non-linear relationship between throttle and voltage, and to let the user choose the level of correction he wants to be applied, the factor is then scaled between 0 and the previously calculated one.

The user can also choose the type of correction to apply, if none, limit only (factor < 1 only), boost only or both.


etracer65:
Strange that you would neglect to mention that your PR to Betaflight was rejected because the implementation is fundamentally flawed. But keep pitching it at other projects I guess and maybe somebody will take the bait.

tylercorleone:
> Strange that you would neglect to mention that your PR to Betaflight was rejected because the implementation is fundamentally flawed. But keep pitching it at other projects I guess and maybe somebody will take the bait.

How do you know I didn't collaborated with people before opening this PR? They **did** see the PR I opened on Betaflight. I don't understand why I should mention here that it wasn't accepted.
That said, what do you expected to obtain with you comment?

Quick-Flash:
Ahh @etracer65 it seems the BF devs do keep some sort of eye on EmuFlight. Yes I have been privately discussing this with @tylercorleone and we have been discussing this in greater detail. I'm sorry to say it, but I don't see how this is fundamentally flawed. Yes I'm not 100% sure that it will work as we want. There is a possibility that your throttle will get weird and messed up as you raise the throttle, your voltage lowers, this raises your throttle, which causes your voltage to drop creating a feedback loop. 

However, before I am going to kill an idea that looks like it has some potential, we are going to test it and fly it and figure out if it works or not. I personally don't see how you can judge the validity of every idea just on paper. Sometimes you need to fly the code, test it and see exactly what is going on. Thats exactly what I intend to do with this PR. If this does end up creating some nasty feedback loops, we have ideas that can help remove that feedback loop. But rather than go through the effort of fixing what might just be a problem we are going to test it and move from there. 

Thanks for pushing this @tylercorleone !!!

etracer65:
@Quick-Flash  LOL! No. Only reason I even found this was because @tylercorleone gloated about it on his now-closed Betaflight PR.

So whatever. This exact thing has been tried the exact same way multiple times and if you search you'll find other PR's with the same fundamental problem. So there's nothing new here.

As to why it's flawed, it's been discussed over and over. But in a nutshell it leads to a runaway feedback loop. Battery voltage drops, throttle is increased. Increased throttle causes battery voltage to drop more due to the IR of the battery as sag. The voltage dropping causes the throttle to increase more, causing more sag. And so on.

So good luck. If this was as easy as presented in this PR it would have already been done a dozen times and long ago.

tylercorleone:
"And so on" is misleading.

At 3.2 volts, with a reference of 3.7 volts, the throttle value is increased by 25%. No more than that.

This means that already at 80% the throttle is clipped.

As we said we want just try it. I tried and it works well (at leas the way I fly...). The user on BF, the same...

Remember also that the current Battery compensation is already increasing PID values (here I'm talking about the Battery PID compensation, of course), but does it in a "wrong" way. IAlso it's not able to **reduce** the PIDs.

I would be interest on seing those other PR you are talking about.

tylercorleone:
Moreover, battery compensation can be configured in "limit only" mode, that is the throttle can only be decreased in order to match a low/medium voltage battery when the battery is full.

In such a way PIDs and throttle can be tuned on a reduced range of variability, and below the voltage reference the quadcopter behaves just like no compensation, without any gap respect the remaining range, since the compensation ends at 1.0 when the battery is at Vref in a continuous way.

etracer65:
Ok, whatever. Since you think you have it all figured out (but still apparently don't grasp the fundamental concepts) I'll just delete my comments and unsubscribe from this so I don't get bothered by further noise.

tylercorleone:
etracer65 > Ok, whatever. Since you think you have it all figured out (but still apparently don't grasp the fundamental concepts) I'll just delete my comments and unsubscribe from this so I don't get bothered by further noise.

before you delete your comments... :-D

gretel:
@tylercorleone thanks for the contribution. looking forward to give this experiement a try!

Quick-Flash:
Just looking at this again, it appears that vbat pid compensation is now tied into the throttle compensation as well, I think it would be worth splitting that up

Quick-Flash:
@tylercorleone one more note i'm currently doing some testing. Vbat by default needs to be increased. could you update this PR to allow you to set vbat pid compensation and vbat throttle compensation separately so you can use either or both if wanted. Also the vbat lpf period should be increased to 30 or 40 along with this PR. That will filter the voltage a lot more than it currently is, which helps remove the issue of running into nasty feedback loops, which I ran into while testing. If you neglect to filter the voltage enough you definitely do run into times where the feedback loops causes throttle oscilations, but that is fixed as you increase the voltage filtering. Thanks for this once again, i'm loving it so far :)

tylercorleone:
Yes, throttle compensation and PID compensation are tied in a some manner, but you can set the compensation level for each of them independently, so that if you want to disable compensation on throttle, just set **vbat_comp_throttle_level** to 0, to disable or tune the compensation on PIDs change **vbat_comp_pid_level**.

They are using the same **vbat_comp_type** indeed.

Glad to hear it is working for you too!

Maybe we have to use a different battery voltage source, since the filtered one is used to display the battery voltage on the OSD if I'm not wrong. Yes, we can try to filter it a little more and see what happens, otherwise we will work to the other filtering we talked about.

Thank you

phobos-:
Sorry to chip in here, I think current filtering that is used should stay as is, and maybe you could have a separate voltage value that would be filtered properly (based on 10s time period or more) just for this feature and pid vbat compensation?
I think that is a better solution as it wont mess up other things i.e. the value displayed in the OSD.

Could you help me understand how the boost works? I don't think it's possible to go over 100% throttle, is it just going to be clipped?

tylercorleone:
@phobos- I agree. We can try to increase the filtering to the value suggested by @Quick-Flash and if it messes the OSD value use a different value.

Yes, the throttle is clipped at 1.0. About this I thought that the clipping should be moved after the "canonical" throttle limit, so we don't lose any information, e.g. a throttle greater than 1.0 that would be brought back by the canonical limit without the need of clipping it.

Quick-Flash:
@phobos- @tylercorleone In my opinion the vbat should be filtered more regardless of whether or not you are using this feature. I'm open to hear what others think, but I really enjoy seeing a vbat that doesn't just bounce around like mad to any small thing that I do. I really don't think it is the instantaneous voltage that matters, but rather the more real estimate of where you battery is at. So regardless of this feature being implemented I do think that the vbat should be filtered more heavily by default, but I'm open to being told why I am wrong about this.

gretel:
merged manually, closing.

PR #99, opened 2020-03-09
blackbox: reflect reality in system information header

gretel:
record actual flight controller version used - same as bf4.2


gretel:
https://github.com/betaflight/betaflight/blob/master/src/main/blackbox/blackbox.c#L1237

gretel:
will break cause cleanflight is not at version 3 :)
https://github.com/emuflight/EmuFlight/blob/master/src/main/blackbox/blackbox.c#L1429

PR #100, opened 2020-03-09
update toolchain to 2019-q4-release

gretel:
same as bf4.2 - confirmed flying on `MATEKF411RX`


gretel:
just do `make arm_sdk_install` and you should be fine.

PR #101, opened 2020-03-09
sorta kinda semantic versioned hex filenames

gretel:
have the git rev and build number (on travis) in the filename like `EmuFlight_FURYF4OSD_0.2.39_c85cef5ee_local.hex`


gretel:
@Quick-Flash we should keep branch names short and precise and ask people to do so.

PR #103, opened 2020-03-10
fix filterGetNotchQ

DzikuVx:
For unknown reason `filterGetNotchQ` was returning a constant Q factor that:
* happily ignored notch filter cutoff frequency
* that was leading to a situation when cutoff (-dB) was approx. center/2 (always)
* phase delay was not constant but was growing  with center frequency as attenuated band was wider and wider


Quick-Flash:
Good catch. I'm not sure why this was ever implemented.

gretel:
@DzikuVx thanks!

PR #107, opened 2020-03-10
remove lagging average filter from code

Quick-Flash:
Remove the lagging average filter which is no longer used.

This filter has not been used for quite a long time as far as I am aware. No need to keep bloat in the code.


gretel:
debloat!

PR #108, opened 2020-03-10
Dzikuvx pid sanity

DzikuVx:
* Fix intendation
* extract Iterm and I-decay computation to separate function to improve readability
* from non-zero check as it will be probably now faster thanks to removal on branching. Also the chance of error being actually zero is very small
* some other minor sanity checks


src/main/flight/pid.c
```diff
@@ -86,127 +86,88 @@ static FAST_RAM_ZERO_INIT bool antiGravityEnabled;
...
+                 },
+
+                 .dFilter = {
+                     [PID_ROLL] = {2, 100, 250, 50},  // wc, dtermlpf, dtermlpf2, smartSmoothing
+                     [PID_PITCH] = {2, 100, 250, 50}, // wc, dtermlpf, dtermlpf2, smartSmoothing
+                     [PID_YAW] = {0, 100, 250, 0},    // wc, dtermlpf, dtermlpf2, smartSmoothing
+                 },
+
+                 .pidSumLimit = PIDSUM_LIMIT_MAX, .pidSumLimitYaw = PIDSUM_LIMIT_YAW, .dterm_filter_type = FILTER_PT1, .itermWindupPointPercent = 50, .pidAtMinThrottle = PID_STABILISATION_ON, .levelAngleLimit = 45, .angleExpo = 10, .feedForwardTransition = 0, .setPointPTransition[ROLL] = 110, .setPointPTransition[PITCH] = 110, .setPointPTransition[YAW] = 130, .setPointITransition[ROLL] = 75, .setPointITransition[PITCH] = 75, .setPointITransition[YAW] = 45, .setPointDTransition[ROLL] = 110, .setPointDTransition[PITCH] = 110, .setPointDTransition[YAW] = 130, .feathered_pids = 100, .i_decay = 4, .errorBoost = 15, .errorBoostYaw = 40, .errorBoostLimit = 20, .errorBoostLimitYaw = 40, .yawRateAccelLimit = 0, .rateAccelLimit = 0, .itermThrottleThreshold = 350, .itermAcceleratorGain = 1000,
```
gretel: looks less readable to my eyes

src/main/flight/pid.c
```diff
@@ -86,127 +86,88 @@ static FAST_RAM_ZERO_INIT bool antiGravityEnabled;
...
+                 .crash_time = 500,                        // ms
+                 .crash_delay = 0,                         // ms
+                 .crash_recovery_angle = 10,               // degrees
+                 .crash_recovery_rate = 100,               // degrees/second
+                 .crash_dthreshold = 50,                   // degrees/second/second
+                 .crash_gthreshold = 400,                  // degrees/second
+                 .crash_setpoint_threshold = 350,          // degrees/second
+                 .crash_recovery = PID_CRASH_RECOVERY_OFF, // off by default
+                 .horizon_tilt_effect = 130, .nfe_racermode = false, .cinematic_setpoint = false, .crash_limit_yaw = 200, .itermLimit = 400, .throttle_boost = 5, .throttle_boost_cutoff = 15, .iterm_rotation = true, .iterm_relax = ITERM_RELAX_OFF, .iterm_relax_cutoff = 11, .iterm_relax_type = ITERM_RELAX_GYRO, .abs_control_gain = 0, .abs_control_limit = 90, .abs_control_error_limit = 20, .antiGravityMode = ANTI_GRAVITY_SMOOTH, .use_integrated_yaw = false, .integrated_yaw_relax = 200, .motor_output_limit = 100, .auto_profile_cell_count = AUTO_PROFILE_CELL_COUNT_STAY, .horizonTransition = 0, );
```
gretel: same here

src/main/flight/pid.c
Quick-Flash: I don't think you intentionally did this. Please just put a new line between each of these.

src/main/flight/pid.c
```diff
@@ -86,127 +86,88 @@ static FAST_RAM_ZERO_INIT bool antiGravityEnabled;
...
 void pgResetFn_pidProfiles(pidProfile_t *pidProfiles)
 {
-  for (int i = 0; i < PID_PROFILE_COUNT; i++) {
+    for (int i = 0; i < PID_PROFILE_COUNT; i++)
+    {
         resetPidProfile(&pidProfiles[i]);
     }
 }
 
-extern void init_pwm_filter(float dT);
```
Quick-Flash: why did you remove this?

src/main/flight/pid.c
```diff
@@ -641,21 +630,23 @@ static void detectAndSetCrashRecovery(
...
-            if (getMotorMixRange() >= 1.0f && !inCrashRecoveryMode
-                && ABS(delta) > crashDtermThreshold
-                && ABS(errorRate) > crashGyroThreshold
-                && ABS(getSetpointRate(axis)) < crashSetpointThreshold) {
+    if ((crash_recovery || FLIGHT_MODE(GPS_RESCUE_MODE)) && !gyroOverflowDetected())
+    {
+        if (ARMING_FLAG(ARMED))
+        {
+            if (getMotorMixRange() >= 1.0f && !inCrashRecoveryMode && ABS(delta) > crashDtermThreshold && ABS(errorRate) > crashGyroThreshold && ABS(getSetpointRate(axis)) < crashSetpointThreshold)
+            {
```
Quick-Flash: not sure if it is worth using this format for long if statements. Simply as the rest of the code uses the previous formatting.

src/main/flight/pid.c
```diff
@@ -742,25 +743,64 @@ static FAST_RAM_ZERO_INIT float setPointDAttenuation[XYZ_AXIS_COUNT];
 static FAST_RAM_ZERO_INIT timeUs_t crashDetectedAtUs;
 static FAST_RAM_ZERO_INIT timeUs_t previousTimeUs;
 
-    void pidController(const pidProfile_t *pidProfile, const rollAndPitchTrims_t *angleTrim, timeUs_t currentTimeUs)
+static void processIterm(
+    uint8_t axis,
+    float previousIterm,
```
Quick-Flash: Would you think of changing the name of this? Since iterm is cumulative seeing accumulatedIterm, or some other term would feel more intuitive. When I first read the code I was thinking previous iterm was the previous change to iterm rather than the old value for iterm.

src/main/flight/pid.c
DzikuVx: interesting what formatter does sometimes to the code... fixed

src/main/flight/pid.c
DzikuVx: string `init_pwm_filter` is not used anywhere else in the code base.

src/main/flight/pid.c
DzikuVx: TBH I do not like autoformatters just because for different things. I wanted to fix formatting manually but gave up after 20 minutes. There were so many formatting problems that fixing this manually was over any reasonable time budget.
Good catch, well split this too

src/main/flight/pid.c
DzikuVx: why not some people say that the hardest part of programming in how to name things

DzikuVx: how about currentIntegral?

Quick-Flash: I like it.

PR #109, opened 2020-03-11
Remove jitter in the accelerometer

Quick-Flash:
Put a quad on a level surface and look at how the quad will slightly jitter. Changing the dcm defaults will fix this and make the angle code work better. If you want to verify this try it for yourself.


gretel:
what does `dcm` mean?

Quick-Flash:
Direction cosine matrix is what I think it stands for. More references found here.
https://en.wikipedia.org/wiki/Rotation_formalisms_in_three_dimensions#Formalism_alternatives

PR #110, opened 2020-03-11
First cut of Matrix Filter

DzikuVx:
it's a not direct port of INAV Matrix Filter that is based on gyro FFT analysis.
* not tested
* not optimized
* so far compiles


gretel:
@DzikuVx thank you! any chance you can give this a test flight? regards

Quick-Flash:
Been doing some testing on a toothpick myself and it seems to work very well. Defaults though I think need a little changing. A higher notch q of say 250 or so and a notch quality of 10 or so I think would make for better defaults.

Quick-Flash:
Noise plots looking very good and you can't see the motor harmonics at all.

image
image

src/main/sensors/gyro.c
```diff
@@ -288,7 +290,8 @@ PG_RESET_TEMPLATE(gyroConfig_t, gyroConfig,
     .yaw_spin_recovery = true,
     .yaw_spin_threshold = 1950,
     .dyn_notch_quality = 70,
-    .dyn_notch_width_percent = 50,
+    .dyn_notch_q_factor = 150,
```
Quick-Flash: I think that setting this higher for default would work better. Same as lowering the dyn_notch_quality

DzikuVx: Probably around values about 200 would be a nice default. FFT anaylysis is however not very precise sometimes and always delays, so high Q values as with RPM filter are not possible since filter would always miss the peak.
On the topic of quality, I'd rather drop this setting completely but not as part of this PR tho

DzikuVx:
Due to COVID, lockdown and quarantine I'm not able to do any testing besides static on a workbench. @Quick-Flash thanks for testing

gretel:
https://www.youtube.com/watch?v=KhbTAS8mTPI

loutwice:
@Quick-Flash @DzikuVx dyn_notch_min_hz is not used in the code for calculation or i miss something ?

PR #111, opened 2020-03-11
build faster, better, wider on travis

gretel:
our friends mik and ez did some work to improve parallel builds on travis in https://github.com/betaflight/betaflight/pull/9493
instead of having a total of 5 groups 3 are enough. as travis seems to limit parallelism on the amount of total jobs running this seems to allow two concurrent builds.
besides, i did read on and mess with `ccache` settings. roughly estimating looking at the travis build times this seems to speed up consequent builds (sharing caches).


gretel:
well it's taking same time per job but does allow two at the same time so its 100% more faster in theory ;)

PR #114, opened 2020-03-13
Gyro.c put fabsf where needed

Quick-Flash:
Fix a bug which could potentially stop gyro overflow from ever being detected. The new compiler showed me that gyro.c had a few bugs, one of them was using abs for floats. Floats should use fabsf, fabs is for doubles, and abs should be for integers. This may fix an issue I have heard of where some users get a yaw spin to the moon. I'm hoping this helps.

### DIFF
--- a/src/main/sensors/gyro.c
+++ b/src/main/sensors/gyro.c
@@ -1044,9 +1044,9 @@ FAST_CODE_NOINLINE int32_t gyroSlewLimiter(gyroSensor_t *gyroSensor, int axis)
 static FAST_CODE_NOINLINE void handleOverflow(gyroSensor_t *gyroSensor, timeUs_t currentTimeUs)
 {
     const float gyroOverflowResetRate = GYRO_OVERFLOW_RESET_THRESHOLD * gyroSensor->gyroDev.scale;
-    if ((abs(gyroSensor->gyroDev.gyroADCf[X]) < gyroOverflowResetRate)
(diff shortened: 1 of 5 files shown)


src/main/sensors/gyro.c
```diff
@@ -1044,9 +1044,9 @@ FAST_CODE_NOINLINE int32_t gyroSlewLimiter(gyroSensor_t *gyroSensor, int axis)
 static FAST_CODE_NOINLINE void handleOverflow(gyroSensor_t *gyroSensor, timeUs_t currentTimeUs)
 {
     const float gyroOverflowResetRate = GYRO_OVERFLOW_RESET_THRESHOLD * gyroSensor->gyroDev.scale;
-    if ((abs(gyroSensor->gyroDev.gyroADCf[X]) < gyroOverflowResetRate)
-          && (abs(gyroSensor->gyroDev.gyroADCf[Y]) < gyroOverflowResetRate)
-          && (abs(gyroSensor->gyroDev.gyroADCf[Z]) < gyroOverflowResetRate)) {
+    if ((fabsf(gyroSensor->gyroDev.gyroADCf[X]) < gyroOverflowResetRate)
```
gretel: 👀

Quick-Flash: Does this mean there is a problem?

gretel: no i would have requested a change. i was just looking like UH!

PR #116, opened 2020-03-13
Remove GA Filter

Quick-Flash:
The GA or gyro average filter was adopted from the old RaceFlightOne code and has not shown itself to be useful. So this filter is being removed in light of simplification and progression. Should make configurator changes easier as well.

### DIFF
--- a/src/main/interface/msp.c
+++ b/src/main/interface/msp.c
@@ -1330,9 +1330,9 @@ bool mspProcessOutCommand(uint8_t cmdMSP, sbuf_t *dst)
         sbufWriteU8(dst, currentPidProfile->dFilter[ROLL].Wc);
         sbufWriteU8(dst, currentPidProfile->dFilter[PITCH].Wc);
         sbufWriteU8(dst, currentPidProfile->dFilter[YAW].Wc);
-        sbufWriteU8(dst, gyroConfig()->averagedGyro[ROLL]);
-        sbufWriteU8(dst, gyroConfig()->averagedGyro[PITCH]);
-        sbufWriteU8(dst, gyroConfig()->averagedGyro[YAW]);
+        sbufWriteU8(dst, 0);
+        sbufWriteU8(dst, 0);
+        sbufWriteU8(dst, 0);
 
 
         break;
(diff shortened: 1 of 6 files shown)


src/main/interface/msp.c
```diff
@@ -1330,9 +1330,9 @@ bool mspProcessOutCommand(uint8_t cmdMSP, sbuf_t *dst)
         sbufWriteU8(dst, currentPidProfile->dFilter[ROLL].Wc);
         sbufWriteU8(dst, currentPidProfile->dFilter[PITCH].Wc);
         sbufWriteU8(dst, currentPidProfile->dFilter[YAW].Wc);
-        sbufWriteU8(dst, gyroConfig()->averagedGyro[ROLL]);
-        sbufWriteU8(dst, gyroConfig()->averagedGyro[PITCH]);
-        sbufWriteU8(dst, gyroConfig()->averagedGyro[YAW]);
+        sbufWriteU8(dst, 0);
```
gretel: either i dont understand how the buffering works or one statement of the same should be enough

src/main/interface/msp.c
```diff
@@ -1968,9 +1968,9 @@ mspResult_e mspProcessInCommand(uint8_t cmdMSP, sbuf_t *src)
             currentPidProfile->dFilter[ROLL].Wc = sbufReadU8(src);
             currentPidProfile->dFilter[PITCH].Wc = sbufReadU8(src);
             currentPidProfile->dFilter[YAW].Wc = sbufReadU8(src);
-            gyroConfigMutable()->averagedGyro[ROLL] = sbufReadU8(src);
-            gyroConfigMutable()->averagedGyro[PITCH] = sbufReadU8(src);
-            gyroConfigMutable()->averagedGyro[YAW] = sbufReadU8(src);
+            sbufReadU8(src);
```
gretel: same here?

src/main/interface/msp.c
```diff
@@ -1996,10 +1996,10 @@ mspResult_e mspProcessInCommand(uint8_t cmdMSP, sbuf_t *src)
         gyroConfigMutable()->imuf_yaw_q = sbufReadU16(src);
         gyroConfigMutable()->imuf_w = sbufReadU16(src);
 #ifdef USE_GYRO_IMUF9001
-        gyroConfigMutable()->imuf_roll_lpf_cutoff_hz = sbufReadU16(src);
-        gyroConfigMutable()->imuf_pitch_lpf_cutoff_hz = sbufReadU16(src);
-        gyroConfigMutable()->imuf_yaw_lpf_cutoff_hz = sbufReadU16(src);
-        gyroConfigMutable()->imuf_acc_lpf_cutoff_hz = sbufReadU16(src);
+        sbufReadU16(src);
```
gretel: same here?

Quick-Flash: basically with how the MSP works if you remove something you need to keep a "blank spot" for it to just read past otherwise it will get out of sync.

gretel: ok i see. cant this be optimized somehow, like having one expression instead of three each?

Quick-Flash: I don't think it easily can, refer to this part of the current msp.c
https://github.com/emuflight/EmuFlight/blob/556f8c6ce906e2a469142d629434203fcebf8bef/src/main/interface/msp.c#L778

Quick-Flash: https://github.com/emuflight/EmuFlight/blob/556f8c6ce906e2a469142d629434203fcebf8bef/src/main/interface/msp.c#L1155

PR #120, opened 2020-03-15
make turtle mode power output configurable

Quick-Flash:
Before the old turtle mode code would only allow for about 70% power when attempting a pitch or a roll movement, but 100% power when spinning a corner motor or doing a yaw. Now its been changed so that motors can only have at max the power level you configure. Default value is 70 to help with safety and motor burning. These values can be increased for whoops and decreased for some craft that are quite powerful. Should make turtle mode overall a little bit better. A wiki write up should accompany this update as well.

Was looking at this code when trying to figure out how I wanted to approach cinematic setpoint and it bugged me enough I dropped this quick fix.

### DIFF
--- a/src/main/flight/mixer.c
+++ b/src/main/flight/mixer.c
@@ -77,6 +77,7 @@ PG_RESET_TEMPLATE(mixerConfig_t, mixerConfig,
     .mixerMode = TARGET_DEFAULT_MIXER,
     .yaw_motors_reversed = false,
     .crashflip_motor_percent = 0,
+    .crashflip_power_percent = 70,
 );
 
 PG_REGISTER_WITH_RESET_FN(motorConfig_t, motorConfig, PG_MOTOR_CONFIG, 1);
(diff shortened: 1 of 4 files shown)


Quick-Flash:
Did not mean to add assignee's meant to add reviewers, my bad I'll try to be more careful. The project was already added so I just finished it.

I personally view this as a bug fix as well as adding a feature. Whoops that cannot currently turtle like the tinyhawk should be able to turtle now. Before it was just thought that they couldn't turtle, but seeing as how before it was just that they weren't capable of giving full power during a turtle mode situation it makes this feel like a bug as turtle mode does not work as expected. Now turtle mode should work as expected and this should allow for a safer way to turtle on rigs that you don't want full power to turtle for. For a huge majority of users they will just continue to turtle as they have in the past.

gretel:
my tinyhawk has tutled ever i put the right props on it. dont see why this is important now. thanks for the PR anyway.

iggut:
Some motor/prop combos around mid battery suffer from inability to turtle.
I think It should be an option to be configurable

PR #123, opened 2020-03-17
Drop gyroanalyse quality factor

DzikuVx:
Gyroanalyse quality factor (?) seemed to work with a very peculiar way:
* it was setting a BPF and center on `fftSamplingRate / 4`
* default Q factor of `0.7` was setting cutoff at approx half of this frequency. Lowering to somewhere recommended `0.1` was setting to approx. 10% of it

As a result:
* gyroanalyse was preferring frequencies closer to `fftSamplingRate / 4` more than any other frequencies
* peak detection was not constant among frequency range
* tuning of this factor was bound to fftFrequency that is not visible to the user and not even directly dependent on looptime frequency
* in general, hard to use, complicated, with questionable gain and apparently useless on a longer run

This PR is an addition to #110


Quick-Flash:
i'll be out to test this soon, no bbl as i'm flying a toothpick but i'll see if this helps with my high throttle oscillations.

Quick-Flash:
just got back from flying definitely feels better to me and it does seem I have less high throttle oscillations. I would like more confirmation though.

gretel:
ugh did i mess the merge up?
```
./src/main/sensors/gyro.c: In function 'gyroInitSensor':
./src/main/sensors/gyro.c:534:41: error: 'gyroConfig_t' {aka 'const struct gyroConfig_s'} has no member named 'dyn_notch_q_factor'
  534 |     gyroSensor->dynNotchQ = gyroConfig()->dyn_notch_q_factor / 100.0f;
      |                                         ^~
```

DzikuVx:
there is a conflict between this and #110 
I'll resolve this later

DzikuVx:
conflicts resolved, builds should proceed now

PR #127, opened 2020-03-18
add workflow for github actions to build

gretel:
works in parallel to travis and is worth exploring. adds a additional status check, too.


Quick-Flash:
what exactly does this do?

gretel:
it adds a workflow for github actions to build. you can read more about github actions at https://github.com/features/actions

gretel:
thank you.

PR #130, opened 2020-03-20
Revert "Fix a bug in kalman.c?"

Quick-Flash:
Reverts emuflight/EmuFlight#129

Just now had a chance to try this and it does not work at all. I thought it would fix it but it does not :( the kalman is more broken now. needs to be reverted asap


gretel:
dang

PR #141, opened 2020-03-25
Fix kalman.c

Quick-Flash:
Some issues with this kalman filter were brought to my attention. I have since been playing with it and getting it fixed. So far this works for me. I'll be having more people test it and see what they think as well, but it should make this filter more similar to the filter in the helio fc which should be for the better me thinks. Pushing this again as the last one had stick pids in it as well, which I did not want.

### DIFF
--- a/src/main/common/kalman.c
+++ b/src/main/common/kalman.c
@@ -116,6 +116,7 @@ FAST_CODE float kalman_process(kalman_t* kalmanState, float input, float target)
   //update last state
   kalmanState->lastX = kalmanState->x;
 
+  if (kalmanState->lastX != 0.0f) {
   // calculate the error
   	float errorMultiplier = fabsf(target - kalmanState->x) * kalmanState->s;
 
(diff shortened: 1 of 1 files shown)


src/main/common/kalman.c
```diff
@@ -116,17 +116,15 @@ FAST_CODE float kalman_process(kalman_t* kalmanState, float input, float target)
   //update last state
   kalmanState->lastX = kalmanState->x;
 
+  if (kalmanState->lastX != 0.0f) {
```
gretel: guess it's fine but try to have the most common case first - for performance reasons :)

Quick-Flash: Yea that isn't a very common case at all, but how would you suggest I change it, and still maintain its functionality?

gretel: try to use `==` whenever possible and match the most probable case first. guess it's fine for now

Quick-Flash: I mean in this code this is added just to prevent a divide by 0 error. This should be good to merge.

nerdCopter:
the internets show "compliancy" to be:
```C
  if (kalmanState->lastX == 0.0f)  {
    /* Handle error */
  } else {
    /* Do the work */
  }
```
but also has more complex solutions like error trapping: http://www.keil.com/support/man/docs/ARMCC/armcc_chr1359124235078.htm
Apparently C++ has better div by zero handling over C (NaN/Infinity)

src/main/common/kalman.c
gretel: i wasnt saying it is not. at the time of my previous writing the status checks were still running. you should be open to suggestions though and not just pushing for approval.

gretel:
dang i clicked the wrong button. hang on travis will just taker another day or so to build :(

PR #144, opened 2020-03-31
do not merge- just travis testing internet solution

nerdCopter:


Quick-Flash:
it appears this has worked... would you like it merged now?

nerdCopter:
no, i rather restart older build.  unsure if this is 100% proper

nerdCopter:
PR #145 worked without this.  therefore i think problem was travis-ci servers.

Quick-Flash:
well i've tried rebuilding my stick pids branch a handful of times to no avail...

gretel:
this wont fix because i deleted the caches manually.
pretty sure what needs to be done is this https://unix.stackexchange.com/questions/45583/argument-list-too-long-how-do-i-deal-with-it-without-changing-my-command
but i dont see any related change in this PR?
pleasse re-submit something related to the actual issue while not observing countereffects, only.

PR #145, opened 2020-03-31
fix double throttle in OSD

nerdCopter:
Kaio found issue & best solution


Quick-Flash:
this is required because we updated our osd right?

nerdCopter:
required because butterflight had two symbols and i have no clue why.  this just prints 1 throttle symbol followed by a blank character.  it keeps the alignment that Kaio had on his screenshot is discord.

PR #149, opened 2020-04-03
Smart smoothing fix

Quick-Flash:
after adding a debug mode to smart smoothing I found that it did not work as expected. So I've gone and updated it. It now works a lot better. Added a change to the defaults as well now because of this. This is why logging can be so important :)

### DIFF
--- a/src/main/flight/pid.c
+++ b/src/main/flight/pid.c
@@ -1073,7 +1073,7 @@ void pidController(const pidProfile_t *pidProfile, const rollAndPitchTrims_t *an
 
             if (smart_dterm_smoothing[axis] > 0)
             {
-                dDeltaMultiplier = constrainf(fabsf(dDelta + previousdDelta[axis]) / (2 * smart_dterm_smoothing[axis]), 0.2f, 1.0f);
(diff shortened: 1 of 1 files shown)


nerdCopter:
does this change the range from 0-25?  if yes, need to adjust firmware variable range and i'll need to modify GUI.

Quick-Flash:
> does this change the range from 0-25? if yes, need to adjust firmware variable range and i'll need to modify GUI.

no. i'm using 35 and am quite happy with the results. However, if we want to get this more solid before a merge I could test more to determine what a better maximum value would be, as i seriously doubt anyone should need to use a value of 250 (or even would be able to get that to fly well). I did change this code to limit the max that it can smooth as well. the most it will decrease dterm is to 25% where as before it could decrease it to nothing. Since changing this I found that I was having issues with that being to much and the quad bobbing around in level flight, but that change seemed to completely take care of that.

Quick-Flash:
This works very well now that i've tweaked it a bit. I really do like how I can get this to fly. Should be good to merge any time.

nerdCopter:
will merge as per discussion with QF.

PR #153, opened 2020-04-05
thrust linearization

tylercorleone:
Resuming a PR I screwed-up with (git rebasing gone wrong).

This PR is intended to port thrust linearization to EmuFlight. Something like betaflight/betaflight#7304, using a different, simpler, formula to compensate thrust non-linearity.

Thrust/RPM relationship is similar to a quadratic curve, but not exactly equal. A configurable mix with a linear relationship (no compensation) is used to find the exact compensation for a given quad.

Efficient quads (motor, propellers etc) will have a thrust_linearization_level close to 100, whereas a less efficient one will have a lower level, something like 65.

throttle_linearization is a boolean parameter that tells if the throttle feel has to be "linear" or de-compensated to match the old fashioned square root feel (default value).

I found the linear throttle feel to be great, since it helps to maintain a mure constant fly height. But since the throttle hovering level will be a little higher (e.g. from 33 to 50), I suggest to use a custom three points curve on the transmitter: (-100, -100), (-33, -70), (100, 100), but I'm thinking to add this type of curve directly on EmuFlight, letting the user choose the type of curve to apply to the throttle.

I tested linearization in this way:
- set idle throttle to 0%
- put my quad uside-down on a balance in acro mode, using a support to hold it firmly
- zeroed the balance
- took note of the thrust (in grams) generated at various throttle level

         STOCK     LINEAR THRUST
20%       6                   16
40%     19                   30
60%     32                   45
80%     56                   61

### DIFF
--- a/src/main/cms/cms_menu_imu.c
+++ b/src/main/cms/cms_menu_imu.c
@@ -271,10 +271,11 @@ static OSD_Entry cmsx_menuRateProfileEntries[] =
     { "TPA RATE D",  OME_FLOAT,  NULL, &(OSD_FLOAT_t) { &rateProfile.dynThrD,              0,  250,  1, 10}, 0 },
     { "TPA BREAKPOINT",   OME_UINT16, NULL, &(OSD_UINT16_t){ &rateProfile.tpa_breakpoint,  1000, 2000, 10}, 0 },
 
-    { "VBAT COMP TYPE",  OME_TAB,   NULL, &(OSD_TAB_t)    { &rateProfile.vbat_comp_type, VBAT_COMP_TYPE_COUNT - 1, cms_throttleVbatCompTypeLabels}, 0 },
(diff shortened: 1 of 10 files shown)


gretel:


```diff

Clones removed
==============
+ src/main/cms/cms_menu_imu.c  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/emuflight/EmuFlight/pullRequest?prid=5308102&bid=17371816)

gretel:
@tylercorleone did some testing.. getting roll of death on hard throttle punches with this enabled. if you want to the binaries you can get the as "artifacts" from the https://github.com/emuflight/EmuFlight/actions. 

would you suggest to revert.. or do you have the time to squish the bug?

i also have dvr footage :)

tylercorleone:
> @tylercorleone did some testing.. getting roll of death on hard throttle punches with this enabled. if you want to test the binaries you can get them as "artifacts" from the https://github.com/emuflight/EmuFlight/actions.
> 
> would you suggest to revert.. or do you have the time to squish the bug?
> 
> i also have dvr footage :)

Hi gretel, I missed out this comment. Punches are something I can't test at all, closed in my apartment. Can I see the footage? Thank you!

gretel:
@tylercorleone thanks hang on i'm lost in footage gonna send it

PR #156, opened 2020-04-07
Update max7456_symbols.h

Kaiowarez:
Minor osd fix, just changing a null value with the value for blank symbol


gretel:
man i was just about to report this :)

PR #159, opened 2020-04-08
Vbat precision increment w resting voltage

tylercorleone:
This PR contains the cherry-pick of this PR: https://github.com/betaflight/betaflight/pull/7255/commits
The branch containing only the changes needed to increment the battery precision (the above commits and other changes needed because of the conflicts) can be found here: https://github.com/tylercorleone/EmuFlight/tree/vbat_precision_increment

Here is contained also the implementation of "resting battery voltage", that is an attempt to compensate the sag caused by the spinning motors on battery voltage, e.g. on a punch, that is the scope of this https://github.com/emuflight/EmuFlight/projects/7

The vbat_max_voltage_sag parameter represents the maximum voltage sag the motors are capable of, e.g. 0.4V -> vbat_max_voltage_sag = 40
It is better to mantain this value BELOW the actual value in order to avoid to over-compensate. A conservative value of 0.2V has been chosen has default value.

The sag is simply calculated as the product of the maximum voltage sag and the "percent" of the motor output, actually the average normalized motor output (without the offset needed by the ESC protocol used).

It is better to separate the two concepts?

I think that a lot of testing and changes at the configurator are needed!


tylercorleone:
Note this line of code in OSD_AVG_CELL_VOLTAGE:
const int cellV = getBatteryRestingVoltage(); // TODO: this is just temporary, for debug purpose. Revert me!

I'm using this to see the difference between the "resting voltage" and the canonical one

gretel:


```diff

Clones removed
==============
+ src/main/cms/cms_menu_imu.c  -1
+ src/main/cms/cms_menu_power.c  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/emuflight/EmuFlight/pullRequest?prid=5330128&bid=17445390)

Quick-Flash:
yeah I think seperating these two things would be good to do. I'd also really like to see the compensated voltage (resting voltage) as something you can choose to show in the osd. That might become just what I use.

gretel:
@tylercorleone on a 'erased all' flashed fc i'm getting `vbat_comp_ref = 84` .. which seems to be not right :) any idea what could be causing this? i'm not setting it at all and code wise it looks fine to me.

tylercorleone:
> @tylercorleone on a 'erased all' flashed fc i'm getting `vbat_comp_ref = 84` .. which seems to be not right :) any idea what could be causing this? i'm not setting it at all and code wise it looks fine to me.

This branch wasn't updated with the revert of the "thrust linearization" that we made on the master.
Now I updated it and just contains the vbat precision increment and the resting voltage stuff.

The vbat_precision increment is also on a dedicated branch: https://github.com/emuflight/EmuFlight/pull/161

Let me know if the problem is still there!

gretel:
@tylercorleone cool thanks going to merge that on my working copy ;)

PR #168, opened 2020-04-14
BETAFPVF411_V3

Kaiowarez:
add target betafpvf411 v3
similar to matekf411 but with BB enabled


mwegner:
Would love to see this pushed through!  (At least, I think this is the board I have on one of my BETAFPV whoops here)

Kaiowarez:
This board is finished would be included in the next release

gretel:
@Kaiowarez not sure if there will be any release.

PR #169, opened 2020-04-14
smoothing out motor mix

tylercorleone:
I smoothed-out the mix of throttle and motorMix (PID controller compensation).
The current code *chops* the throttle in the [-motorMixMin, 1 - motorMixMax] range, while we can obtain a smoother motor output (and hence thrust and torque) by gradually applying some transformation.

Here some images that compares the current mixer, the new proposal and a third one that could be used after we introduce thrust linearization.

On the X axis the throttle, on the Y axis the motorMixDelta, that is (motorMixMax - MotorMixMin) / 2 and on the Z axis the thrust or the torque generated by two opposite motors with a motorMix = 0.2.

airmode_off
airmode_on

The difference while flying is that the wobbles and the strange oscillations that sporadically occurs are no longer present.

We also have configurable authority on six points, that we could use for something else other that AirMode OFF/ON.

Link to the Octave project (should work with Matlab too): 
[https://drive.google.com/drive/folders/1zq5MPOINuYMCWhX1X9RylwGbz4AmqF1F?usp=sharing](url)

The *SMOOTH 2* mix I propose provides also a better throttle authority on sharp movements (motorMixDelta = 0.5), as you can see from the graphs. The thrust continues to depend on the throttle!


Quick-Flash:
This is really awesome. I've been thinking that airmode needed some improvements like this for a while, but didn't know how one would go about implementing it. I wonder how this improves the bounce that airmode can cause if you bump into the ground??? Maybe if we can get airmode good enough we can just make airmode always on? Could simplify the code a bit and get it to run a bit faster.

src/main/flight/mixer.c
```diff
@@ -814,6 +814,22 @@ float applyThrottleLimit(float throttle)
...
 }
 
+void applyAirMode(float *motorMix, float motorMixMax)
+{
+    float absThrottle = ABS(throttle);
+    float normalizationFactor = motorMixRange > 1.0f && hardwareMotorType != MOTOR_BRUSHED ? motorMixRange : 1.0f;
+    float motorMixDelta = 0.5f * motorMixRange;
+    for (int i = 0; i < motorCount; ++i) {
+        motorMix[i] += motorMixDelta - motorMixMax; // let's center the values exactly around the zero
+        float result = motorMix[i] + scaleRangef(absThrottle, 0.0f, 1.0f, motorMixDelta, - motorMixDelta); // this is the actual AirMode trick
```
Quick-Flash, line 824: I do have one question about scaleRangef, I've never used this function before. Do you happen to know the kind of cpu load that this adds vs scaling things yourself? I've never seen this function used until you started showing its potential to me, but I'm wondering if I've never encountered it because it uses more cpu then we would like. Just a thought and I figured you might know.

tylercorleone:
> This is really awesome. I've been thinking that airmode needed some improvements like this for a while, but didn't know how one would go about implementing it. I wonder how this improves the bounce that airmode can cause if you bump into the ground??? Maybe if we can get airmode good enough we can just make airmode always on? Could simplify the code a bit and get it to run a bit faster.

Unfortunately the bounce issue still remains, because at low throttle the AirMode shifts everything to the left (to avoid clipping) making the quad rise-up while trying to compensate. The same for the quad getting stuck on the wall when you crash. :(

But there is the possibility to apply only a given percent of airmode at low throttle.
For a 50% AirMode on low throttle I obtain this: 
air_mode_50_single_motors_thrust
air_mode_50_thrust_and_torque

src/main/flight/mixer.c
tylercorleone: Good question.
I don't know if we would obtain better performances by using a custom/in-line scaling, the compiler should optimize the formula the scalerangef uses (just a couple of additions and multiplications). Personally I prefer legibility when I'm not sure if there is actually a performance issue, but we could substitute it with a macro etc. if needed!

Quick-Flash:
> > This is really awesome. I've been thinking that airmode needed some improvements like this for a while, but didn't know how one would go about implementing it. I wonder how this improves the bounce that airmode can cause if you bump into the ground??? Maybe if we can get airmode good enough we can just make airmode always on? Could simplify the code a bit and get it to run a bit faster.
> 
> Unfortunately the bounce issue still remains, because at low throttle the AirMode shifts everything to the left (to avoid clipping) making the quad rise-up while trying to compensate. The same for the quad getting stuck on the wall when you crash. :(
> 
> But there is the possibility to apply only a given percent of airmode at low throttle.
> For a 50% AirMode on low throttle I obtain this:
> air_mode_50_single_motors_thrust
> air_mode_50_thrust_and_torque

Yes perhaps we can play with the airmode strength with throttle, maybe its worth something, maybe its not. Maybe it would be worth figuring out ways besides throttle to change the airmode strength. So airmode strength would be determined by both throttle and perhaps somethings else that helps prevent airmode from freaking out when you land or crash into something.

src/main/flight/mixer.c
```diff
@@ -814,6 +814,22 @@ float applyThrottleLimit(float throttle)
     return throttle;
 }
 
+void applyAirMode(float *motorMix, float motorMixMax)
+{
+    float absThrottle = ABS(throttle);
```
Quick-Flash, line 819: is it possible for throttle to have a negative value??? if not then this is not needed

tylercorleone: Can't be negative on 3D mode?

tylercorleone: As not said... I remove the ABS

Quick-Flash: I just looked and at line 652 of mixer.c constrains the throttle to be between 0 and 1.

gretel: 3d mode centers at 0.5 so it should be fine

gretel:
@tylercorleone nice graphs!

src/main/flight/mixer.c
```diff
@@ -910,7 +910,7 @@ uint16_t yawPidSumLimit = currentPidProfile->pidSumLimitYaw;
     motorMixRange = motorMixMax - motorMixMin;
 
     if (IS_RC_MODE_ACTIVE(BOXBLACKBOX)) {
-        applyAirMode(motorMix, motorMixMax);
+        applyAirMode(motorMix, motorMixMax, isAirmodeActive() ? 1.0f : 0.75f);
```
Quick-Flash, line 913: for testing purposes at least I would love to see us being able to change the active airmode percent. I know a lot of tiny whoop pilots who don't fly airmode because of some issues it gives when you bump into things. So for at least testing purposes I love the idea of being able to mess with this. :)

tylercorleone:
After the changes we discussed about I updated the PR description. I'm really satisfied of the result :)

tylercorleone:
> thanks mate. the naming of `useAirmode2_0` is a bit cumbersome in my opinion :) how about `useThrustLin` or sth? regards

Yes, but it's there just for testing purpose, like the blackbox dependency. If someone would like to try how does it feels can set that variable to true.

gretel:
@tylercorleone alright. "btw" maybe you can check on https://github.com/betaflight/betaflight/pull/9608, too 🥇

gretel:
@tylercorleone awesome thank you!

tylercorleone:
I merged here the predictive airmode, [idea](https://github.com/emuflight/EmuFlight/tree/QuickFlash-predictive-airmode) of @Quick-Flash.
The idea is that the quad's authority level can be also tied to the stick deflection.
I think we have to try different airmode_min_slow_authority, airmode_med_slow_authority and airmode_max_slow_authority levels.

"slow_authority" is referred to the percentage of authority that will be available (that is the percentage of the torque) when the aircraft is substantially stable.

Quick-Flash:
this really needs testing. i'm going to get this going

borisbstyle:
Interesting idea! I will have a better look soon.

Do you also have some flight logs with and without?

src/main/flight/pid.c
```diff
@@ -179,6 +179,13 @@ void resetPidProfile(pidProfile_t *pidProfile)
         .motor_output_limit = 100,
         .auto_profile_cell_count = AUTO_PROFILE_CELL_COUNT_STAY,
         .horizonTransition = 0,
+        .airmode_min_slow_authority = 50,
+        .airmode_min_fast_authority = 50,
+        .airmode_med_slow_authority = 100,
+        .airmode_med_fast_authority = 100,
+        .airmode_max_slow_authority = 100,
+        .airmode_max_fast_authority = 100,
```
borisbstyle, line 187: is there really a need to make them all configurable?

Quick-Flash:
> Interesting idea! I will have a better look soon.
> 
> Do you also have some flight logs with and without?

I've actually be working with Tyler on a slightly revised version of this (you can find the code in the different mixer branch). Yes I have logged, but I haven't saved any of them. My logging rig is also out of commission at the moment so it will be a while until I can test this code with logging again.

src/main/flight/pid.c
Quick-Flash: No but it was a nice way to get to test out code easily without needing to ever refresh. Updates to this code found in the "different mixer" branch of the emuflight code.

borisbstyle: Thanks for pointing that one out.

borisbstyle:
> > Interesting idea! I will have a better look soon.
> > Do you also have some flight logs with and without?
> 
> I've actually be working with Tyler on a slightly revised version of this (you can find the code in the different mixer branch). Yes I have logged, but I haven't saved any of them. My logging rig is also out of commission at the moment so it will be a while until I can test this code with logging again.

Ok. I guess this is the new code. Would be easier for reviewing though PR
https://github.com/emuflight/EmuFlight/blob/fa8e3e648e35969f5e4fb1aab99cde01a8aa9e0f/src/main/flight/mixer.c#L818-L858

nerdCopter:
@borisbstyle "Compare" ? may help
https://github.com/emuflight/EmuFlight/compare/different-mixer
image

borisbstyle:
> @borisbstyle "Compare" ? may help
> https://github.com/emuflight/EmuFlight/compare/different-mixer
> image

It it is pretty in sync in master, than it works ;D

Quick-Flash:
> > I've actually be working with Tyler on a slightly revised version of this (you can find the code in the different mixer branch). Yes I have logged, but I haven't saved any of them. My logging rig is also out of commission at the moment so it will be a while until I can test this code with logging again.
> 
> Ok. I guess this is the new code. Would be easier for reviewing though PR
> https://github.com/emuflight/EmuFlight/blob/fa8e3e648e35969f5e4fb1aab99cde01a8aa9e0f/src/main/flight/mixer.c#L818-L858

Not sure its quite ready to be made as a PR just yet. Give a little more time to testing and possible code tweaks and then I think it will be. Going to make a video describing all the changes here real soon

Quick-Flash:
@borisbstyle heres a video describing some of the new code in case that interests you. 
https://www.youtube.com/watch?v=7lQjsbGN8TY

borisbstyle:
@tylercorleone Why are your graphs only going up to 0.5 mix delta? Mix delta between motors can go up to 1.

@Quick-Flash Thanks I just watched the vid.

tylercorleone:
Continuing here https://github.com/emuflight/EmuFlight/pull/398

tylercorleone:
> @tylercorleone Why are your graphs only going up to 0.5 mix delta? Mix delta between motors can go up to 1.
> Edit: nevermind. I see you divide motormix by 2. I don't know why, but ok it is clear.
> 

Hi @borisbstyle, it comes from motorMixDelta = 0.5f * motorMixRange (https://github.com/emuflight/EmuFlight/pull/398/files#diff-c98e21d8f3cf216db2a51ba27291626147f6f48724944160963e60565ae349c3R807), but effectively a 0-1 range would be better. I'll fix it for the graphs of the new PR 🙂

PR #185, opened 2020-04-20
github actions compile fix

gretel:
kinda sorta self explanatory :)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,20 +1,63 @@
 name: Build EmuFlight
-on: [push, pull_request]
+
+on:
+  push:
+    tags:
+    - '*'
+  pull_request:
+    branches:
+    - '*'
+
 jobs:
   build:
     strategy:
       max-parallel: 1
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - uses: fiam/arm-none-eabi-gcc@v1
+
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Version String
+      id: get_version
+      run: echo "::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}"
+
+    - name: Setup Python
+      uses: actions/setup-python@v1
+
+    - name: Setup Toolchain
+      uses: fiam/arm-none-eabi-gcc@v1
       with:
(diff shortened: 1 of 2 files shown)


gretel:
the artifacts resulting on the action is what would be put on bintray. meaning this makes bintray superflous.

nerdCopter:
I agree.
Only issue is Test-Pilots have zero access to Drafts.  Thus the reason for Bintray.

gretel:
that is not a reason for bintray. `draft` is a attribute per deploy which can be toggled. it's just not such a good idea to do this for initial testing.
furthermore, what is keeping you bintray lovers from using the according github action?!
as i said before, it's super waste of resource to run travis which fails often just to put binaries up on bintray.
https://github.com/marketplace/actions/action-upload-bintray and others.

nerdCopter:
i don't love bintray. i hate bintray. 
"we" need an alternative to allow test-pilots access to builds we wish to share, yet keep relatively private and "unreleased".   but i digress.  i'm not wanting conflict of interest.

gretel:
i brought up an alternative months ago.
what are you talking about. closed this due to frustration

nerdCopter:
no need to close due frustration. i'm certainly not against anything. we are open to solutions.  if this is a proper solution then awesome and thank you.  let the team members review.

.github/workflows/build.yml
```diff
@@ -1,20 +1,63 @@
...
+
+    - name: Release Binaries
+      uses: softprops/action-gh-release@v1
+      if: startsWith(github.ref, 'refs/tags/')
+      with:
+        prerelease: true
+        draft: true
+        files: obj/*.hex
+      env:
+        GITHUB_TOKEN: ${{ secrets.release_token }}
```
gretel, line 63: due to current limitations of github the token is bound to my account. it should be able to issue token via a organizations in the future. any valid token by a orga member having admin priviliges will do.

gretel:
adjusted github settings accordingly so merges depend on github action builds, not travis. if this works not out as intented feel free to revert. but this should resume approval workflow now.

gretel:
somebody needs to issue a token asap and use it cause i can't sustain working on this.

nerdCopter:
Welcome to leave it un-maintained if you wish.
The existing token works as-is, but if you choose to delete it, then we will create a new one.  Thank you for all your efforts and work thus far.  It is well appreciated.

PR #186, opened 2020-04-20
Travis upload to Bintray only for master and [Bb]uild*

nerdCopter:
only deploy to Bintray for:
* `master`
* `build`*
* `Build`*
 where * represents wildcard filename

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,8 +75,12 @@ deploy:
     file: bintray-conf.json
     skip_cleanup: true # do not delete what has just been built
     on:
-      all_branches: true
       #tags: true
+      branches:
+        only:
+          -master
+          -build*
+          -Build*
 
 after_deploy:
   - wget https://raw.githubusercontent.com/DiscordHooks/travis-ci-discord-webhook/master/send.sh


nerdCopter:
@Quick-Flash you will still be able to name your branches whatever you want.  Then if you want it built, but yet not ready to merge into master, simply create create a branch prepended with `build`. (obviously branch-from-branch or merge-in)
e.g.
* `build/myFeature`
* `build_myFeature`
* `Build-this-branch`

I know we tried tags (a great idea), but were not getting what we wanted.  However, without tags it's too much.  this is a good compromise.

gretel:
why?

gretel:
https://github.com/emuflight/EmuFlight/pull/185

nerdCopter:
* alternative for limiting Bintray usage.
* keeps Bintray clutter-free & less confusing to Test-Pilots

gretel:
that is not a reason, but another workaround for an issue that needs to be fixed. why stick with bintray?
> I know we tried tags (a great idea), but were not getting what we wanted. However, without tags it's too much. this is a good compromise.
what is that supposed to mean?
why circumvent basic github workflow principles?
why make it even harder to contribute on this project?
why re-invent wheel?

nerdCopter:
* when i tried tags, it would not deploy as expected.
* @Quick-Flash and @loutwice didnt understand it either.  
* i learned about it and now understand.
* i'm completely open, willing, and happy to try again, but if it fails to deploy, then it's a problem.

separately, travis and github builds are slow.  I started using azure this weekend and found it's "fast" at about 45minutes to build EmuFlight with 3 parallel tasks.  I'm attempting to get a waiver of cost for azure from an M.S. insider. this might be a proper solution, only time will tell.  otherwise the free version is limited to 30 hours per month which seems too limited in some months like this past couple.

gretel:
tags and performance? that sounds very esoteric to me.
azure? why would one use that?
what is wrong with github actions? it has been compiling and uploading artifacts for like two months now.
travis is unreliable. the caches bloat. i'm the one cleaning that up ocassionally.
if you feel like ignoring my work just say so but please stop doing voodoo engineering.

nerdCopter:
i appreciate your work. thank you tremendously.  i dont want conflict, i'm open to all options. if my verbiage offends in any way, please describe how i might cease such.

PR #187, opened 2020-04-20
github compile fix aftermath

gretel:
adopt Makefile so it should work for both travis and github actions; enhanced semantics for the github actions build; fixed artifacts not getting stored on pull requests; minor formatting of Makefile


gretel:
@nerdCopter not required at all. it's easier to tell people where to download the artifacts instead of maintaining that broken mess.

PR #190, opened 2020-04-23
Fix bug with logging.

Quick-Flash:
previousPidSetpoint[axis] is used in blackbox logging to log your current setpoint. However when feedforward was removed this was also removed, and the unintended consequence was that logs of setpoint were no longer quite accurate. This should fix that and hopefully fix an issue with logging.

this is what we want to see 
image

instead this is what we see 
image

Rate Dynamics is not getting logged properly and is making the quad look sporadic in how it handles error. Looks actually very very bad. However if you look at pitch on the second image you can easily see that a lot of that pitch error comes from doing backflips while using rate dynamics.


Quick-Flash:
now i just need to add the rest of the settings to the logger

Quick-Flash:
[
image 

confirmed that it works. Just need to merge when possible.

PR #196, opened 2020-04-25
Fix turtle power

Quick-Flash:
since changing turtle mode it appears that turtle mode has less power than it did before. This fixes that issue and restores the power that turtle mode should have. (this bug was found as more whoop pilots had issues in turtle mode)


frizzle101101:
this kinda contradicts your previous goal https://github.com/emuflight/EmuFlight/commit/6ea711cc8113e640d8382bcc04a58574923dac4b#diff-c98e21d8f3cf216db2a51ba27291626147f6f48724944160963e60565ae349c3,
i like the idea of configurable power for turtle mode

PR #208, opened 2020-05-06
dterm kalman

Quick-Flash:
this uses a non covariance version of the kalman filter to add the kalman filter to the dterm. Over time I plan to test some slight variations to this filter to see how it can be improved to work specifically for the dterm and how its predictions work. Just meant for testing at the minute.


gretel:
please squash merges.

PR #213, opened 2020-05-08
20200507 kalman covar without cross influence

nerdCopter:
* rebuilt from another branch (`git reset --hard d2a99f9 ; git checkout -b 20200507_kalmanCovar_without_crossInfluence`).
* this is refactored Kalman code.
* initially it was the removal of cross-axis influence as recommended by Marinus
* it was refactored for per-axis calculations
* it was then refactored as a function


gretel:
please try to have commit messages giving context.

PR #242, opened 2020-05-20
Added additional battery voltage field to OSD

mbilozub:
Fixed Battery average cell voltage for the DJI OSD


loutwice:
thank you @mbilozub ! i have to validate it as weel will try soon !! thank you really much for your help

Quick-Flash:
if this is working without issues we should merge it to both master and the 0.3.x maintenance branch

nerdCopter:
whomever tests this should approve and merge both, ty.

loutwice:
@mbilozub tested and approved ! thanks for contributing !

PR #251, opened 2020-05-24
Motor output linearization

tylercorleone:
Thrust linearization.

The thrust_linearization_level represents how much the thust transfer function of the motor/propellers relative to the motor output is close to a linear (low efficiency) or quadratic. Default value is 0, that is no linearization. I expect values to be in the 70/90 range (70/75 for a tiny whoop, 85/90 for a bigger quad).
You can tune this value by enabling vbat_compensation and find the value that makes the throttle hovering level at full battery equal to the one at low battery.

You can choose if enable linearization on throttle too or not (default OFF, but I prefer it ON).


nerdCopter:


```diff

Clones removed
==============
+ src/main/cms/cms_menu_imu.c  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/emuflight/EmuFlight/pullRequest?prid=5576219&bid=18281313)

Quick-Flash:
how does this effect the current vbat throttle compensation?

tylercorleone:
> how does this effect the current vbat throttle compensation?

With this PR the battery compensation is done on motor output directly, so you don't have to tune PID comp and throttle compensation level but just choose the type of compensation (LIMIT, BOOST, BOTH) and voltage reference.

The old vbat_comp_throttle_level and vbat_comp_pid_level, don't exist anymore! Their scope was to match the motors response to voltage variation, that now (more correctly) is the scope of the thrust_linearization_level.

So I think it's a great simplification, but also an enhancement since it places responsibilities on the right place.

AndreySemjonov:
We need to give to test to our testers :)

PR #259, opened 2020-05-26
fix alti for dji and fc whitout baro

loutwice:
revert an old change that was incorrect in position.c 

and dji is getting from msp_altitude the getestimatedalt only for USE_BARO or USE_RANGEFINDER 
i added USE_GPS so it is enable for all fc without BARO using GPS

### DIFF
--- a/src/main/interface/msp.c
+++ b/src/main/interface/msp.c
@@ -931,7 +931,7 @@ bool mspProcessOutCommand(uint8_t cmdMSP, sbuf_t *dst)
         break;
 
     case MSP_ALTITUDE:
-#if defined(USE_BARO) || defined(USE_RANGEFINDER)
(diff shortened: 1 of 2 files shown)


src/main/interface/msp.c
```diff
@@ -931,7 +931,7 @@ bool mspProcessOutCommand(uint8_t cmdMSP, sbuf_t *dst)
         break;
 
     case MSP_ALTITUDE:
-#if defined(USE_BARO) || defined(USE_RANGEFINDER)
+#if defined(USE_BARO) || defined(USE_RANGEFINDER) || defined(USE_GPS)
```
gretel, line 934: take care https://github.com/emuflight/EmuFlight/pull/323 won't bring in similar.

PR #263, opened 2020-05-27
iTermRelax_cleanup

AndreySemjonov:
Cleaning up code iTermRelax related


Quick-Flash:
ahh i missed those, thanks for doing this

PR #267, opened 2020-05-31
Update README.md with DJI OSD Compatibility info

mricho:
Changing filtering settings in DJI OSD can cause unusual config changes in emuflight

Examples are here...
Google docs link of my testing with DJI OSD: https://docs.google.com/spreadsheets/d/1MlzeTgJpdGB2y0S2GhVLUJf304ZfkVu2dAei0YrJhqU/edit?usp=sharing

I think adding this wording would be appropriate


nerdCopter:
? maybe consider some verbiage that we have not focused on any DJI compatibility whatsoever as of yet.

PR #269, opened 2020-05-31
Correctly compute variance for Kalman

DzikuVx:
The Kalman implementation was not computing variance in a correct way.

This change uses the `average((x - mean)^2)` formula instead of `average(x^2) - average(mean^2)` (which is not a variance at all)

Bear in mind, that as a result, filter might behave in a different way


AndreySemjonov:
Thanks. Need to test this one for Helio kalman

AndreySemjonov:
So anyone tested it? I have only Helio boards right now so I can't )))

nerdCopter:
I plan to.
 I can build .hex's today.

nerdCopter:
I flew FoxeerF405, flew fine.  Due to strong winds, i could not say if better or not, but will presume accurate math is better :)  I did not change any LPF's as a result of new .hex.

@Quick-Flash @AndreySemjonov , we would need to back-port to IMUF binary.

src/main/common/kalman.c
```diff
@@ -49,22 +49,31 @@ void kalman_init(void)
...
-     kalmanFilterStateRate[axis].axisSumMean -=  kalmanFilterStateRate[axis].axisWindow[ kalmanFilterStateRate[axis].windex];
-     kalmanFilterStateRate[axis].axisSumVar =  kalmanFilterStateRate[axis].axisSumVar - ( kalmanFilterStateRate[axis].axisWindow[ kalmanFilterStateRate[axis].windex] *  kalmanFilterStateRate[axis].axisWindow[ kalmanFilterStateRate[axis].windex]);
-     kalmanFilterStateRate[axis].axisMean =  kalmanFilterStateRate[axis].axisSumMean *  kalmanFilterStateRate[axis].inverseN;
-     kalmanFilterStateRate[axis].axisVar =  fabsf(kalmanFilterStateRate[axis].axisSumVar *  kalmanFilterStateRate[axis].inverseN - ( kalmanFilterStateRate[axis].axisMean *  kalmanFilterStateRate[axis].axisMean));
-
-    kalmanFilterStateRate[axis].r = sqrtf(kalmanFilterStateRate[axis].axisVar) * VARIANCE_SCALE;
+  kalmanState->axisWindow[kalmanState->windex] = rate;
+
+  kalmanState->axisSumMean += kalmanState->axisWindow[kalmanState->windex];
+  float varianceElement = kalmanState->axisWindow[kalmanState->windex] - kalmanState->axisMean;
```
nerdCopter, line 57: @DzikuVx , sir, why doesn't the initial value of `axisMean` need to be calculated?  is it assumed to initialize at 0? i see it without initial value calcs in other code too.

src/main/common/kalman.c
```diff
@@ -49,22 +49,31 @@ void kalman_init(void)
...
+  if (kalmanState->windex >= kalmanState->w) {
+    kalmanState->windex = 0;
+  }
+
+  kalmanState->axisSumMean -= kalmanState->axisWindow[kalmanState->windex];
+  kalmanState->axisSumVar -= kalmanState->varianceWindow[kalmanState->windex];
+
+  //New mean
+  kalmanState->axisMean = kalmanState->axisSumMean * kalmanState->inverseN;
+  kalmanState->axisVar = kalmanState->axisSumVar * kalmanState->inverseN;
```
nerdCopter, line 72: @DzikuVx , we don't need absolute value here?  
i.e. older-code has `axisVar = fabsf(...);` & helio-imuf-code has `[x|y|z]Var = ABS(...);`.
i'm asking these things because i'm trying to back-port this math-fix to Helio-imu-firmware, thank you very much.

DzikuVx: @nerdCopter not really as `varianceElement` is squared, so it's always a positive number.

AndreySemjonov:
Positive test for me on Helio

PR #271, opened 2020-06-02
Fix dyn notch and static notch for helio

AndreySemjonov:
* Remove software lowpass filter for Helio since it is prefiltered with gyro lpf on IMUF side. No need to apply in double.
* Fix non-Applied dynamic_filter for helio.


nerdCopter:
* @AndreySemjonov I'm splitting this into two PR's, so that 
* we can merge the notch fix for certain, and
* the question of gyroLPF double filtering can be meditated on.

PR #274, opened 2020-06-03
20200603 andrey disable gyro lpf for helio

nerdCopter:
* disable standard gyroLPF's for Helio (Helio uses IMUF LPF's)


nerdCopter:
oops, still included included notches.

PR #275, opened 2020-06-03
Remove software lowpass filter for Helio

nerdCopter:
* disable standard gyroLPF's for Helio (Helio uses IMUF LPF's)
* need for this is debatable both perspectives.


loutwice:
i agree with @AndreySemjonov idea about gyro and helio . do  user need double flitering possibility?

PR #291, opened 2020-06-23
add debug modes for Kalman R and iDecay ITermNew

nerdCopter:
* `kalmanFilterStateRate[axis].r` is the measure of noise.  `set debug_mode = R_NOISE_MEASURE` to record this.
* iDecay's `ITermNew` is the calculation of decay. `set debug_mode = I_DECAY_CALC` to record this (applies only when non-zero and < Ki).

* `R_NOISE_MEASURE` works.
* `I_DECAY_CALC` needs testing. it records, but unsure how to analyze.


Vanalio:
We this we have actual/resulting frequency of LPF filter or only an evaulation of gyro "noisiness"?

nerdCopter:
* @Vanalio , only records the measurement of `R`
* R calclulation is `r = squirt * VARIANCE_SCALE;`
* record R multiplied by 1000 to visualize.

nerdCopter:
* not important to be in mainline
* was useful for some tests

PR #294, opened 2020-06-24
ZeezF7, Whoop

Kaiowarez:
blackbox functionality is still being worked on atm, current state is working without blackbox
need drivers updated for the blackbox to work

### DIFF
--- a/src/main/target/ZEEZF7/config.c
+++ b/src/main/target/ZEEZF7/config.c
@@ -1,0 +1,47 @@
+/*
+ * This file is part of Cleanflight and Betaflight.
+ *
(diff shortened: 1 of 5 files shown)


nerdCopter:
looks good, i think the drivers can just be ported in. will merge this.

nerdCopter:
@Kaiowarez 
why `#define TARGET_BOARD_IDENTIFIER                          "S7X2"`
and not `#define TARGET_BOARD_IDENTIFIER                         "ZEEZF7"`  ?

Kaiowarez:
I been defining all the boards lately by the stm processor it's using makes
it easy to work on later on.., it just makes more sense tbh

PR #297, opened 2020-06-24
ReWrite EmuBoost so that it uses less cpu

Quick-Flash:
Recode emuboost so that it uses a less cpu and is just better coded. I just need verification that the code is indeed the same.


nerdCopter:
* QF reported flew worse
* maybe revisit idea later

Quick-Flash:
maybe look into this again later,

PR #305, opened 2020-07-01
differentiate KakuteF7 V1.5

nerdCopter:
* differentiate  KAKUTEF7V15 from KAKUTEF7HDV
* defines camera-control, current-scale, and max7456 differently

PR #306, opened 2020-07-02
adding kakutef7v2 build routine

Kaiowarez:


nerdCopter:
@Kaiowarez  , so for V2, we don't need:
```
#define MAX7456_SPI_INSTANCE    SPI3
#define MAX7456_SPI_CS_PIN      SPI1_NSS_PIN
```
?

and we dont need to skip baro for v2:
```
#if !defined(KAKUTEF7V2)
#define USE_BARO_BMP280
#endif
```
?

Kaiowarez:
Need the max7456 for v2 that's the bf osd driver
V2 doesn't have baro?

nerdCopter:
> Need the max7456 for v2 that's the bf osd driver V2 doesn't have baro?

 yes it has BMP280, i was using your screencap from Vanadio, but searched myself instead and found yes for Baro

Kaiowarez:
Sometimes the unified config file also says baro, but the hardware isn't
there, we end up with i2c errors

nerdCopter:
closing this one only because i put these mods into #305

PR #310, opened 2020-07-07
SilF4Ware mix change limit

Quick-Flash:
Quite a good filter for the motor. This filter comes from SilF4Ware all credit to them. This filter places a limit on the maximum motor change. It actually works quite well at reducing high freq motor noise without much added latency. So its basically a free lunch for once. Nice.


gretel:
seems related to unmerged https://github.com/emuflight/EmuFlight/pull/251

gretel:
and https://github.com/emuflight/EmuFlight/pull/169

PR #313, opened 2020-07-08
HGLRCF411 enable flash

nerdCopter:
* enable flash as per @Kaiowarez


nerdCopter:
@ viruy reported success.

PR #315, opened 2020-07-09
gimbal mode `CINEMATIC` to hack around resource constraints when using `FEATURE_SERVO_TILT`

gretel:
requested and tested by @B-nutze-RR 🚁 

in context of `FEATURE_SERVO_TILT` this hack adds another gimbal mode… which bascially does not what it is named like but copies the same value for the roll servo to the pitch servo.

it reduces the number of timers required by one, which allows to use one resource less, i.e. keep the buzzer working on `HELIOSPRING` but have one servo to stabilize camera on the roll axis.

this should possibly not be merged to mainline while at the same time it does not change anything if not enabled by `set gimbal_mode = CINEMATIC`.

### DIFF
--- a/src/main/flight/servos.c
+++ b/src/main/flight/servos.c
@@ -376,10 +376,10 @@ void writeServos(void)
         break;
     }
 
-    // Two servos for SERVO_TILT, if enabled
+    // One servo for SERVO_TILT, if enabled
     if (feature(FEATURE_SERVO_TILT) || currentMixerMode == MIXER_GIMBAL) {
         updateGimbalServos(servoIndex);
-        servoIndex += 2;
+        servoIndex += 1;
     }
 
     // forward AUX to remaining servo outputs (not constrained)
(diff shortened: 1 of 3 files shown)


gretel:
i'd appreciate if you can leave the draft or not decision to the orignal authors. i cant merge in any way myself.

Quick-Flash:
looks good to me. thanks for this :)

gretel:
https://www.youtube.com/watch?v=iKIzSIF14Us

gretel:
consolidating the "cinema" features might be worth a thought.

src/main/flight/servos.c
```diff
@@ -376,10 +376,10 @@ void writeServos(void)
         break;
     }
 
-    // Two servos for SERVO_TILT, if enabled
+    // One servo for SERVO_TILT, if enabled
     if (feature(FEATURE_SERVO_TILT) || currentMixerMode == MIXER_GIMBAL) {
         updateGimbalServos(servoIndex);
-        servoIndex += 2;
+        servoIndex += 1;
```
Quick-Flash, line 382: What exactly is this line doing? Why has it been changed to +1 instead of +2?

gretel: just one servo

Quick-Flash: So how does it effect things if you want 2 serves.

gretel: this code ancient. the original idea was to control one servo per axis, pitch and roll. nicecrash was only looking to stabilize his gopro as shown on youtube - so i guess this should work for most cinema pilots.
it might be a breaking change for people using this with two servos but i bet the risk is super low 😸

gretel:
> consolidating the "cinema" features might be worth a thought.

was saying this should probably not get merged as is - while the risk is low tho.

PR #322, opened 2020-07-14
add FOXEERF722V2

nerdCopter:
* waiting on test results
* coded as per @Kaiowarez


nerdCopter:
`Grec's Whirlybird: Flies perfect.  Only Emuflight can make a Foxeer F7 feel like. Helio Spring.`

PR #327, opened 2020-07-20
Quickflash iterm relax

Quick-Flash:
Forces iterm relax on setpoint. It also allows you to set iterm relax for yaw different than for p/r. Set this to 0 to turn it off.


AndreySemjonov:
Yes it is good to go

PR #329, opened 2020-07-22
Differentiate Kakute F7 V15, HDV, Mini2 

Kaiowarez:
Redefining multi-build Kakute F7 V15, HDV, Mini2


Kaiowarez:
Oh true my mistake needs to be doubles

PR #330, opened 2020-07-22
Hglrcf722 1 gyro fix

Kaiowarez:
Fixed Hglrcf722 BB weird behavior


nerdCopter:
@Kaiowarez 
* this build, for whatever reason (makes no sense), causes BB log weirdness as reported by DaveSkyBiker; it fails to start immediatly on arming for some flights, but starts a few seconds late.
* apparently `EmuFlight_0.3.1_HGLRCF722_Build_ba4d71a12.hex` (`kaiowarez/PICKPOCKETS`) works correctly on two quads, but when i look at that code it still said `gyro_1_` & `gyro_2_` so do you think you code was not committed/pushed?

nerdCopter:
maybe `#define ENSURE_MPU_DATA_READY_IS_LOW` needs to stay? (i have zero clue)

Kaiowarez:
Hglrcf722 is weird lol

PR #344, opened 2020-08-25
Dynamic calculation of rate dynamics

Quick-Flash:
This fixes an issue that occurs which makes the feeling of dynamic rates depend on your receiver update rate. If you run crsf in dynamic mode and you go from 150hz to 50hz and run agressive rate dynamic settings you won't feel the transition nearly as badly as you used to. This may even fix auto smoothing by fixing currentRxRefreshRate.


nerdCopter:
this is good as far as i'm concerned, but waiting on testers consensus

Kaiowarez (approved):

PR #345, opened 2020-08-27
fix HGLRCF411 softserial assignment

nerdCopter:
* modded as per @Kaiowarez request.


nerdCopter:
* i recall one success, any others tested?

PR #350, opened 2020-09-03
crsf rssi

pleasuretek:
When crsf is selected as the rx protocol to use, the osd rssi will now display rfmode and link quality data in place of rssi in the format of "SYM rfmode : lq" with rfmode being (2,1,0) for (150hz, 50hz,4hz) and lq being 0-99 scale.


nerdCopter:
using a modified version (https://github.com/emuflight/EmuFlight/pull/353), but we thank you tremendously!  it would not have happened without you, thank you.

PR #357, opened 2020-09-21
Redo brushed flipoveraftercrash

mywinchester:
- Removed the old newbeedrone brushless motor flip over after crash code, which was not protected by macro definitions and therefore risked affecting the underlying layers.
- Restore BOXUSER4 named USER4, the new changes do not occupy PINIOBOX.
- Rewrite the brushed flipoveraftercrash function, the new change uses macro definition `USE_BRUSHED_FLIPOVERAFTERCRASH` to divide, and use macro definition `BRUSHED_REVERSE_PIN` to define the used pin.
- The `BRUSHED_REVERSE` pin resource has been added to the CLI so that the pin definition can be changed or viewed through the CLI.


nerdCopter:
@mywinchester , can you update your fork (sync) and merge master into this?

```
git remote add upstream https://github.com/emuflight/EmuFlight.git
git remote set-url --push upstream disabled
git remote -v #view remotes

git checkout master
git pull upstream master

git push origin #update fork's master

git checkout redo_brushed_flipoveraftercrash
git merge master #merge master into branch
```

Quick-Flash:
Thanks for this, it should have been done a long while ago, however, I have two questions/suggestions, shouldn't we still allow 3d mode to be enabled? and it doesn't look like the current target which supports this action will work with this code until we change that code as well. Could you make the changes to support 3d and support the newbeedrone brushed pro flight controller with this option of turtle mode?

mywinchester:
@Quick-Flash Well, maybe we can implement it in the next feature patch.

nerdCopter:
` QuickFlash: i've made targets using that pull request and they don't fly` -- for records so we stop referencing this PR internally until fixed.

PR #373, opened 2020-10-03
Kaiowarez crsf lq part 2

nerdCopter:


nerdCopter:
kaio wrongly included clracingf7. i tried to delete , but not from repo. closing

PR #376, opened 2020-10-04
Crsf lq part 2

Kaiowarez:
finallized CRSF OSD elements..
tested LQ properly switching
added noise level
added signal to noise ratio
added Tx Power

LQ test = https://youtu.be/CDPf6unYv24
snr/noise/tx elements =https://youtu.be/CVroHVWUAcQ


nerdCopter:
* Conflicts
* CLRACING files dont belong here

PR #377, opened 2020-10-04
CRSF LQ Proper [PR round 2!]

Kaiowarez:
* Show CRSF LQ mode 1 as 0-99, mode 2 as 0-300
* add dB and SNR
* configurable blink threshold `osd_lq_alarm`

finallized CRSF OSD elements..
tested LQ properly switching
added noise level
added signal to noise ratio
added Tx Power

LQ test = https://youtu.be/CDPf6unYv24
snr/noise/tx elements =https://youtu.be/CVroHVWUAcQ

### DIFF
--- a/src/main/cms/cms_menu_osd.c
+++ b/src/main/cms/cms_menu_osd.c
@@ -65,7 +65,7 @@ OSD_Entry menuOsdActiveElemsEntries[] =
     {"RSSI / LQ",               OME_VISIBLE, NULL, &osdConfig_item_pos[OSD_RSSI_VALUE], 0},
     {"CRSF TX POWER",      OME_VISIBLE, NULL, &osdConfig_item_pos[OSD_CRSF_TX], 0},
     {"CRSF SNR",           OME_VISIBLE, NULL, &osdConfig_item_pos[OSD_CRSF_SNR], 0},
-    {"CRSF NOISE LVL",     OME_VISIBLE, NULL, &osdConfig_item_pos[OSD_CRSF_RSSI], 0},
+    {"CRSF RSSI",     OME_VISIBLE, NULL, &osdConfig_item_pos[OSD_CRSF_RSSI], 0},
     {"BATTERY VOLTAGE",    OME_VISIBLE, NULL, &osdConfig_item_pos[OSD_MAIN_BATT_VOLTAGE], 0},
     {"BATTERY USAGE",      OME_VISIBLE, NULL, &osdConfig_item_pos[OSD_MAIN_BATT_USAGE], 0},
     {"AVG CELL VOLTAGE",   OME_VISIBLE, NULL, &osdConfig_item_pos[OSD_AVG_CELL_VOLTAGE], 0},
--- a/src/main/cms/cms_menu_osd.c
+++ b/src/main/cms/cms_menu_osd.c
@@ -62,7 +62,10 @@ static long menuOsdActiveElemsOnExit(const OSD_Entry *self)
 OSD_Entry menuOsdActiveElemsEntries[] =
 {
     {"--- ACTIV ELEM ---", OME_Label,   NULL, NULL, 0},
-    {"RSSI",               OME_VISIBLE, NULL, &osdConfig_item_pos[OSD_RSSI_VALUE], 0},
+    {"RSSI / LQ",               OME_VISIBLE, NULL, &osdConfig_item_pos[OSD_RSSI_VALUE], 0},
+    {"CRSF TX POWER",      OME_VISIBLE, NULL, &osdConfig_item_pos[OSD_CRSF_TX], 0},
+    {"CRSF SNR",           OME_VISIBLE, NULL, &osdConfig_item_pos[OSD_CRSF_SNR], 0},
+    {"CRSF NOISE LVL",     OME_VISIBLE, NULL, &osdConfig_item_pos[OSD_CRSF_RSSI], 0},
     {"BATTERY VOLTAGE",    OME_VISIBLE, NULL, &osdConfig_item_pos[OSD_MAIN_BATT_VOLTAGE], 0},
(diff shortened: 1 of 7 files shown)


OptimusTi:
Quick question and suggestion; have you guys read the latest Crossfire documentation?

I suggest reading pages 57-58 of the v4.0 Crossfire manual. Based on this information you could get a simpler solution than just throwing stuff at the wall and see what sticks. :)

I worked with the inav solution for this. Let me know if I can help.

antennanull

nerdCopter:
We welcome help @OptimusTi :)
FWIW, Trappy directly said he prefer 0-300.

OptimusTi:
> We welcome help @OptimusTi :)
> FWIW, Trappy directly said he prefer 0-300.

Haha yes, I saw the stream. But please understand that he may be a bit biased because that is what the TBS hardware (Full/Lite TX, EVO VTX, Fusion, Core, FPVision OSD, and who knows what's coming...) uses. ;)

The reality is that from firmware 2.93 CRSF sends LQ from [0..100%] to FC and OpenTx sensors, and that's what we have to work with.

Here are a few points:
  1. If you want the 300% range thing, then you'd need to scale the data sent by the firmware [0..100%] to [170-300%] when in 150Hz RFMD (see page 65 of 9/14/20 manual). No need for scaling when in 50Hz/4Hz since It's [0..100%].

  2. There's a video on the UAVFutures' YouTube channel where Trappy mentions SNR and how it works. SNR only shows when it reaches critical values. The Crossfire manual recommends 4dB as the "warning" level and 0dB as the "critical level (page 66). According to old docs, any SNR value above 8dB cannot be accurately measured, so showing values above that is useless. In Inav there's a setting that defaults to 5dB (personally, I set it to 3dB and forget about it).

  3. Looks like RSSI in dBm is working. In Inav we can also use and show RSSI for the active antenna when using diversity receivers.

  4. What's "CRSF NOISE LVL"?

Kaiowarez:
Noise level is what trappy said rssi is..
Well the way crsf Lq works is its only when crsf is the protocol
If we need a diffrent format for a diffre t protocol we can prolly auto
switch it as well

OptimusTi:
Well, I don't know what to say then. Noise level is not RSSI. I think you need to read up a bit. Start with the manual and some basic RF terms...

Good luck with the project.

Kaiowarez:
Ah rssi is rssi?
I figured it was kinda wrong all I got was trappy telling me what stuff is

Kaiowarez:
Appreciate leading me in the right direction tho.
My only real basis for stuff was how it looked on radix,

nerdCopter:
approve #378 as v1 of this, can make better in future.

Kaiowarez:
@OptimusTi with your lead, ended up with this solution https://youtu.be/7p1SK9HZSWk

OptimusTi:
> @OptimusTi with your lead, ended up with this solution https://youtu.be/7p1SK9HZSWk

Yes, that's pretty good! Someone suggested something similar for Inav and if the menu system allows for it, I may add LQ format.

Only a couple of comments:
SNR:  You only need to show SNR when it reaches warning or critical values (4dB and 0dB. page 66) Anything over 8dB (according to TBS), is too random and noisy to give any useful information. Remember, SNR is the ratio between noise floor and received signal were 1:1 SRN (0dB) is bad, m'kay? This gif shows how I we have snr working:

snrcmsfix

FREQ LQ mode: I'd skip this one or add a separate element to show the frequency. We had an element for the rate mode in Inav but since it's a bit redundant to crsf users, we took it out to save screen space. For racers, I think it would be good as its own element.

Diversity: TBS hardware (EVO vtx, etc) shows a second element for Ant2 RSSI if you use a diversity receiver. In Inav, to save screen space, I used "active antenna" to get the "active" RSSI and only use one element A symbol shows the active antenna. Few people use diversity receivers but hey, why not kill two birds with one stone.

Test all rfmodes, Dynamic, forced 150Mhz, forced 50Mhz and Forced Telemetry to make sure the functionality behaves as expected by TBS/Crossfire.

Cheers

Kaiowarez:
For diversity, i have a function. That compares rssi1 vs 2 and shows lower
value

Snr I just followed along radix setup just coz me and madchiller was used
to it always being shown. But I do get the point of 8db or lower is
important aspect of it after reading the manual.

Still need a few more work but I was just mentioning that I did take your
advice, reading thru the messages it sounded a Lil rude didn't mean it that
way.
Writing emails @ work had to keep it short side.

MatPottsFPV:
Being able to have RFMD in OSD would be nice and seperating RSSI from LQ for OSD settings would also be nice. As its hard to understand what is being shown if both could be being shown at the same time?

OptimusTi:
> Being able to have RFMD in OSD would be nice and seperating RSSI from LQ for OSD settings would also be nice. As its hard to understand what is being shown if both could be being shown at the same time?

Since 3.0, pretty much everything you are asking was added. On the next release, RSSI will be based on signal strength instead of LQ, like you suggest.

MatPottsFPV:
> > Being able to have RFMD in OSD would be nice and seperating RSSI from LQ for OSD settings would also be nice. As its hard to understand what is being shown if both could be being shown at the same time?
> 
> Since 3.0, pretty much everything you are asking was added. On the next release, RSSI will be based on signal strength instead of LQ, like you suggest.

Oops sorry then! I actually found the info on how to change different lq settings on osd shortly after you replied yesterday. You all do wicked work! Thank you!

OptimusTi:
> > > Being able to have RFMD in OSD would be nice and seperating RSSI from LQ for OSD settings would also be nice. As its hard to understand what is being shown if both could be being shown at the same time?
> 
> > 
> 
> > Since 3.0, pretty much everything you are asking was added. On the next release, RSSI will be based on signal strength instead of LQ, like you suggest.
> 
> 
> 
> Oops sorry then! I actually found the info on how to change different lq settings on osd shortly after you replied yesterday. You all do wicked work! Thank you!

Oops sorry, I didn't notice this was Emuflight. I thought it was an Inav comment. Ignore what I said.

PR #379, opened 2020-10-07
Minimum allowed IMUF Q 100

nerdCopter:
* Do not allow Q <100
* tested CLI good
* did not test OSD configuring


Quick-Flash (approved):
send it!

PR #380, opened 2020-10-07
ability to turn off kalman

Quick-Flash:
Add a way to turn off the kalman filter if you just want to see for yourself how great it is and how much you will miss it (can't see why else you would do this) If w is less than 3 turn off kalman filtering.


Quick-Flash:
this should be good to go at any time.

nerdCopter (approved):
* i understood from tester that it does what it supposed to, but obviously better when enabled.

PR #386, opened 2020-10-09
port of (rejected) betaflight#9263 - Add no-RX option for broken Tramp implementations on some VTX 

gretel:
port of (rejected) https://github.com/betaflight/betaflight/pull/9263/files - Add no-RX option for broken Tramp implementations on some VTX 

bf devs dont get paid by `eachine` so they say the hardware is [no good](https://github.com/betaflight/betaflight/pull/9263#issuecomment-568939830) and should not be supported.
personally i feel `eachine nano` is good vtx and should be supported.

tested on `FURYF4OSD` with **and** without a `eachine nano v1`:
 * afaik the `v2` has the tramp implementation fixed, but i dont own one.
 * the `v1` one does not work without `vtx_tramp_norx = true`.


Quick-Flash:
question, how will this work in the 1.0.0 version of emuflight we are working on? does this effect bf 4.2+?

gretel:
> question, how will this work in the 1.0.0 version of emuflight we are working on? does this effect bf 4.2+?

no idea seems like bf has no tolerance for eachine :) https://github.com/betaflight/betaflight/pull/9263#issuecomment-568939830

Quick-Flash:
> > question, how will this work in the 1.0.0 version of emuflight we are working on? does this effect bf 4.2+?
> 
> no idea seems like bf has no tolerance for eachine :) [betaflight/betaflight#9263 (comment)](https://github.com/betaflight/betaflight/pull/9263#issuecomment-568939830)

Ok, we just need to be looking into more of the solutions like this and really determining how we can get things like this to port over to the 1.0.0, as this code is just in the last few maintenance stages at this point and 1.0.0 is starting to get more attention.

Quick-Flash (dismissed):
Also looks good!

gretel:
dont worry it's just a little change i did so i do not have to trash working gear just because people say it's eachine fault :)

gretel:
@Quick-Flash why am i blocked commenting on issues?

PR #390, opened 2020-10-09
get rid of 'FEATURE_DYNAMIC_FILTER' and have a setting 'dynamic_filter' instead

gretel:
get rid of 'FEATURE_DYNAMIC_FILTER' and have a setting 'dynamic_filter' instead; needs to be reflected gui side; seems to work but i cant tell the difference indoors @Quick-Flash


Quick-Flash:
question??? after looking into the code more, what is the purpose of "features" what makes something a feature and something not a feature? It doesn't look like there is any clear standard. Unfortunate. in 1.0.0 we should really make time to rethink things like this and come up with a solid answer as to where you should stick everything in code and make a good standard which we follow.

gretel:
enjoyed diggin' in this old code.. guess features were meant to switch things on a global level.. eventually depending on hardware support. while settings bring the more fine-grained configuration in. there are many cases that could be clarified.. unified.. certainly.

PR #396, opened 2020-10-22
CRSF SIMPLE(0-100) & SCALE(170-300)

Kaiowarez:

### DIFF
--- a/src/main/cms/cms_menu_osd.c
+++ b/src/main/cms/cms_menu_osd.c
@@ -52,7 +52,7 @@ static long menuOsdActiveElemsOnEnter(void) {
 }
 
 static const char * const cms_osdcrsfformat[] = {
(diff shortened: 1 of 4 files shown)


nerdCopter (changes requested):
* i think we should remove case TBS, and let it fall-though to default. i.e. no reason to have duplicate code, and results the same.
* optionally scale 0-100 to 170-300 as discussed. `osdLQfinal = scaleRangef(osdLQ, 0, 100, 170, 300);`, AFAIK.

Kaiowarez:
Could make tbs default, then scaled, mode, freq, simple?

nerdCopter (commented):
@Kaiowarez , can you scout and find out if Tracer "just works" and if we need mode 4 "250hz" text?

nerdCopter (approved):

PR #397, opened 2020-10-27
fix: make rate dynamics feel the same regardless rx rate

knoopx:
Hi there, I noticed there was a flaw in the current implementation when I recently switched my racing whoops to redpine protocol (1.5ms latency) on my own betaflight fork with backported rate dynamics from emuflight.

Using the same rate profiles I used before on frsky d8 spi, rate dynamics felt like not working at all because the protocol latency was so low that RD was unnoticeable.

This PR makes rate dynamics feel the same regardless of the RX protocol refresh rate you are currently using. I made a web-based prototype before implementing this out, so you can check it out here: https://tth0v.csb.app/. Untick _RX Refresh Rate Invariant_ and set _RX rate_ to max to see how RX Rate affects rate dynamics. When ticking _RX Refresh Rate Invariant_, my current proposal will be active and the "springiness" should be the same regardless of the rx rate.

Note that the original implementation notes mentions that RD settings should match 150hz but https://stoot.tech/emu-rate-dynamics.html uses 16ms (62,5hz) and most of the example wiki profiles feel like tuned on frsky rather than on crossfire.

### DIFF
--- a/src/main/fc/fc_rc.c
+++ b/src/main/fc/fc_rc.c
@@ -55,8 +55,6 @@
 #include "sensors/battery.h"
 #include "sensors/acceleration.h"
 
-#define M_PI_FLOAT      3.14159265358979323846f
-
 enum {
     ROLL_FLAG = 1 << ROLL,
     PITCH_FLAG = 1 << PITCH,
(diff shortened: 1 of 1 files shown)


Quick-Flash:
Thanks for catching this, I must have made some sort of algebraic mistake when trying to get this right myself. I'll need to port this over to the emuflight 1.0.0 branch as well (its the betaflight port of emuflight). I'm happy to see that even some betaflight users are enjoying rate dynamics, its a bummer that it wont see implementation there, but thanks for this PR.

src/main/fc/fc_rc.c
```diff
@@ -740,22 +738,17 @@ bool rcSmoothingInitializationComplete(void) {
...
-    // then using that frequency calculate a new k (or multiplier) for correction and weight
-    // did some math to find how to calculate freq if you know k and dt in a pt1
-    // k is our weight/correction value in decimal
-    const float freq = k / (2.0f * M_PI_FLOAT * 0.00666666666f * (1.0f - k));
-    const float dT = time * 1e-6f;
-    const float RC = 1.0f / (2.0f * M_PI_FLOAT * freq);
+
+    // scale so it feels like running at 62.5hz (16ms) regardless of the current rx rate
+    const float rxRate = 1.0f / time * 1000000.0f;
+    const float dT = time / 1000000.0f;
```
Quick-Flash, line 748: time * 1e-6f is more consistent with the rest of the codebase compared to time / 1000000.0f and is mathematically the same number.

The calculation of rxRate is simply 1/dT. instead of manually calculating dT twice, just move the dT calculation to before rxRate and set rxRate = 1.0f / dT;

src/main/fc/fc_rc.c
```diff
@@ -740,22 +738,17 @@ bool rcSmoothingInitializationComplete(void) {
...
+    const float RC = 1.0f / (2.0f * M_PI * k / (2.0f * M_PI * (1 / ((rxRate / 62.5f) * rxRate)) * (1.0f - k)));
```
Quick-Flash, line 749: question about this line. Isn't (2*pi*k/(2*pi*(other random things)*(more things)) = k/((other random things)*(more things))?

Quick-Flash (changes requested):

knoopx:
@Quick-Flash simplified ;)

Quick-Flash (approved):
Thanks for the changes, looks good to me!

knoopx:
Works on the web prototype, will try tomorrow on fc.

knoopx:
tried today on two whoops (frsky d8/redpine) and a 3" toothpick with crossfire, working fine 👍

PR #398, opened 2020-10-30
Mixer smoothing

tylercorleone:
This PR brings different changes to the way the controller's output (roll, pitch and yaw) is mixed with throttle to obtain a motor output.

The needs of roll and pitch can interfere with yaw's needs, controller's output interferes with throttle demands etc.
Here are proposed two alternatives to the LEGACY mixer, SMOOTH and 2PASS.

Both SMOOTH and 2PASS can use an option called "laziness" that enables a smarter "clipping prevention" strategy, that is called lazy because it adds only the minimum required amount of offset to remain in the [0, 1] for each "motor group" (that is a group of opposite motors that can actually work in conjunction in order to obtain a certain amount of roll/pitch) instead of applying the biggest needed amount. In this way we avoid to add/increase the total thrust when not needed, obtaining a better "throttle authority".

The scope of the SMOOTH mixer is just to avoid the harsh clipping strategy that the legacy mixer uses to keep the sum of the  controller's output (roll + pitch + yaw) and throttle in the [0, 1] range.

These graphs show the total thrust and the torque applied by two opposite motors with a respective output difference of 0.2 varying throttle and motorMixRange (maximum controller's output range).

AirMode OFF

1 - LEGACY
2 - SMOOTH - laziness OFF
3 - SMOOTH - laziness ON

airmode_off

AirMode ON

1 - LEGACY
2 - SMOOTH - laziness OFF
3 - SMOOTH - laziness ON

airmode_on

Laziness assumes that the torque obtained given a difference of X between two motors is independent from the offset they are working to. That is, it requires thrust linearization (TL).
This PR brings a TL formula that speculates that the thrust/RPM strategy can be expressed as the composition of a quadratic and a linear curve and uses its mathematical inverse to compensate for that non-linearity.

Here a tool to play with to calculate the correct parameters to use for low and high RPM compensation (0.65 and 0.3 is a good midpoint): https://www.desmos.com/calculator/zal1gyqrcn

IMG_20201221_065858

In this example you first have to match the thrust response curve of your quad (there are a couple already provided taken from www.miniquadtestbench.com), and then find the correct values to make it linear.

The two pass mixer (2PASS) handles yaw output and roll/pitch output differently. The first needs RPM linearity, hence motorOutput linearity, although it's not exactly a linear relationship, as you can see from other Mini Quad Bench graphs, while the seconds need *thrust* linearity.

This is the motorOutput at different throttle levels, when roll+pitch sum has the same magnitude of yaw (50/50):

image

yaw output is linear, roll+pitch output is linear too (what you see here is motor output, so motor and props will make it linear)

Here the roll+pitch/yaw rate is 90/10:

image

same roll and pitch authority while preserving a little more throttle authority

Here the roll+pitch/yaw rate is 10/90:

image

again yaw linearity while also preserving throttle authority

Here is the comparison between LECACY and 2PASS with laziness ON, that is the preferred combination so far (50/50 roll+pitch/yaw rate):

image

Keep in mind that the 0 axis represent the separation between different motors.


src/main/flight/mixer.c
```diff
@@ -766,17 +800,59 @@ float applyThrottleLimit(float throttle) {
...
+        } else {
+            motorMix[i] = scaleRangef(throttle, 0.0f, 1.0f, authorityZeroThrottle * (motorMix[i] + motorMixDelta), authorityFullThrottle * (motorMix[i] - motorMixDelta));
+        }
+        motorMix[i] /= motorMixNormalizationFactor;
+    }
+}
+
+// Predictive AirMode actually predicts the need of greater authority based on stick movements
+float calculatePredictiveAirModeAuthorityMultiplier() {
+    float maxStickMovement = MAX(stickMovement[ROLL], MAX(stickMovement[PITCH], stickMovement[YAW])); // [0, 1], the max r/p/y stick movement
```
borisbstyle, line 836: Using stick input with only pt1 filter will make the motor outputs more jerky?

Quick-Flash: This value is the filtered rc command. So we add an additional pt1 filter.

borisbstyle: There is also smoothed/filtered stick data
See: https://github.com/emuflight/EmuFlight/blob/master/src/main/fc/fc_rc.c

Quick-Flash: If im not mistaken the filtered RC command you are referring to is what is being used here. Maybe im wrong though.

src/main/flight/mixer.c
```diff
@@ -766,17 +800,59 @@ float applyThrottleLimit(float throttle) {
...
+    float motorMixNormalizationFactor = motorMixRange > 1.0f && hardwareMotorType != MOTOR_BRUSHED ? motorMixRange : 1.0f;
+    float motorMixDelta = 0.5f * motorMixRange;
+    float authorityZeroThrottle, authorityFullThrottle;
+
+    if (isAirmodeActive()) {
+        authorityZeroThrottle = authorityFullThrottle = 1.0f;
+    } else {
+        float authorityMultiplier = predictiveAirModeAuthorityMultiplier ? calculatePredictiveAirModeAuthorityMultiplier() : 1.0f;
+        authorityZeroThrottle = MAX(minAuthorityZeroThrottle * authorityMultiplier, 1.0f);
+        authorityFullThrottle = MAX(minAuthorityFullThrottle * authorityMultiplier, 1.0f);
```
borisbstyle, line 815: I am not sure if this makes sense to me. This will always result into 1? minAuthorityZeroThrottle * authorityMultiplier cannot exceed 1 in  calculatePredictiveAirModeAuthorityMultiplier()

src/main/flight/mixer.c
```diff
@@ -766,17 +800,59 @@ float applyThrottleLimit(float throttle) {
...
 
-FAST_CODE_NOINLINE void mixTable(timeUs_t currentTimeUs) {
+FAST_RAM_ZERO_INIT float lastRcDeflection[3], stickMovement[3];
+
+void applyAirMode(float *motorMix, float motorMixMax) {
+    float motorMixNormalizationFactor = motorMixRange > 1.0f && hardwareMotorType != MOTOR_BRUSHED ? motorMixRange : 1.0f;
+    float motorMixDelta = 0.5f * motorMixRange;
+    float authorityZeroThrottle, authorityFullThrottle;
+
+    if (isAirmodeActive()) {
```
borisbstyle, line 810: This is a bit contradicting to the name of the function. In here applyAirMode() is called, with still inside check if airmode is active.
In other words. applyAirMode functions does everything. AIRMODE OFF, AIRMODE ON and AIRMODE 2.0

src/main/flight/mixer.c
Quick-Flash: Authority multiplier can be greater than 1. In fact authority multiplier can only be 1 or greater than 1.

skylupop: Yes, I have to fix it, it's an error I made refactoring the Predictive AirMode logic. Thanks you!

src/main/flight/mixer.c
skylupop: I agree, we have to change the name of the method "applyAirMode" into something more correct. Any idea?

tylercorleone: I preferred refactoring the logic a little, at least now we don't call that method inside here.
The problem still remains, because the "airMode" logic gets always applied, even if the "isAirmodeActive" flag isn't active (but also the old logic did).
It's something like isAirmodeActive should be called useFullAuthority or something like that.

Quick-Flash: well, to be fair this code always forces you to have some level of airmode present even when airmode is off, but perhaps calling it applyMixerClipFix()??? don't quite like the name but it describes it fairly well.

tylercorleone: Yes, something like that, it prevents clipping but I wouldn't call it "fix", maybe adjustment.

tylercorleone:
Hi guys, I updated both graphs updating plots order and labeling (I think that "authority" explains better the concept than "torque %", but I leaved it in the axis legend).

Also I updated the graphs of the AirMode OFF because in the current implementation we don't have a selectable authority value for the middle throttle, so it scales between the values for zero throttle and full throttle.

Maybe it can be useful to introduce a configurable third value for an authority value, e.g. 30% throttle -> 100% authority.
I say that because I think that having full authority at "medium" throttle is more important than having it at full throttle, especially in a quad that doesn't use thrust linearization, where having less authority at 100% throttle acts as a TPA.

src/main/flight/mixer.c
borisbstyle: Yes you are right. You are getting it from rcCommandf, which is already filtered before being converted to the floats. Is extra filtering really needed?

tylercorleone:
I wanna stress out the Motor Mixer 2.0 only changes the behavior of motors not directly interested in the main/principal quad's movement.

These are the same graphs of above where the two opposite motors don't have a constant respective output difference, but is the motorMixRange itself.
V2.0 is the same as the normal smooth motor mixer:

AirMode OFF:

airmode_off_mitorMix_same_motorMixRange

AirMode ON:

airmode_on_mitorMix_same_motorMixRange

src/main/flight/mixer.c
Quick-Flash: The extra filtering gives us two advantages. It delays it slightly and keeps the effect going once we have stopped moving our sticks for a bit of time.

gretel:
this pr what is shown in https://www.youtube.com/watch?v=A_rQs3Uqkq0 ?

borisbstyle:
Ok I finally managed to do some mixer checks. I did not fly yet. But your mixer strategy to keep the thrust equal seems fine. It gives the full authority, but the throttle offset is selected better. Just like shown in your graphs.
However I was getting exactly same results with 2_0 airmode and without
Basically in here I was getting exact same behaviour:
```
        if (useAirMode2_0) {
            motorMix[i] = scaleRangef(throttle, 0.0f, 1.0f, authorityZeroThrottle * (motorMix[i] + ABS(motorMix[i])), authorityFullThrottle * (motorMix[i] - ABS(motorMix[i])));
        } else {
            motorMix[i] = scaleRangef(throttle, 0.0f, 1.0f, authorityZeroThrottle * (motorMix[i] + motorMixDelta), authorityFullThrottle * (motorMix[i] - motorMixDelta));
        }
```

I do think you are trying to change too much in this single PR. In order to really flight test this linear mixer behavior it may be better to get rid of all other stuff, that you introduced in this PR.
I think it may be good to only keep applyMixerClipAdjustment() part of code and don't make it too much complicated yet.

Perhaps just keep the authorityZeroThrottle and authorityFullThrottle to configure for now?

gretel:
have to agree with @borisbstyle here in terms of this being too much for one pr.

gretel:
btw https://github.com/betaflight/betaflight/pull/10294

tylercorleone:
I agree too, basically I should leave the applyMixerClipAdjustment() and the thrust linearization (it is needed by the 2.0 thing, it's something I really care about and it integrates well and simplifies the battery compensation logic) and leave out the @Quick-Flash part for the moment. I didn't understand "Basically in here I was getting exact same behaviour".
Also I would rename AirMode 2.0 in something else, since it applies to the clipping issues. Idk but I was thinking to "lazy mixer" 🤣, but I don't like it actually.
Any ideas?

gretel:
@tylercorleone i would like to do `set mixer_laziness = ON` just by the name of it :)

borisbstyle:
> I agree too, basically I should leave the applyMixerClipAdjustment() and the thrust linearization (it is needed by the 2.0 thing, it's something I really care about and it integrates well and simplifies the battery compensation logic) and leave out the @Quick-Flash part for the moment. I didn't understand "Basically in here I was getting exact same behaviour".
> Also I would rename AirMode 2.0 in something else, since it applies to the clipping issues. Idk but I was thinking to "lazy mixer" 🤣, but I don't like it actually.
> Any ideas?
Well I wrote a small piece of simulation software, where I can play with different scenarios from stick inputs, gyro data and also test separate mixer scenarios.
For mixer I introduced some basic test scenarios, where mixer is tested under several circumstances, like different throttle inputs, mixdeltas and old vs new mixer behaviour. I pretty much wanted to verify behavior you claim in your graphs.
In these simple scenarios I didn't see any difference between these 2 calculations.
motorMix[i] = scaleRangef(throttle, 0.0f, 1.0f, authorityZeroThrottle * (motorMix[i] + **ABS(motorMix[i])**), authorityFullThrottle * (motorMix[i] - **ABS(motorMix[i])**));
behaved equal to:
motorMix[i] = scaleRangef(throttle, 0.0f, 1.0f, authorityZeroThrottle * (motorMix[i] + **motorMixDelta**), authorityFullThrottle * (motorMix[i] - **motorMixDelta**));_

I am not sure why. Maybe I did something wrong and it also got late, but I will double check. I used (1.0 factor) for zero and full throttle authority for my tests and I didn't enable predictive airmode.

I think for naming convention it might be good to first flight test it and see how it feels :D. Than perhaps relate it to some "feeling". I would expect and hope it gives more organic feeling on sharper stick inputs, but that is just how I think it might be.
Back in the days, when I came with Airmode it was, because it gave a lot more authority on low throttle, compared to traditional mixer, where authority was being capped up to 50%.
Don't also forget, that more "lazy" mixer might also need retuning of PID's. I would expect, that PID's generally could be increased.

borisbstyle:
> @tylercorleone i would like to do `set mixer_laziness = ON` just by the name of it :)

Yeah that is what users want to see. Enable something and fly :D or make it default if it's good.

Then eventually additional stuff can be implemented later on as enhancements.

tylercorleone:
@borisbstyle yes, the "lazy" mode works exactly the normal mode if you look only at the motors with the highest offset, like I wrote in a comment above. The difference is that not all the motors will receive the offset high or low

borisbstyle:
> btw [betaflight/betaflight#10294](https://github.com/betaflight/betaflight/pull/10294)



> @borisbstyle yes, the "lazy" mode works exactly as the normal mode if you look only at the motors with the highest offset, like I wrote in a comment above (identical graphs). The difference is that not all the motors will receive the same amount of offset because ABS(motorMix[i]) is not equal to motorMixDelta for all the motors.
> 
> There is a discussion on Discord, from the first tests it seems that it behaves as it did on my quad. Less nervous, more predictable, no bounces when on the ground, no "vacuum effect" when you get hit the wall etc.

Ok. That makes sense. I said it was simple test. I was only testing single axis.

src/main/flight/mixer.c
```diff
@@ -623,6 +648,14 @@ static void calculateThrottleAndCurrentMotorEndpoints(timeUs_t currentTimeUs) {
 #define CRASH_FLIP_DEADBAND 20
 #define CRASH_FLIP_STICK_MINF 0.15f
 
+static float mapThrustToMotorOutput(float thrust)
+{
+    float vbatCompFactor = currentControlRateProfile->vbat_comp_type != VBAT_COMP_TYPE_OFF ? calculateBatteryCompensationFactor() : 1.0f;
```
tylercorleone, line 653: TODO: move this calculation out

src/main/flight/mixer.c
```diff
@@ -826,19 +922,9 @@ FAST_CODE_NOINLINE void mixTable(timeUs_t currentTimeUs) {
 #endif
     loggingThrottle = throttle;
     motorMixRange = motorMixMax - motorMixMin;
```
mituritsyn, line 924: propably would better to include motorMixRange calc to applyMixerClipAdjustment func?

Quick-Flash:
After more testing, I'm really unsure of the usefulness of predictive airmode. Axis lock (or whatever we want to call it) does however seem useful.

tylercorleone:
> After more testing, I'm really unsure of the usefulness of predictive airmode. Axis lock (or whatever we want to call it) does however seem useful.

Yes, with the lazy mixer maybe we don't need Predictive AirMode. I didn't tried the axis lock, but I think the suggestion of Borisbstyle was just to make the PR smaller and easier to test.

src/main/flight/mixer.c
```diff
@@ -623,6 +635,12 @@ static void calculateThrottleAndCurrentMotorEndpoints(timeUs_t currentTimeUs) {
 #define CRASH_FLIP_DEADBAND 20
 #define CRASH_FLIP_STICK_MINF 0.15f
 
+static float mapThrustToMotorOutput(float thrust, float vbatCompFactor)
+{
+    float linearizedThrust = vbatCompFactor * ((1.0f - thrust_linearization_level) * thrust + thrust_linearization_level * SIGN(thrust) * vbatCompFactor * sqrtf(ABS(thrust)));
```
skylupop, line 640: Not sure if the sign management is necessary. I did it because of the 3D mode, but actually I don't know if here we can have a negative thrust value. If not the formula could be cleaned up.

Quick-Flash: ah i didn't notice this, but your completely correct. Thrust should never be able to have a negative sign.

tylercorleone: Thank you Quick 🙂

src/main/flight/mixer.c
```diff
@@ -793,7 +759,7 @@ float applyThrottleLimit(float throttle) {
 
 void mixWithThrottleLegacy(float *motorMix, float controllerMixMin, float controllerMixMax) {
     // clipping adjustment
-    if (controllerMixRange > 1.0f && (hardwareMotorType != MOTOR_BRUSHED)) {
```
Quick-Flash: why was this removed?

skylupop: because if we don't normalize we obtain negative values that cannot be linearized (sqrt).

But other that that, considering  motorOutput = motorOutputMin + motorMix[i] * motorOutputRange, without normalization if the mixer has a value > 1 (we can have up to +-3 if I'm not wrong) this would exceed the motorOutputRange. The scope to have a normalized MotorMix is to distribute controller/throttle decisions over the full range of operation of the motors. Without normalization I think that the values > 1 or < 0 would just be clipped.

Quick-Flash: yes correct values above and below 1 and 0 are just clipped. However this line of code was added because of a lot of testing that silverware developers made when it comes to brushed motors. They found that brushed motors react so much slower than their brushless counterparts that just clipping without scaling the outputs leads to brushed motors feeling as though they have more power. Since the brushed motors act slow you don't notice the imbalance that not normalizing the outputs creates because they don't react fast enough for that to become an issue.  Hence for brushed whoops (the only place brushed is still used) you can get a whoop that feels like it has more power and can thus go around the track faster. This has been tested and proven by pilots such as project mockingbird.

skylupop: I understand, so can I constrain values for brushed motors to avoid negative values? And to make explicit that values beyond 0-1 are just clipped? I think it's better to guarantee that mixer's value are always in the 0/1 range because all the logic works with that assumption.

borisbstyle:
Just out of curiosity. Why did you call the mixer "LAZY"? I would prefer the name DYNAMIC, because it really dynamically adopts to the best possible thrust position. It especially becomes better, when multiple axes are requesting authority.

tylercorleone:
> Just out of curiosity. Why did you call the mixer "LAZY"? I would prefer the name DYNAMIC, because it really dynamically adopts to the best possible thrust position. It especially becomes better, when multiple axes are requesting authority.

I called like that because it does the minimum needed work to avoid clipping. The name sounded funny, the idea of a lazy Emu was funny to me 😋 Dynamic maybe describes better what it's doing, but if you think about it "dynamic" could be referred to different aspect of the mixing. LAZY is just an un-serious "battle name" 🙂

borisbstyle:
> > Just out of curiosity. Why did you call the mixer "LAZY"? I would prefer the name DYNAMIC, because it really dynamically adopts to the best possible thrust position. It especially becomes better, when multiple axes are requesting authority.
> 
> I called like that because it does the minimum needed work to avoid clipping. The name sounded funny, the idea of a lazy Emu was funny to me 😋 Dynamic maybe describes better what it's doing, but if you think about it "dynamic" could be referred to different aspect of the mixing. LAZY is just an un-serious "battle name" 🙂

OK clear. That is indeed another way to look at term "lazy" :D

src/main/flight/mixer.c
```diff
@@ -766,51 +743,79 @@ float applyThrottleLimit(float throttle) {
...
+                throttle = constrainf(throttle, -motorMixMin, 1.0f - motorMixMax);
+            }
+        }
+    }
+
+    float vbatCompFactor = calculateBatteryCompensationFactor();
+
+    for (int i = 0; i < motorCount; i++) {
+        if (mixerImpl == MIXER_IMPL_SMOOTH) {
+            motorMix[i] += motorMixDelta - motorMixMax; // let's center values around the zero
```
borisbstyle, line 778: Is this needed? In my opinion this line isn't doing anything. motorMixDelta - motorMixMax will always result to 0.

tylercorleone: You are right, it could also cause problems in those mixers that couple e.g. one 1.0 component with two -0.5 components (tricopters or something like that). I removed this line in the new version of the mixer!
Thank you.

tylercorleone:
> heres how I would write this code.
> if (!currentPidProfile->mixer_thrust_linearization_level) { // TPA is not applied when Thrust Linearization is enabled
> // calculating the PID sum and TPA and SPA
> // multiply these things to the pidData so that logs shows the pid data correctly
> pidData[axis].P = pidData[axis].P * getThrottlePAttenuation() * setPointPAttenuation;
> pidData[axis].I = temporaryIterm[axis] * getThrottleIAttenuation() * setPointIAttenuation;
> pidData[axis].D = pidData[axis].D * getThrottleDAttenuation() * setPointDAttenuation;
> } else {
> // calculating the PID sum and TPA and SPA
> // multiply these things to the pidData so that logs shows the pid data correctly
> pidData[axis].P = pidData[axis].P * setPointPAttenuation;
> pidData[axis].I = temporaryIterm[axis] * setPointIAttenuation;
> pidData[axis].D = pidData[axis].D * setPointDAttenuation;
> }

done

Quick-Flash (approved):
its time, defaults for thrust linear may still need to be played with though

PR #403, opened 2020-11-04
port of betaflight#8613 - Correct IMU gyro input to be based on sample loop time rather than actual delta time

gretel:
port of https://github.com/betaflight/betaflight/pull/8613/files - Correct IMU gyro input to be based on sample loop time rather than actual delta time

flight tested using a `MATEKRX411` on `8/8` 👍


gretel:
also tested on `FURYF4OSD` and `KAKUTEF7MINIV1`.

Quick-Flash (approved):

gretel:
@Quick-Flash i can feel this. maybe it's placebo. like all.

PR #404, opened 2020-11-04
Mix roll and yaw

Quick-Flash:
This code adds an option to mix roll stick into yaw, or yaw stick into roll. This is something usually done on the radio, but our slightly different method vs the radio method seems to have a slight benefit. Really really good for top angle whoop racing pilots. Also adds roll pitch magnitude expo. Roll pitch magnitude expo adds an expo to your roll and pitch when the magnitude of pitch and roll are greater than 1.0f. Then it applies extra expo to these sticks and reduces the max value of both pitch and roll. Seems to be nice for racing angle whoops as well. Hasn't been tested to much on acro quads, but its a totally optional stick feel thing. Naming might want to be changed to make it easier for users to understand, but I think its good for now.


nerdCopter (approved):
* "LGTM"
* approving based on ya'll internal testing.

PR #407, opened 2020-11-17
github action workflow update/fixes

nerdCopter:
* github actions update
* required due to GitHub deprecations

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,28 +43,28 @@ jobs:
       uses: actions/setup-python@v1
 
     - name: Setup Toolchain
-      uses: fiam/arm-none-eabi-gcc@v1.0.2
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -43,28 +43,28 @@ jobs:
       uses: actions/setup-python@v1
 
     - name: Setup Toolchain
-      uses: fiam/arm-none-eabi-gcc@v1.0.2
+      uses: fiam/arm-none-eabi-gcc@master
```
DusKing1, line 46: Just a question, why fiam?

nerdCopter: is there another action source that does the same?

Quick-Flash (approved):

madchiller (approved):
Moving forward

PR #417, opened 2020-12-03
enhancement: CLI: allow vtx power to be changed by a switch

Igorshp:
Backported from Betaflight: Adds an option to change vtx power with an aux channel.

more info on usage:
https://github.com/betaflight/betaflight/wiki/VTX-CLI-Settings

Example cli settings:
```
#channels on a 6 pos switch
vtx 0 7 5 1 0 900 1100
vtx 1 7 5 2 0 1100 1300
vtx 2 7 5 3 0 1300 1500
vtx 3 7 5 4 0 1500 1700
vtx 4 7 5 6 0 1700 1900
vtx 5 7 5 8 0 1900 2100
#vtx power on a 2 pos switch
vtx 6 6 0 0 1 900 1200
vtx 7 6 0 0 2 1800 2100
```

thanks @Quick-Flash  and @nerdCopter for help! 

This was was tested on Tinyhawk2 board (matekf411rx target) and works perfectly.

### Testing

To test this, flash your FC with the build.
Setup a switch to control a spare Aux Channel on the radio.

Run the following commands in cli:
```
vtx 1 6 0 0 1 900 1200
vtx 2 6 0 0 2 1800 2100
```

The above uses AUX7 ( channels are zero indexed, change to whatever channel number you have set up on the radio minus 1) and will switch between power modes 1 and 2 when channel is withing related ranges.

Easiest way to tell is to enable VTX osd element. You will see power change in the OSD. Make sure you have armed the quad once (to make sure you are not in lowpower disarm mode)

### Note:

For Smart audio, the power levels are hardcoded as:
1: 25mw
2: 200mw
3: 800mw

make sure you do not use more than your vtx allows.

I'm not sure but I do not think this applies to Tramp, I think there the levels are preset in the vtx itself

### DIFF
--- a/src/main/interface/cli.c
+++ b/src/main/interface/cli.c
@@ -1978,7 +1978,7 @@ static void cliFlashRead(char *cmdline) {
 #ifdef USE_VTX_CONTROL
 static void printVtx(uint8_t dumpMask, const vtxConfig_t *vtxConfig, const vtxConfig_t *vtxConfigDefault) {
     // print out vtx channel settings
-    const char *format = "vtx %u %u %u %u %u %u";
+    const char *format = "vtx %u %u %u %u %u %u %u";
     bool equalsDefault = false;
     for (uint32_t i = 0; i < MAX_CHANNEL_ACTIVATION_CONDITION_COUNT; i++) {
         const vtxChannelActivationCondition_t *cac = &vtxConfig->vtxChannelActivationConditions[i];
@@ -1990,6 +1990,7 @@ static void printVtx(uint8_t dumpMask, const vtxConfig_t *vtxConfig, const vtxCo
                                  cacDefault->auxChannelIndex,
                                  cacDefault->band,
                                  cacDefault->channel,
+                                 cacDefault->power,
                                  MODE_STEP_TO_CHANNEL_VALUE(cacDefault->range.startStep),
                                  MODE_STEP_TO_CHANNEL_VALUE(cacDefault->range.endStep)
                                 );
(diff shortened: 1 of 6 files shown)


Igorshp:
Added unit tests for the vtx cli command.

Tests helped find 2 issues:
- a segfault that can happen in nextArg() function if given empty string (happens if fewer parameters are passed in)
- more than 6 parameters could be passed to the command. Condition would still save but could be confusing to the user.

This test can be run by:
```
cd src/test && make test_cli_unittest
```

nerdCopter:
@Igorshp , does the unittests results convey that some code needs further work to prevent segfault (when use inputs invalid parameters)?
or is this PR 100% good?

Igorshp:
> @Igorshp , does the unittests results convey that some code needs further work to prevent segfault (when use inputs invalid parameters)?
> or is this PR 100% good?

Code is solid, however there is still at least one other unittest i'd like to add. 
Also when i submitted the segfault fix to betaflight, they raised an interesting point. Instead of fixing this function, it might be better to fix the instances where it is called and I kind of agree with that. So I will be moving that null check to another location.

Igorshp:
PR is ready from my side. 

Been using this myself for 2 weeks now, works great.

nerdCopter (approved):
Thank you for contributing to EmuFlight!!

nerdCopter:
The hardcoded power levels are from existing master? or something new? 
I did not see 25,200,800 in your code. I only saw `const char * const powerNames[4] ={ "---", "25", "100", "200"} ;`.

Igorshp:
the power levels are not part of the code this PR touches at all.
The ones you found are for a unit test and only used as a placeholder value.

As emuflight doesn't have vtx tables yet, the smart audio power levels are set here:
https://github.com/emuflight/EmuFlight/blob/master/src/main/io/vtx_smartaudio.c#L69

I don't have full understanding of smart audio protocol, but i believe it needs an actual power value (eg. 25,100,800) and Tramp takes a level ( e.g 1,2,3). If anyone has more knowledge on this, i'd love to hear it.

If you are using smartaudio, the levels you can set on cli are are:
0: no change: keeps whatever power level is already selected 
1: 25mw
2: 200mw
3: 500mw
4: 800mw

PR #418, opened 2020-12-04
Fix FrSky SPI receiver range and lockups

Igorshp:
Betaflight recently fixed a bug in frsky_d16 mode for SPI based recievers. The bug caused random drop out even in short range.

This is a port of 2 commits from betaflight:
https://github.com/mikeller/betaflight/commit/cb66ee09ba56a5f52f3355a404b43bbd3b7f0ec4

and one to fix lockups in frsky_x mode:
https://github.com/mikeller/betaflight/commit/77eb4f217272f9febeaa2dd5bec5b74060b407f7

All credit goes to @mikeller and @kevindehecker

Tested on Tinyhawk2 MATEKF411RX target.

### DIFF
--- a/src/main/rx/cc2500_frsky_d.c
+++ b/src/main/rx/cc2500_frsky_d.c
@@ -167,7 +167,7 @@ rx_spi_received_e frSkyDHandlePacket(uint8_t * const packet, uint8_t * const pro
     switch (*protocolState) {
     case STATE_STARTING:
         listLength = 47;
-        initialiseData(0);
+        initialiseData(false);
         *protocolState = STATE_UPDATE;
         nextChannel(1);
         cc2500Strobe(CC2500_SRX);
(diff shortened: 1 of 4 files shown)


nerdCopter:
* Flew 4 1S packs on NBDHMBF4PRO. No known issues.
* Flew farther than i have before just for testing. No ill effect, but i had none prior either.

Quick-Flash (approved):

cynfewl:
@Igorshp, @mikeller, @kevindehecker  this PR is amazing! Previously I was unable to use the built in iSRM module on my X-Lite Pro as it had problems with telemetry, regardless of whether telemetry was enabled or not. With this change I can now just fly my CrazyBeeF4FR whoops without having to use a MPM, which I had previously thought was only going to be possible in EmuFlight 1.0.
Thank you so much!

PR #431, opened 2020-12-08
FPORT fix: port of Betaflight https://github.com/betaflight/betaflight/pull/10388

Igorshp:
All credit goes to @tonycake for fixing this.

This is an awesome fix. I personally experience this bug on one of my quads with FPORT receiver (and MatekF722-STD FC).
Sometimes when plugging the quad in, telemetry does not come through. I have to turn it off and on again, even multiple times.

I've just tested this change and telemetry on FPORT is now rock solid. Comes through every single time the quad powers up. 🎉 

**Copy of message from betaflight pr:**

Fix the issue described in Issue #10220, typos in serial drivers for F7, F3, G4, H7 platforms.
Before this change, the serial line is incorrectly pulled down for bidirectional comms, which may cause dropped packets (silently) with any Rx protocol that uses bidirectional comms on the Tx line (GHST, SRXL-2, FPORT)

Issue this addresses: https://github.com/betaflight/betaflight/issues/10220


nerdCopter (approved):
LGTM

PR #436, opened 2020-12-12
NFE as a switch

Quick-Flash:
Adding NFE mode as a switch as well as making the logic for horizon mode better so that its harder for users to implement bad settings.

### DIFF
--- a/src/main/fc/config.c
+++ b/src/main/fc/config.c
@@ -236,19 +236,6 @@ static void validateAndFixConfig(void) {
            ) {
             rxConfigMutable()->rssi_src_frame_errors = false;
         }
-    if (!rcSmoothingIsEnabled() || rxConfig()->rcInterpolationChannels == INTERPOLATION_CHANNELS_T) {
-        for (unsigned i = 0; i < PID_PROFILE_COUNT; i++) {
-            pidProfilesMutable(i)->pid[PID_ROLL].F = 0;
-            pidProfilesMutable(i)->pid[PID_PITCH].F = 0;
-        }
-    }
-    if (!rcSmoothingIsEnabled() ||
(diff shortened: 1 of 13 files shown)


nerdCopter:
@Quick-Flash , can you explain in totality how this affects Configurator (pid-tab NFE)(modes-tab), and if MSP is required for `semver` checks and value setting?

Quick-Flash:
> @Quick-Flash , can you explain in totality how this affects Configurator (pid-tab NFE)(modes-tab), and if MSP is required for `semver` checks and value setting?

It just works in the configurator, you can flash yourself, the NFE button is now useless, and no MSP changes are needed besides removing NFE mode.

Quick-Flash:
> Interesting, so both ANGLE and NFE need to to be enabled in configurator?

currently, would it be better if enabling NFE on its own gave the angle mode effect on roll?

src/main/fc/config.c
```diff
@@ -236,19 +236,6 @@ static void validateAndFixConfig(void) {
...
-            pidProfilesMutable(i)->pid[PID_PITCH].F = 0;
-        }
-    }
-    if (!rcSmoothingIsEnabled() ||
-            (rxConfig()->rcInterpolationChannels != INTERPOLATION_CHANNELS_RPY &&
-             rxConfig()->rcInterpolationChannels != INTERPOLATION_CHANNELS_RPYT)) {
-        for (unsigned i = 0; i < PID_PROFILE_COUNT; i++) {
-            pidProfilesMutable(i)->pid[PID_YAW].F = 0;
-        }
-    }
```
Quick-Flash, line 251: i forgot to mention this but i removed more of the old FF code here

nerdCopter:
@nerdCopter reminder: to-do: if semver(firmware) <=0.3.2 {pid-tab} show NFE, else don't.

nerdCopter (approved):
LGTM

PR #437, opened 2020-12-12
add axis lock

Quick-Flash:
This pull request adds a new stick feel to the code. It allows you to enable an axis lock sort of feel. Essentially this code will disable the pidsums of axis which your stick is not moving. IE if you do a roll, yaw and pitch pidsums will be lowered drastically. In order to tune this there are two parameters axis_lock_hz, and axis_lock_multiplier. If the multiplier is set to 0 this is disabled. If the multiplier is higher it is enabled. Hz puts a filter over the stick movement. The lower the filter the longer it take for the effect to happen, but the longer the effect will last. If the hz value is high only during direct stick movement will you see any effect, as you set it lower the effect becomes more held out, and as such you can do quick rolls and ignore pitch and yaw for most of the roll.

LOOKING FOR FEEDBACK. This code is currently a part of the pidProfile, I'm thinking of moving it to the rate profiles and perhaps adding it as a mode that you can use a switch to flip into. Let me know your thoughts and I can code it so that it will work for any of those situtations.


gretel:
> This code is currently a part of the pidProfile, I'm thinking of moving it to the rate profiles and perhaps adding it as a mode that you can use a switch to flip into. Let me know your thoughts and I can code it so that it will work for any of those situtations.

thats true for all the previous modifications to the mixing. and i guess this is why it didnt get that much of attention cause its pretty cumbersome to test in the current way.

src/main/flight/pid.c
```diff
@@ -766,7 +789,8 @@ void pidController(const pidProfile_t *pidProfile, const rollAndPitchTrims_t *an
         pidData[axis].P = pidData[axis].P * getThrottlePAttenuation() * setPointPAttenuation[axis];
         pidData[axis].I = temporaryIterm[axis] * getThrottleIAttenuation() * setPointIAttenuation[axis]; // you can't use pidData[axis].I to calculate iterm or with tpa you get issues
         pidData[axis].D = pidData[axis].D * getThrottleDAttenuation() * setPointDAttenuation[axis];
-        const float pidSum = pidData[axis].P + pidData[axis].I + pidData[axis].D;
+        float pidSum = pidData[axis].P + pidData[axis].I + pidData[axis].D;
+        pidSum = pidSum * scaledAxisPid[axis];
```
skylupop, line 793: I'm trying to refactor this method to make it really easy to understand. Also there are a couple of things that can be optimized, e.g. the calculation of errorBoostAxis and errorLimitAxis (some operations are executed over and over for each pid loop and axis) and the if-else branch used for the nfe mode.

Can we rename scaledAxisPid -> axisLockMultiplier or something like that?

Quick-Flash: Error boost needs to be calculated per axis as the effect depends on how much error each individual axis has. Same as the limit.  Yes this can be named better, but the things you mentioned canr be changed to much except nfe mode could be made into a case statement.

tylercorleone: Of course I see the logic, I was talking of some multiplication that can be avoided to be repeated over and over simply using a couple of global variables. I'll show you how later. I was asking if you agree with the name I'm suggesting. I think that renaming "scaledAxisPid" into "axisLockMultiplier" could help to better understand what's the meaning of that variable without having to look at how it's used.

Quick-Flash: axisLockScale is probably a better name. axisLockMultiplier is already used to determine how strong of an effect this has, and i think i know what you mean by adding global variables, I think thats how i implemented things into 1.0.0, I mean I don't want to put a ton more work into maintaining this code base, this is kind of just a last hurra for the butterflight codebase.
I hope that makes sense.

tylercorleone: Yes absolutely makes sense!

nerdCopter (approved):

PR #438, opened 2020-12-13
add direct ff

Quick-Flash:
Yaw acts different than pitch and roll. This adds a feedforward based on setpoint. The more setpoint the higher the yaw ff, so just moving your setpoint will make an addition to the pidsum. The idea is that yaw always needs some motor output in order to create a yaw movement whereas pitch and roll just needs a push to start the move. The idea is to tune the true yaw FF so that it is pretty close to what the yaw needs to spin at that rate. This reduces the amount of iterm windup that happens with iterm. This allows yaw to act faster and have less iterm windup.

Also allowing for direct ff low and high for angle modes. Direct ff does work fine in angle mode as angle mode could fly with exclusively just direct ff and all pid gains set to 0, as such direct ff only works for pitch and roll while in angle mode and this only allows for that.


src/main/flight/pid.c
```diff
@@ -766,7 +780,7 @@ void pidController(const pidProfile_t *pidProfile, const rollAndPitchTrims_t *an
         pidData[axis].P = pidData[axis].P * getThrottlePAttenuation() * setPointPAttenuation[axis];
         pidData[axis].I = temporaryIterm[axis] * getThrottleIAttenuation() * setPointIAttenuation[axis]; // you can't use pidData[axis].I to calculate iterm or with tpa you get issues
         pidData[axis].D = pidData[axis].D * getThrottleDAttenuation() * setPointDAttenuation[axis];
-        const float pidSum = pidData[axis].P + pidData[axis].I + pidData[axis].D;
+        const float pidSum = pidData[axis].P + pidData[axis].I + pidData[axis].D + yawFeedForward;
```
tylercorleone, line 783: I don't like the fact that pidSum isn't the P-I-D anymore, actually... Also it's a little confusing seeing a *yaw*FeedForward being added when the axis is not yaw (yes I know it would be 0 i  that case). There isn't a way to obtain the same result in a more elegant way? E.g. adding the component to some other already existing one?

Quick-Flash: PIDF is common for a pidsum. bf does that as well. I wouldnt want to add this to p I or d as it would ruin logging them. Perhaps I can put the yaw check right at the pidsum instead though so it's a little easier to read.

src/main/interface/settings.c
```diff
@@ -909,6 +909,7 @@ const clivalue_t valueTable[] = {
     { "p_yaw",                      VAR_UINT8  | PROFILE_VALUE, .config.minmax = { 0, 200 }, PG_PID_PROFILE, offsetof(pidProfile_t, pid[PID_YAW].P) },
     { "i_yaw",                      VAR_UINT8  | PROFILE_VALUE, .config.minmax = { 0, 200 }, PG_PID_PROFILE, offsetof(pidProfile_t, pid[PID_YAW].I) },
     { "d_yaw",                      VAR_UINT8  | PROFILE_VALUE, .config.minmax = { 0, 200 }, PG_PID_PROFILE, offsetof(pidProfile_t, pid[PID_YAW].D) },
+    { "true_f_yaw",                 VAR_UINT8  | PROFILE_VALUE, .config.minmax = { 0, 200 }, PG_PID_PROFILE, offsetof(pidProfile_t, yaw_true_ff) },
```
skylupop, line 912:
```suggestion
    { "yaw_true_ff",                 VAR_UINT8  | PROFILE_VALUE, .config.minmax = { 0, 200 }, PG_PID_PROFILE, offsetof(pidProfile_t, yaw_true_ff) },
```

skylupop: Oh... now I understand what you wanted to do, recall the "i_yaw", "d_yaw" etc... but could we just call it ff_yaw or f_yaw and add a pid[PID_YAW].F or pid[PID_YAW].FF field? I didn't find a way to simulate BF's FF yet. Thinking about future implementations so we would avoid breaking changes in naming

Quick-Flash: I suppose that we still have a pid.f in the code that this could easily become a part of. The downside to doing it that way is that this codebase is being ditched soon in favour of 1.0.0 which is built of off bf 4.3, in that case this would be conflicting with the normal FF which bf has. So just some simple way we can implement this is totally fine by me. Giving it a better name i'm find with as well, it just needs to have a distinction from the FF which betaflight uses which is the derivative of setpoint while this is proportional to the setpoint. So it'll be tuned differently.

nerdCopter:
@Quick-Flash , needs consistent variable naming across new items for the CLI `get`.
please choose one:
* `direct_f_`  for ALL
* `df_` for ALL
* `dff_` for ALL
* `direct_ff_` for ALL

nerdCopter (approved):
trusting that all the features richness is worth it.  hope i dont have to GUI all this, lmao. (because MSP && pilot overload)

nerdCopter:
* up for debate, but seems default `directFF_yaw = 15` was more widely accepted.

nerdCopter (dismissed):

Quick-Flash:
@nerdCopter can you approve this?

nerdCopter (approved):

Issue #445, opened 2020-12-19
Crossfire telemetry reports pack voltage not cell voltage


sweebee:
**Describe the bug**
I've set ```report_cell_voltage = ON``` but I still shows the full voltage of the pack, not an avarage cell voltage. This works fine in Betaflight.

**To Reproduce**
use crossfire, it repots rxbat as 15.2v instead of 3.8v.

**Expected behavior**
Show cell voltage

**Flight controller configuration**
```
# diff all

# version
# EmuFlight / FURYF4OSD (FY4O) 0.3.2 Oct  7 2020 / 19:11:01 (e0270d5fd) MSP API: 1.49

board_name 
manufacturer_id 
mcu_id 001c002c394e500220343942
signature 

# reset configuration to default settings
defaults nosave

# name
name Glide

# resources

# mixer

# servo

# servo mix


# feature
feature TELEMETRY

# beeper

# beacon
beacon RX_LOST
beacon RX_SET

# map
map TAER1234

# serial
serial 0 0 115200 57600 0 115200
serial 2 1 115200 57600 0 115200
serial 5 64 115200 57600 0 115200

# led

# color

# mode_color

# aux
aux 0 0 0 1700 2100 0 0
aux 1 2 1 975 1650 0 0
aux 2 13 1 1850 1950 0 0
aux 3 13 1 1650 1725 0 0
aux 4 13 1 1475 1525 0 0
aux 5 35 1 1625 1850 0 0
aux 6 35 1 1275 1450 0 0

# adjrange

# rxrange

# vtx

# rxfail
rxfail 5 s 950
rxfail 7 s 1000

# master
set gyro_lowpass_hz_roll = 130
set gyro_lowpass_hz_pitch = 130
set gyro_lowpass_hz_yaw = 130
set imuf_roll_q = 6200
set imuf_pitch_q = 6200
set imuf_yaw_q = 6200
set imuf_w = 16
set imuf_sharpness = 1
set acc_calibration = -37,-38,-79
set min_check = 1020
set max_check = 1980
set rssi_channel = 8
set fpv_mix_degrees = 25
set serialrx_provider = CRSF
set blackbox_device = NONE
set bat_capacity = 1500
set current_meter = ADC
set ibata_scale = 275
set ibatv_scale = 275
set yaw_motors_reversed = ON
set deadband = 5
set yaw_deadband = 5
set report_cell_voltage = ON
set osd_warn_arming_disable = ON
[...]

PR #446, opened 2020-12-19
Avarage cell voltage for CSRF

sweebee:
```report_cell_voltage``` is not used currently for crossfire. This request fixes it.

Issue: #445


nerdCopter (approved):

nerdCopter (commented):
Can you please validate no div zero with `/=` ? Or should it be constrained somehow?

sweebee:
I get no issues when no lipo is plugged in. I just looked ad the ibus telemetry and its doing it the same way. 

https://github.com/emuflight/EmuFlight/blob/master/src/main/telemetry/ibus_shared.c#L189

PR #448, opened 2020-12-22
Add cliDebugPrint functions to facilitate easy debug printing to CLI

Igorshp:
port of https://github.com/betaflight/betaflight/pull/8905

thanks to @etracer65 for original commit in betalfight

### From linked PR:
To use, include cli/cli_debug_print.h in your code and be sure USE_CLI_DEBUG_PRINT is defined. Or alternately, build using make TARGET=<target_name> OPTIONS="USE_CLI_DEBUG_PRINT". Then you'll have access to the following functions to print debugging messages in the CLI:
```
cliDebugPrintLineFeed
cliDebugPrint
cliDebugPrintLine
cliDebugPrintf
cliDebugPrintLinef
```

Output is suppressed when the CLI is not open. The functions are equivalent to their cliPrint... counterparts.

May interfere with the autocomplete initialization when first entering the CLI if your code is outputting data when the CLI first opens. But as this is only meant for debugging it shouldn't be much of a concern.

You may also need to rate limit your messages if printing data in a loop.

Using these calls in other modules should only used for debugging and not left in any finished code.

Co-authored-by: etracer65 <github@etracer.net>


Igorshp:
This proved quite handly debugging redpipeline code.  I still wish there was a better way to look inside the brains of FC

Quick-Flash:
so how does this exactly work?

nerdCopter:
if this is still helpful/required, please update with master and we can merge.

nerdCopter (approved):

nerdCopter:
merging it.  if it's not manually added to code-files at compile time, it changes nothing.

PR #451, opened 2020-12-29
EmuGravity

Quick-Flash:
Antigravity but different. Me thinks this will work better.

Just remembered i had written this months ago, really unsure if it will make it into the 1.0.0 branch, but it sure is way simpler than the bf implementation of antigravity, and i do think it may have some promise. This just boosts the error when you are changing the throttle. Perhaps a better solution is will look at the motor output command to achieve a better solution to this, but this may work for now, I'll need to rebase this as its horribly old and stale.


nerdCopter:
`float motorMix[MAX_SUPPORTED_MOTORS];` in there twice:
```
./src/main/flight/mixer.c: In function 'mixTable':
./src/main/flight/mixer.c:861:11: error: redeclaration of 'motorMix' with no linkage
  861 |     float motorMix[MAX_SUPPORTED_MOTORS];
      |           ^~~~~~~~
./src/main/flight/mixer.c:829:11: note: previous declaration of 'motorMix' was here
  829 |     float motorMix[MAX_SUPPORTED_MOTORS];
      |           ^~~~~~~~
```

nerdCopter:
* will math be modded?
* will type be modded for uint8 (0-255)? or keep uint16 (0-65535)
* default value?  (4k was too much, whereas 800-1000 seemed viable)

nerdCopter (dismissed):

nerdCopter (approved):

PR #467, opened 2021-01-11
ABG Filter

Quick-Flash:
This code is untested but ports over the ABG filtering that has been tested in version 1.0.0! Haven't had a time to flash and test, but things look and compile fine, be careful on first test, but should be exciting to see how this can help for 0.4.0.

Settings:
dterm_abg_alpha // sets an alpha value as though it was a pt1 filter makes the filter act the same regardless of pidloop
dterm_abg_boost // adds a boost to the error seen inside the abg filter
dterm_abg_half_life // how long in ms * 10 the abg filter looks at memory. example, half life of 50 means after .5 seconds the abg has forgotten 50% of the data that is .5 seconds out and forgotten 75% of the data 1 second out, etc...

gyro_abg_alpha
gyro_abg_boost
gyro_abg_half_life


nerdCopter (dismissed):

nerdCopter:
why would one want half-life of 100 seconds? (HL Range 0-10000)

nerdCopter (dismissed):

Quick-Flash:
> why would one want half-life of 100 seconds? (HL Range 0-10000)

i changed the max. Yeah no reason for that high of a half life

nerdCopter (dismissed):

nerdCopter (approved):

PR #500, opened 2021-02-24
update the Matrix filter, turn it into a dynamic notch filter

Quick-Flash:
This filter is far more precise than the previous filter was. I've currently set this up with 36 bins (so accuracy is (max-min)/36 which for defaults is, 12.5hz, the old filter had a bin size of 16 which would have given an accuracy of 28.125hz, so this is over double as accurate.) This can also track up to 5 peaks and place 5 notch filters. It is also nicer on the cpu than the older version was, so all around really good. Using this filter + abg i was easily able to raise my gyro lpf filters to 300hz with dterm filters at around 150hz. I'm really excited to see what you all think about this filter :), P.S. all credit goes to @KarateBrot for coding this into betaflight, I may want to update this later to have dynamic widths and dynamic on the number of notches it uses, but so far, hey its awesome. Probably going to stay like this for 0.4.0. Only the defaults need testing right now.

For more info check out https://github.com/betaflight/betaflight/pull/10554


Quick-Flash:
should be good for some hard testing

nerdCopter:
requires #638

nerdCopter:
* would prefer codacy cleanup
https://github.com/emuflight/EmuFlight/pull/500/checks?check_run_id=3127643577

nerdCopter:
* request #657 merged into this

nerdCopter (approved):
🚀

PR #502, opened 2021-02-25
fix yaw from causing motor increase. fix in 2pass mode skipping smooth mode for now.

Quick-Flash:
Try flying the legacy mixer, and do some los yaws, notice what they act like. Now set the mixer to smooth. Notice how yaw jumps the drone up in the air. Now with this code set the mixer to 2pass and notice how this has now fixed all the yaw throttle issues and that it acts like legacy. This has not been added to the smooth mode (not sure why you wouldn't want to use 2pass as its awesome) as smooth handles yaw the same as pitch and roll. I'd love to see some testing of this code on 2pass verses the previous version on 2pass. I do prefer the way yaw handles in this version. Let me know if you agree.


tylercorleone:
If 2PASS handles well yaw spin, why are you changing 2PASS instead of smooth mode?

src/main/flight/mixer.c
```diff
@@ -1006,7 +1009,16 @@ static void twoPassMix(float *motorMix, const float *yawMix, const float *rollPi
                                         : SCALE_UNITARY_RANGE(throttleMotor, -yawMixMin, -yawMixMax);
 
         motorMix[i] += (yawMix[i] + yawOffset) * controllerMixNormFactor; // yaw is an output-proportional value (RPM-proportional, actually)
+        postYawThrottle += motorMix[i];
+    }
+    // find the extra thrust that yaw is adding
+    postYawThrottle = postYawThrottle / motorCount;
+    yawThrottleCorrection = postYawThrottle - throttleMotor;
```
tylercorleone, line 1016: yawThrottleCorrection = postYawThrottle - throttleMotor = postYawThrottle / motorCount - throttleMotor = (K + throttleMotor * motorCount) / motorCount - throttleMotor = K / motorCount + throttleMotor - throttleMotor = K / motorCount

Where K is the yaw component average (that should be 0).

What am I missing?

Quick-Flash: K is not the yaw component average. K is motorMix[i] which at this point is the average of throttleMotor + (yawMix[i] + yawOffset) * controllerMixNormFactor. If you look at the bf pull request for your mixer modes you will see that the reason that this mixer is more smooth is because it moves the drones total throttle in order to minimize the total thrust changes to the motor. Just give this code a try yourself LOS and you'll quickly see a difference. Even on the current master branch try flying in legacy mode then try smooth or 2pass. You'll notice yaw is way different between the 2. Just hover it and yaw and notice what happens. Then try this code. I could even log this to just see how the yawThrottleCorrection changes over time.

Quick-Flash:
> If 2PASS handles well yaw spin, why are you changing 2PASS instead of smooth mode?

Both 2pass and smooth will cause throttle jumps due to yaw. The idea of 2pass is to seperate yaw from pitch and roll more so it made more sense to add this to 2pass. In this PR you wont see this in 2pass as I've reduced the throttle change from yaw. Give it a test I'm really curious what you think of it as you originally came up with these mixers and did some real good work.

tylercorleone:
> > If 2PASS handles well yaw spin, why are you changing 2PASS instead of smooth mode?
> 
> Sorry if my title was confusing. What I meant was fix this issue for 2pass mode. Both smooth and 2pass show it.
> 
> Both 2pass and smooth will cause throttle jumps due to yaw. The idea of 2pass is to seperate yaw from pitch and roll more so it made more sense to add this to 2pass. In this PR you wont see this in 2pass as I've reduced the throttle change from yaw. Give it a test I'm really curious what you think of it as you originally came up with these mixers and did some real good work.

Ok, now I understood what you mean. I'll definitely give it a try!

nerdCopter (approved):
* conceding for lack of alternative solution. :D

nerdCopter:
for records: JoshG does not like it, DrSchnick does not like it, and finally i'm finding delay in yaw as well.
delayed feedback hinders development, but at least we receiving some.

Quick-Flash:
hey hey hey, i think i've fixed this :)

Quick-Flash:
a little bit more testing and this should be good to go.

nerdCopter:
i'm flying it today. guess i should A/B test.

nerdCopter:
tested good AFAICT.  logs sent for review. awaiting more feedback as well.

Quick-Flash:
> tested good AFAICT. logs sent for review. awaiting more feedback as well.

I finally fixed them bugs

nerdCopter:
```
[May 20] Josh G: i have a hard time getting generic 2pass to rise consistently, sometimes it goes up a tad, sometimes it doesn't. seems like certain deg/s trigger it. (maybe I'm just correcting with throttle/muscle memory) I will say, that yaw comp = ON feels a lot more consistent across the whole yaw range
```

```
[May 23] nerdCopter: [ XL7, in the moment] yaw throttle balanced (unless placebo)
[Screech6, in the moment]  unsure about yaw throttle ballooning or not (seemingly need to add throttle in some occasion)
[May 28] [retrospect after HD review] yes works. need throttle on occasion. can still minor balloon, but that is likely normal yaw effect plus stick position at the moment.
```

```
[May 28] DrSchnickistein: Yaw fix working as intended
It's not wandering and coordinated rolls look normal
Altitude dip on turns needs some getting used to though. I guess that's the only real giveaway on which is which
```

nerdCopter (approved):
* looking good, please read last comment.

PR #504, opened 2021-02-26
clean up osdWarnDjiEnabled - disable usbVcpIsConnected() check for dj…

RiskoFPV:
This fix allowed TRANSTECF7 to work with dji warnings. 

The usbVcpIsConnected() check was preventing it from working even when usb was disconnected. Credit @nerdCopter for suggesting this. I do not see a reason why DJI warnings should be disabled when usb is connected, you are able to see analog warnings when usb is connected.

The check was done by a boolean function, that was removed to save memory and now it just tests if the osd setting is enabled.

These targets were known to have the issue: 

MatekfF722
mamba f7 mini
HGLRCF722
TRANSTECF7

### DIFF
--- a/src/main/interface/msp.c
+++ b/src/main/interface/msp.c
@@ -766,9 +766,9 @@ bool mspProcessOutCommand(uint8_t cmdMSP, sbuf_t *dst) {
     }
     break;
     case MSP_NAME: {
-      // Show warning for DJI OSD instead of pilot name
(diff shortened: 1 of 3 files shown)


Quick-Flash:
I see no problems with removing the check to see if its connected to the pc. :) thanks for doing this @RiskoFPV

Quick-Flash (approved):

PR #514, opened 2021-03-10
Fix CPU status, UART Listing, and Add calibrate accelerometer CLI command

RiskoFPV:
command nemesis_status now reports CPU % as the same as the max load listed in the tasks command
UART list in config command now shows all available UARTs correctly (credit to QuickFlash for the fix)
new cli command to calibrate the accelerometer, to allow Nemesis to calibrate


Quick-Flash:
hmm i'm beginning to rethink this method of doing nemesis

PR #529, opened 2021-04-02
Fix the XILOF4 NO_GYRO problem.

loos-br:
Enable the SPI gyro and fix the XILOF4 target.

### DIFF
--- a/src/main/target/XILOF4/target.h
+++ b/src/main/target/XILOF4/target.h
@@ -42,16 +42,17 @@
 
 #define USE_GYRO
 #define USE_GYRO_SPI_MPU6000
-//#define USE_GYRO_SPI_MPU6500
+#define USE_GYRO_SPI_MPU6500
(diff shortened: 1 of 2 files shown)


Quick-Flash (approved):

nerdCopter:
there are different versions.
maybe if we merge this, then old ones or new ones break?
https://www.getfpv.com/xilo-stax-f4-flight-controller.html :
```
Version 1.2 update (March 2020)

    Changed ICM20602 gyro to MPU6000 gyro for less sensitivity to noise and vibrations

NOTE: The newest version (board sold after Feb. 2020) of the XILO STAX FC with the MPU6000 Gyro have custom firmware  for this board with BF 4.1.3 pre-loaded. Do no flash anything before 4.1.5 on this FC or it will not work.
```

nerdCopter:
@Kaiowarez  said this will work on both new and old boards. Thanks!

PR #531, opened 2021-04-06
increase auto detect cell count to 12

JulioCesarMatias:


JulioCesarMatias:
increase `#define VBAT_CELL_VOTAGE_RANGE_MAX 50` to 55 is safe?

Quick-Flash:
> increase `#define VBAT_CELL_VOTAGE_RANGE_MAX 50` to 55 is safe?

Depends on what you mean by safe. The code will safely use it, but no lipo should be charged much more than 4.2 volts per cell. 5.5 or even 5.0 volts per cell is to high. I remember some past discussion in betaflight about allowing higher auto cell counts. It just begins to get real tricky. For instance a 12s at full charge is 50.4v and discharged to say 4v per cell it now sits at 48v. If you've set your highest cell voltage to even say 4.4 (default I think) that is going to show up as a 11s battery (4.4*11=48.4). So we may want to come up with better ways to deal with auto cell voltage or perhaps leave the high cell counts to non auto modes. 

Do you use 12s batteries?

JulioCesarMatias:
> > aumentar `#define VBAT_CELL_VOTAGE_RANGE_MAX 50`para 55 é seguro?
> 
> Depende do que você entende por seguro. O código irá usá-lo com segurança, mas nenhum lipo deve ser cobrado muito mais do que 4,2 volts por célula. 5,5 ou mesmo 5,0 volts por célula é muito alto. Lembro-me de uma discussão anterior no betaflight sobre permitir contagens de células automáticas mais altas. Simplesmente começa a ficar realmente complicado. Por exemplo, um 12s com carga total é de 50,4 V e descarregado para dizer 4 V por célula, ele agora fica a 48 V. Se você configurou a tensão mais alta de sua célula para até mesmo 4,4 (padrão, eu acho), isso vai aparecer como uma bateria 11s (4,4 * 11 = 48,4). Portanto, podemos querer encontrar maneiras melhores de lidar com a voltagem automática da célula ou talvez deixar as contagens altas de células em modos não automáticos.
> 
> Você usa baterias 12s?

I had deleted the question, because afterwards I realized that 50 is 50/10 = 5, which refers to the maximum voltage of each cell, at first I thought it was the maximum battery voltage, so I asked if it was safe increase to 50 because the charged 12s battery is around 50.4v (in 4.2 per cell). I don't use 12s battery, but this can be useful for the future.

JulioCesarMatias:
if that was not useful, can i close the PR?

JulioCesarMatias:
@Quick-Flash  @nerdCopter  would it be cool to bring the launch-control (race start assistance system) from betaflight to emuflight?

Quick-Flash:
> @Quick-Flash  @nerdCopter  would it be cool to bring the launch-control (race start assistance system) from betaflight to emuflight?

If you would like to port it. I tried a long while ago and never finished

Quick-Flash:
> @Quick-Flash , i don't foresee a problem with merging this, do you?

That depends. We already have issues with hv 1s batteries thinking they are 2s. Just imagine how bad our code is going to be at getting a cell count for a 12s battery.

Quick-Flash (approved):

PR #534, opened 2021-04-13
Simplify the Rate Dynamics processing

DzikuVx:
`dT` computation for Rate Dynamincs is performed for each axis multiple time while it can be executed once and just passed below. 
Also, the  `rateDynamics` function can be made static and not present in header file


Quick-Flash:
Thanks for fixing a bit of my dodgy coding :)

Quick-Flash (approved):

nerdCopter (approved):

PR #540, opened 2021-04-14
Updates to attitude and angle mode (AKA Angle mode 2.0)

Quick-Flash:
These changes should be considered a big enough change to be considered an angle mode 2.0.

Improvements to how attitude is calculated. Roll is now calculated differently for angle mode and should deal with singularities at pitch +-90 better. Should be a smoother angle all around as roll will play nicer. Will need to do something to horizon mode as well, but that might be a bit trickier.

This will need testing by some of our super angle mode pilots.


Quick-Flash:
This code is all likely good, just more tests with NFE mode and Horizon mode are needed.

nerdCopter (approved):
* approved based on QF's feedback

PR #541, opened 2021-04-15
update the kalman filter

Quick-Flash:
Remove sharpness and update the kalman filter. Needs testing, but I've tested something very similar in 1.0.0 that works well. Q has a whole lot more effect as well as w. Will need testing feedback, but the Kalman filter will have a lot more of an effect now.


Quick-Flash:
i'd love for this to get tested :)

nerdCopter (approved):
* certainly flies.
* BBL should be A/B compared.
* @Vitalis BBL available in discord.  requested review.
* @Vitalis opinion: it was better, but BBL is best analysis

> [Jun 16 6:42 AM] Vitales: Now I flew in 1824, I will say this, I liked 1824 more than 1822, from the improvements I noticed that the motors are colder, then the sound of the motors is softer, the flight is softer, and in my opinion the flight time also increased slightly.  In short, 1824 is better firmware.

nerdCopter:
* so no PT(n) on IMUF like planned for Helio? @Quick-Flash

Quick-Flash:
> * so no PT(n) on IMUF like planned for Helio? @Quick-Flash

the imuf contains 2 filters, a kalman filter, and a lowpass filter (in this case now a ptn filter). So this is the same as helio, we have a kalman filter, and then when you choose your gyro lpf to be a ptn filter it is the same as what helio has. That is why you don't see helio fc using gyro filters. Because it already has the kalman and the ptn.

PR #544, opened 2021-04-19
bump version 0.3.5

nerdCopter:
* i think it appropriate to bump version to 0.3.4 since we released 0.3.4
* should avoid confusion for next test-builds


Quick-Flash (approved):

nerdCopter:
* i'd like to wait to see if `yaw-throttle-correction` (#502) gets merged along with potential on/off toggle.
* also would like blackbox headers for ABG first as well.

PR #548, opened 2021-04-24
toggle for yaw-throttle-correction (compensation) PR#502

nerdCopter:
* needs code review for if a good `mixer.c` logic or not
* enabled by default.
* OSD PID>Advanced>`MIXER YAW THR COMP`.
* CLI `set mixer_yaw_throttle_comp = `.
* includes blackbox header.
* test hovered good on/off.
* OSD works
* CLI works


Quick-Flash (commented):
yep this will totally work :)

PR #550, opened 2021-04-27
Fix strength computation

DzikuVx:
The Smith Compensator strength computation was broken and resulted in `0`


Quick-Flash:
already fixed this, but still good catch.

PR #553, opened 2021-05-03
Ehh its just smith predictor

Quick-Flash:
Heres what @DzikuVx has to say about the smith predictor.

> tests show that it indeed works and the compensated signal is indeed around 4ms in front of the not-compensated one.
> 
> Subjective flight results:
> 
> * no change in smooths flight
> * seems to reduce propwash quite noticeably
> * fast turns to feel more controllable and locked-in (I know it's a abused phrase). Quad felt more on the rails
> 
> Cons: when working with a noisy gyro, can amplify noise. In those tests, both gyro LPF was off, only Matrix and Unicorn Filter were enabled

P.S. this code is not ported from inav, its ported from emu 1.0.0


nerdCopter (approved):
* approved based on tester/QF testing/discussions

nerdCopter:
some reference material:
* https://github.com/d-ronin/dRonin/pull/2044
* https://www.controleng.com/articles/overcoming-process-deadtime-with-a-smith-predictor/

PR #556, opened 2021-05-07
narrower matrix Q (+100 Q)

nerdCopter:
~~* helps speedup the notch calculations, less delay~~
* narrower notch, reduces delay slightly


Quick-Flash:
Doesnt speed up notch calculations, lol. It just makes the notch narrower. LGTM

Quick-Flash (approved):

PR #557, opened 2021-05-07
adjust range for matrix max hz 400-1000.

nerdCopter:
* allow for lower max for bigger craft with lower noise profiles
* does not modify default value of 600


Quick-Flash (commented):
LGTM

Quick-Flash (approved):

PR #558, opened 2021-05-07
raise default dTermLPF1 to 90

nerdCopter:
* raise dTermLPF1 to 90 to reduce over-filtered defaults.
* as the original lowering to 65 was debatable, so is raising it to 90.


Quick-Flash (approved):
I do like this. I'd also like to see more testing with these defaults before we have our next release, but overall I like this.

PR #559, opened 2021-05-08
resolve #555; DJIwarning only if OSD defined

### LINKED ISSUE #555: OSD dependency in the MSP module (compiling without OSD not possible)
**Describe the bug**
As discussed on discord, there is an OSD depency in the MSP module that prevents to compile emuflight without the OSD feature.

**To Reproduce**
add to \src\main\target\<TARGET>\target.h
> #undef USE_OSD 

**Expected behavior**
Compile error on MSP.c

>         if (osdWarnGetState(OSD_WARNING_DJI)) {
>             sbufWriteString(dst, djiWarningBuffer);
>             break;
>         }

I managed to make it do with the following hack:

> **#ifdef USE_OSD**
>         if (osdWarnGetState(OSD_WARNING_DJI)) {
>             sbufWriteString(dst, djiWarningBuffer);
>             break;
>         }
> **#endif**	

**Additional context**
My old DTFc / DOGE F3 flight controller does not have OSD. Current Master does fit nicely in the memory once OSD is removed.

nerdCopter:
* fix compile for FC's without OSD.
* solution for #555 as provided by @raemin .


src/main/interface/msp.c
```diff
@@ -768,10 +768,12 @@ bool mspProcessOutCommand(uint8_t cmdMSP, sbuf_t *dst) {
     case MSP_NAME: {
         // Show warning for DJI OSD instead of pilot name
         // works if osd warnings enabled, osd_warn_dji is on and usb is not connected
```
RiskoFPV, line 770: It no longer cares if USB connected. suggest change comment to // works if osd warnings enabled, and osd_warn_dji is on

src/main/interface/msp.c
```diff
@@ -768,10 +768,12 @@ bool mspProcessOutCommand(uint8_t cmdMSP, sbuf_t *dst) {
...
+        #ifdef USE_OSD
```
RiskoFPV, line 771: This should be okay, however, if the FC does not have an analog OSD chip, it still technically could perform DJI OSD functions. But DJI custom osd would not work without enabling USE_OSD. I feel like theres a small user base that would lose DJI OSD on these FCs

RiskoFPV (changes requested):
Comment fix and opinionated rant about dji osd

src/main/interface/msp.c
nerdCopter: interesting. i don't know how to resolve such.

PR #561, opened 2021-05-10
smaller default emuGravity value [`emu_gravity`]

nerdCopter:
* emugravity default 50 rather than 100.
* open to any values suggested by dev-team & test-team


nerdCopter:
* i've been flying my two quads set = `50`, fwiw.
* 50 was chosen as fair value after reading some test-pilot feedback.
* Kumo likes 40, DrSnick setting even lower on ABG/Smith testing.

Quick-Flash (approved):

PR #562, opened 2021-05-11
add pt2-pt4 no longer suggest users use BIQUAD

Quick-Flash:
Add ptn filter and no longer recommend users use the biquad filter as gyro and dterm filter options. Add filter type to the osd menus.

@AndreySemjonov has been asking for this so here you go!

For more indepth discussion regarding the pt2 and pt3 please refer to this BF PR https://github.com/betaflight/betaflight/pull/10727 however, this code has been implemented in EmuFlight 1.0.0 before BF made this pull request. This just happens to coincide and have lots of useful information.


nerdCopter:
needs alienwhoop target fix

Quick-Flash:
> removing biquad from configurator not possible in totality for backward compat. gonna be a mess for emuConfig: programmatically adding and removing biquad to list every time an fc is plugged in... assuming if that's even possible.

could it not be possible to just make the filter types list dynamic in the config? and have it change when we have 0.4.0 or newer flashed?

nerdCopter:
BEEBRAIN_BL Error

Quick-Flash:
> BEEBRAIN_BL Error

fixed this

nerdCopter (approved):
* booooom

nerdCopter:
* i'm thinking this is good after a few hovers, BBL, and conversation.  
* keeping BiQuad is good for pilots to refrain from needing immediate retunes.
* Configurator mod was easy.

nerdCopter:
* I can't get acceptable motor temperatures from PT(n)'s on dTerm until D's are very low.  Went 16 D roll on PT3, but i dont know if this is side affect of this PR or from new IMUF 254, 255.  if side affect from updatedKalman, then non-Helio's will be affected by https://github.com/emuflight/EmuFlight/pull/541 as well.  Unclear to me the source of high-motor temps, atm. BiQuad seemed high-temp as well, but not intolerable.  91F weather at time of testing did not help. 
* OSD only allowing PT1 and BiQuad. shows PT1 if higher PT's are saved via Configurator.

nerdCopter:
* it occurs to me feathered=0 may be a factor, i can try feathered=50.
* not "broken", fwiw. willing to merge after OSD fixed.

nerdCopter:
* do we want PT2 as default?

nerdCopter:
* redacting/editing last comment.
* still questioning which dterm lowpass should be default?? biquad is "safest", but seemingly either PT2 or PT3 is better than current PT1.
* found that pid.c:134 `.dterm_filter_type = `, does indeed set default properly.

PR #566, opened 2021-05-14
Change VTX powers labels

nerdCopter:
* this is @madChiller's branch


RiskoFPV (commented):
Does the capitalization matter? is osd all uppercase anyway? do we want to make it more uniform inside the code?

Quick-Flash:
> Does the capitalization matter? is osd all uppercase anyway? do we want to make it more uniform inside the code?

Yes capitalizing matters. All that the osd chip stores is capital letters. Lowercase letters do not work and display garbage.

Quick-Flash:
this works, but do we want to do a dynamic implementation?

RiskoFPV:
Tested and it shows LV1, LV2.. etc in the OSD.
VTX channel OSD element still shows single number, like: A:1:2 for power level 2

Should be good to go

Quick-Flash (approved):

PR #567, opened 2021-05-16
update the abg filter (filter on abg velocity, acceleration, jerk)

Quick-Flash:
Make the ABG filter filter the velocity, acceleration and jerk. worth a good test. Currently both the acceleration and jerk have a noise floor that is ridiculous. So much noise on all frequencies. This aims to filter the velocity jerk and acceleration to help remove some of the noise that they have and perhaps make the jerk and acceleration give us better data out of them. This is for sure worth testing to see if slowing down the acceleration and jerk (via filtering) is worth doing or if we should just leave the raw noise as it is.


src/main/common/filter.h
```diff
@@ -46,7 +46,7 @@ typedef struct alphaBetaGammaFilter_s {
     float ak, vk, xk, jk, rk;
     float dT, dT2, dT3;
     float halfLife, boost;
-    pt1Filter_t boostFilter;
+    pt1Filter_t boostFilter, velFilter, accFilter, jerkFilter;
```
JulioCesarMatias, line 49: `boostFilter` is declared but not used

Quick-Flash: ha, dang your right.

src/main/common/filter.c
```diff
@@ -237,7 +237,7 @@ FAST_CODE float alphaBetaGammaApply(alphaBetaGammaFilter_t *filter, float input)
   rk = input - filter->xk;
   
   // artificially boost the error to increase the response of the filter
-  rk += (fabsf(rk) * rk * filter->boost);
+  rk += pt1FilterApply(&filter->velFilter, (fabsf(rk) * rk * filter->boost));
```
JulioCesarMatias, line 240: you added the `velFilter` pointer,the correct is ` boostFilter`

nerdCopter (approved):

nerdCopter:
i'm setting this draft, only because i cannot figure out how to revoke my mistaken approval.

PR #570, opened 2021-05-19
Fix Barometer pressure calculation

JulioCesarMatias:
1 - A setting of 1 would result in division by zero in the code. A setting of 0 would result in division by -1 causing altitude measurements to be inverted.

2 -  Also fixes the problem that changing `baro_tab_size` to a smaller value could result in a buffer overflow due to a check for equality with the buffer limit.


nerdCopter:
Test build 1739 and 1740 contain this.
https://github.com/emuflight/dev-unstable/releases/tag/hex-1739
https://github.com/emuflight/dev-unstable/releases/tag/hex-1740

i requested testing, but no feedback.

Quick-Flash (approved):

PR #571, opened 2021-05-19
Prevent a division of zero in the battery current measure

JulioCesarMatias:
Prevents division by zero if current meter scale is set to 0 (zero).Can't prevent with parameter ranges because scale ranges from -16000 to 16000.


nerdCopter:
Test build 1739 and 1740 contain this.
https://github.com/emuflight/dev-unstable/releases/tag/hex-1739
https://github.com/emuflight/dev-unstable/releases/tag/hex-1740

I flew 6 packs containing this, without issue, but do not know any specific tests.

Quick-Flash (approved):

PR #574, opened 2021-05-21
Keeping up with the joneses

Quick-Flash:
Adding all the latest BF commits. All credit to those who committed the BF code. 

Missing some dynamic notch code, need to reevaluate if I had added the dynamic notch correctly the first time. Perhaps the default width was just far to small.


DusKing1:
The title, joneses, I can't say that that's a nice word, if you don't have resistance to betaflight could you please use some nice words? Like neighbor? No offense.

nerdCopter:
* obsolete first rebase

PR #579, opened 2021-05-22
Small filter efficiency improvements

Quick-Flash:
* Should reduce a small amount of the cpu load associated with the dynamic notch.
* thanks goes to @karatebrot


nerdCopter:
* this is applied where? dTerm LPF BiQuad?
* how much testing does this need?

Quick-Flash:
> * this is applied where? dTerm LPF BiQuad?
> * how much testing does this need?

Any biquad or notch. Just need to show that it flies, so let's get this merged.

nerdCopter (approved):
* it flies

PR #583, opened 2021-05-24
Smith-Predictor on/off toggle [smith_predict_enabled]

Quick-Flash:
Enable/disable options for smith predictor. I'll let others deal with the MSP side of things.


Quick-Flash:
> @Quick-Flash ,
> ```
> # get smith_predict_enabled
> smith_predict_enabled = 1
> Allowed range: 0 - 0
> ```
> ON|OFF would be better, yes?

Something must be wrong in settings.c. it should be an on/off

nerdCopter:
fixed
```
# get smith_predict_enabled
smith_predict_enabled = OFF
Allowed values: OFF, ON
Default value: ON
```

nerdCopter:
i'd like to investigate casting `smith_predict_str` to boolean before merging this.  maybe one is better than the other, but i want to see first.

nerdCopter:
@Quick-Flash , if we use a toggle, should strength range be limited to `1` minimum?

Quick-Flash:
> @Quick-Flash , if we use a toggle, should strength range be limited to `1` minimum?

Delay and strength need to be minimum of 0

nerdCopter:
i'll add BBL header as well.

nerdCopter:
image

nerdCopter (approved):

PR #602, opened 2021-06-07
IMUF ptn order

Quick-Flash:
This allows the latest IMUF to select the ptn filter order (1-4). Default is pt3 filtering equivalent to 3 stacked pt1 filters. 

Also remove sharpness on all fc, looking at refactoring the kalman filter anyhow. Was getting annoying compiler errors due to kalman.c and sharpness, so removing this is the way I'm going for now.

Thank you for contributing to EmuFlight! ;)


nerdCopter (commented):
* exciting!
* will approve after some basic testing

Quick-Flash:
> * exciting!
> * will approve after some basic testing

Merging should likely occur after we test the kalman changes I have made as this removes sharpness (for ease of coding cases and for making non-helio/helio more similar).

src/main/interface/msp.c
```diff
@@ -1209,8 +1209,8 @@ bool mspProcessOutCommand(uint8_t cmdMSP, sbuf_t *dst) {
         sbufWriteU16(dst, gyroConfig()->imuf_pitch_q);
         sbufWriteU16(dst, gyroConfig()->imuf_yaw_q);
         sbufWriteU16(dst, gyroConfig()->imuf_w);
-        sbufWriteU16(dst, gyroConfig()->imuf_sharpness);
-#ifdef  USE_GYRO_IMUF9001
+        sbufWriteU16(dst, 0); //???? is this the way to do it??? used to be sharpness
```
nerdCopter, line 1212: yes.

nerdCopter:
are saying this requires #541?
& it this basically programmed the same as IMUF254?

Quick-Flash:
> are you saying this requires #541?
> & it this basically programmed the same as IMUF254?

No doesnt require #541, but we should look at merging our kalman changes before this so we miss sharpness even less. This just allows for a imuf254 that you can choose your imuf ptn level instead of being stuck at a set pt2 filter.

nerdCopter:
* in agreement on updated kalman, jsut need some testing.  (specifically, the hex i sent you will have "all the things" with exception of IMUF256(PTN) which need to get fixed.)

nerdCopter:
for records:
* this code applies to https://github.com/emuflight/imu-f/pull/13, but such is not working yet.

PR #606, opened 2021-06-10
purge SDS and WC (smartDtermSmooting and witchCraft)

nerdCopter:
* removes SmartDTermSmoothing
* remove WitchCraft

please validate code.


Quick-Flash:
looks good to me, will the gui work be difficult?

Quick-Flash (approved):

nerdCopter:
i think gui will be easy enough :crossed_fingers: , the kicker will be making sure expert-mode not re-enabling view for non-existing feature.

Quick-Flash:
ok just lemme know if we should merge this before the gui work is done, we probably need to come up with some sort of standard for the future though, lol

nerdCopter (commented):
fixed.

PR #616, opened 2021-06-22
[common/filter.c | common/filter.h | unit/maths_unittest.cc] Remove unused vector math functions

JulioCesarMatias:


JulioCesarMatias:
functions not used by emu

Quick-Flash (approved):

Quick-Flash:
At some point doing this isn't needed so much. We are going to ditch the current codebase for bf 4.3 as our base. So to an extent this sort of cleanup isn't super necessary. It is an improvement though and won't effect anything so it'll definitely get merged, but it does show that we should maybe put more effort into the 1.0.0 branches instead of master branch. Perhaps after we release 0.4.0 we will be able to make 1.0.0 the master branch to make this easier to see.

JulioCesarMatias:
> At some point doing this isn't needed so much. We are going to ditch the current codebase for bf 4.3 as our base. So to an extent this sort of cleanup isn't super necessary. It is an improvement though and won't effect anything so it'll definitely get merged, but it does show that we should maybe put more effort into the 1.0.0 branches instead of master branch. Perhaps after we release 0.4.0 we will be able to make 1.0.0 the master branch to make this easier to see.

ok,Kevin

PR #617, opened 2021-06-22
[common/filter.c] Add PT2 and PT3

JulioCesarMatias:
to use in the future

### DIFF
--- a/src/main/common/filter.c
+++ b/src/main/common/filter.c
@@ -64,6 +64,16 @@ FAST_CODE float pt1FilterApply(pt1Filter_t *filter, float input) {
     return filter->state;
 }
 
+/*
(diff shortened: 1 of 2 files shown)


Quick-Flash (changes requested):
This is already possible with the ptn filter. Pt1-pt4 type filtering is possible with such.

JulioCesarMatias:
> This is already possible with the ptn filter. Pt1-pt4 type filtering is possible with such.

oh,ok,sorry

Quick-Flash:
> > This is already possible with the ptn filter. Pt1-pt4 type filtering is possible with such.
> 
> oh,ok,sorry

No worries thanks for being proactive and thinking about this

PR #618, opened 2021-06-22
Smith predict uint8 rather than 16

nerdCopter:
* use `uint8_t` for `smith_predict_filt_hz` rather than `uint16_t`
* fix range for both CLI and OSD (1-250)
* needs testing/validation before merge


nerdCopter:
@nerdCopter's WIP MSP will need modification as well.

Quick-Flash (approved):
Lgtm

nerdCopter:
* another fix of OSD int16.
* tested OSD, worked (retreived same save OSD value in cli)

PR #623, opened 2021-06-26
EmuBoost2.0

AndreySemjonov:
New emuboost 2.0
Settings to test:
Emuboost 2.0: more boost -> more aggressive.
Emuboost 2.0 filter: more filter (higher number) = faster response, more noise.
Emuboost 2.0 cutoff: higher cutoff = takes longer for effect to happen, effect is smaller.
Emuboost 2.0 expo: how the effect starts, cutoff will be where effect happens, and expo is how it gets there.
Emuboost 2.0 dboost: ON/OFF for boosting of dterm.
Tests needed for best settings. Later some settings can be hide.

My settings used:
set emuboost2 = 8
set emuboost2_filter = 160
set emuboost2_cutoff = 200
set emuboost2_dboost = ON

Co-authored-by: Kevin Plaizier <46289813+Quick-Flash@users.noreply.github.com>


Quick-Flash:
This pr includes the sfdt stuff and really shouldnt.

nerdCopter:
AI Generated comment

Code-level analysis (this is the earlier iteration, superseded by #670 which drops the `emuboost2_dboost` toggle but keeps the same core design):

1. **Stale `axisLock` from reordered computation** — same defect as #670. `axisLock[axis]` is now computed inside the main per-axis loop, but `scaledAxisPid[ROLL/PITCH/YAW]` is still computed before that loop runs, and `axisLock` was converted to a persistent `static FAST_RAM_ZERO_INIT` array. Result: `scaledAxisPid` reads the previous PID-loop cycle's `axisLock` values every tick.

2. **D-term scaling injects a setpoint-shaped value, not a scaled error** — same formula as #670: `dtermErrorRate = currentPidSetpoint - gyroRate * inverseScaledError[axis]` scales only the measurement, not the setpoint. Collapses toward `currentPidSetpoint` itself during large/fast tracking error (flips, snaps), injecting a setpoint-scaled value into the D-term derivative at the same moment P is boosted. This branch additionally has `emuboost2_dboost`: when enabled, `inverseScaledError[axis] = 1.0f - scaledError` stays in effect even when `scaledError <= 0` (outside the P-boost direction), so D-term scaling applies more broadly here than in #670's version.

Full write-up on #670; both PRs share the same two defects. See issue #1277 for tracking.

## Recommended fix

Same as #670:
1. Revert `axisLock` to its own pre-loop pass, computed before `scaledAxisPid[*]`.
2. Compute the D-term derivative unscaled and apply `inverseScaledError[axis]` to the finished `pidData[axis].D` instead of baking it into `dtermErrorRate`. This branch already has the right idea commented out: `//pidData[axis].D *= inverseScaledError[axis]; // this gets applied elsewhere...` — uncommenting that (and removing the scaler from `dtermErrorRate`) is the fix. With `emuboost2_dboost` enabled, this same final-output scaling should apply regardless of `scaledError` sign, matching the toggle's intent without corrupting the error history.

nerdCopter:
AI Generated comment

Comparison against the existing EmuBoost implementation in master (`src/main/flight/pid.c`), assuming both defects noted above are fixed:

**Existing EmuBoost (master):**
- P/I boost (`errorBoost`): `boostedErrorRate = errorRate * |errorRate| * errorMultiplier`, clamped by `errorBoostLimit`, added into both P and I. Pure magnitude-based quadratic — no memory, no filtering, no distinction between error growing vs shrinking.
- D boost (`dtermBoost`, default 0/off): same quadratic-magnitude approach applied to `dDelta`, and it adds to D rather than shrinking it.
- The two are independent, uncoupled mechanisms — nothing ties how much P is boosted to how much D is reduced. Both are delay-free (no filter stage).

**EmuBoost 2.0:**
- Gates on error magnitude × rate-of-change of error together — only engages hard when tracking error is both large and actively getting worse, distinguishing that from "large but already recovering." A more targeted trigger than master's magnitude-only quadratic.
- Explicitly couples the two effects: the same `scaledError` value simultaneously boosts P and shrinks D, vs. master's two independent, potentially inconsistent knobs.
- Costs: adds a PT1 filter stage (group delay master's boost doesn't have) plus expo shaping — 4 additional tunable parameters layered on top of, not replacing, the 2 that already exist.

**AI opinion:** the underlying idea is a legitimate refinement — rate-of-change gating and P/D coupling are conceptually better targeted than a magnitude-only quadratic. But it does not fill a gap; it is a second, overlapping mechanism for a problem `errorBoost`/`dtermBoost` already addresses, with its own filter delay and added parameter surface, and no demonstrated evidence it outperforms a well-tuned existing setup. Given one tested flight with poor blackbox results and no community interest across two 2021 submission attempts (#623, #670), the added complexity is not justified without a real head-to-head comparison against a properly-tuned `errorBoost`/`dtermBoost` configuration — none exists. Not recommended for further pursuit.

PR #626, opened 2021-06-28
add dLPF2 type

nerdCopter:
* ~return~ add dLPF2 type.
* needs testing. (OSD, flight, BBL, etc)
* ~needs validation same variable names as historical.~ never existed in BuF/Emu.


Quick-Flash (approved):

nerdCopter:
* BBE doesn't see the dTerm LPF2 :disappointed: 
* makes me wonder if it needs a different variable name

PR #624, opened 2021-06-27
BlackBox version string

AndreySemjonov:
Do not need 3.7.0 anymore :)

PR #634, opened 2021-07-03
Update blackbox.c

AndreySemjonov:
Need separate TARGET field in BB headers

### DIFF
--- a/src/main/blackbox/blackbox.c
+++ b/src/main/blackbox/blackbox.c
@@ -1090,6 +1090,7 @@ static bool blackboxWriteSysinfo(void) {
     switch (xmitState.headerIndex) {
         BLACKBOX_PRINT_HEADER_LINE("Firmware type", "%s",                   "Cleanflight");
         BLACKBOX_PRINT_HEADER_LINE("Firmware revision", "%s %s (%s) %s",    FC_FIRMWARE_NAME, "3.7.0", shortGitRevision, targetName);
+        BLACKBOX_PRINT_HEADER_LINE("Target", "%s",                          targetName);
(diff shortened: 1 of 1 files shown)


src/main/blackbox/blackbox.c
```diff
@@ -1090,6 +1090,7 @@ static bool blackboxWriteSysinfo(void) {
     switch (xmitState.headerIndex) {
         BLACKBOX_PRINT_HEADER_LINE("Firmware type", "%s",                   "Cleanflight");
         BLACKBOX_PRINT_HEADER_LINE("Firmware revision", "%s %s (%s) %s",    FC_FIRMWARE_NAME, "3.7.0", shortGitRevision, targetName);
```
Quick-Flash, line 1092: will this "tricking" of BBL ever end? are we to the point that we can display the correct firmware version and use the emu bbl instead?

Quick-Flash (approved):

src/main/blackbox/blackbox.c
AndreySemjonov: What do you mean nu tricking? I made already that we can show correct version :) I need target to separate IMuF for helio and non helio in headers

Quick-Flash: "3.7.0" we are telling it that we are using version 0.3.7 which is clearly shown here 
image

AndreySemjonov: https://github.com/emuflight/EmuFlight/pull/624
This need to be merged. @nerdCopter want to wait before EBB completed completely

PR #643, opened 2021-07-14
Fix level mode/horizon issues

Quick-Flash:
Hopefully this fix any issues with level mode or horizon.


nerdCopter:
* reported working by @StratusFPV with exception of upside down does not autoflip, but this was by design from @PJC code-request.
* maybe need a mode-toggle of some sort.

nerdCopter (approved):
* approving this PR with openness to a PJC-mode toggle of some sort.
* do you want to keep the debug as is (roll only)?

nerdCopter:
* July 22 [7:58 PM] @StratusFPV: `I tested horizon after getting my 5" but only for about 10 minutes; it got dark. It seems okay. It's definitely not as snappy as angle mode to the "level" position but it's going back. The forward on pitch seems a little slower than the rest of the inputs.`

Quick-Flash:
i'll try to add a case to keep the old functionality now that I have a bit of time.

PR #647, opened 2021-07-19
Better kalman debugging

nerdCopter:
* @Quick-Flash 's Kalman debug branch
* only on roll, maybe should not merge.


nerdCopter:
@Quick-Flash , Oh, this is only on roll??  maybe i see why you never PR'd

Quick-Flash (approved):

PR #657, opened 2021-07-28
fix PR#500; dynamicFilter/dynamicDtermNotch

nerdCopter:
* fix Codacy `The scope of the variable 'XXXXX' can be reduced.`
* set `dtermDynNotch = false` by default

### DIFF
--- a/src/main/flight/pid.c
+++ b/src/main/flight/pid.c
@@ -197,7 +197,7 @@ void resetPidProfile(pidProfile_t *pidProfile) {
     .dterm_ABG_half_life = 50,
     .emuGravityGain = 50,
     .angle_filter = 100,
-    .dtermDynNotch = true,
+    .dtermDynNotch = false,
     .dterm_dyn_notch_q = 400,
                 );
 }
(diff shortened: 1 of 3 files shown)


src/main/flight/pid.c
```diff
@@ -197,7 +197,7 @@ void resetPidProfile(pidProfile_t *pidProfile) {
     .dterm_ABG_half_life = 50,
     .emuGravityGain = 50,
     .angle_filter = 100,
-    .dtermDynNotch = true,
+    .dtermDynNotch = false,
```
Quick-Flash, line 200: this and the notch q should be determined what is best. perhaps less dterm filtering and this on is the best way to proceed, IDK. Also one thing to note is that the number of filters isn't what gives issues with overfiltering, but rather how much each filter is doing that is the problem that creates over filtering.

nerdCopter: I think the code is good and operates well. my concern with dynamic D term notch = ON as default and having the requirement of feature dynamic_filter enabled, is that once you turn dynamic filter off you now have a single D term LPF only. or in the case of Helio dynamic filter is off by default thus again giving only a single determ LPF.

@sfonkter reported reported too much delay with dlpf1+dlpf2+dynDnotch.

otherwise I did fly dynamic D notch with a single LPF completely out of spectrum, and it flew very well. see discord. will post HD to accompany, later.

Quick-Flash:
lets get the gui stuff added

nerdCopter:
> lets get the gui stuff added

it's done :) , just have to cleanup and send to testing :wink:

PR #663, opened 2021-08-03
fix dterm dyn notch q limits

Quick-Flash:
Fixes a bug that @nerdCopter found. Min q should not be allowed to go to 0, likely we should rethink the min for the dynamic notch q as a whole as something like 50, but i guess thats for another time.


Quick-Flash:
superceded by nerds work.

Issue #592, opened 2021-06-01
[target request] ZEEZF7V2


TTGG4400:
Vendor/Model

- ZEEZF7V2
- ZEEZ RC F7 3030 FLIGHT CONTROLLER V2
- https://zeezrc.com/shop/zeez-f7-flight-controller/

CLI dump or Unified Target

- https://github.com/betaflight/unified-targets/blob/master/configs/default/ZEEZ-ZEEZF7V2.config

TTGG4400:
this was entered as a target request - dunno why it doesn't say that though

BeauBrewski:
Needs tested. Check orientation, motors, OSD, all that other stuff. PROPS OFF!
https://cdn.discordapp.com/attachments/547211754845765639/874799667429396490/EmuFlight_0.3.7_ZEEZ_ZEEZF7V2_Build_uncommitted_20210810EDT.hex

TTGG4400:
Unsure of my part in this, if any - should I install this version and test it on my 5" quad that hosts this FC?

nerdCopter:
[EmuFlight_0.3.4_ZEEZ_ZEEZF7V2_Build_5fb813e43.hex.zip](https://github.com/emuflight/EmuFlight/files/6976420/EmuFlight_0.3.4_ZEEZ_ZEEZF7V2_Build_5fb813e43.hex.zip)

TTGG4400:
I installed EmuFlight_0.3.4_ZEEZ_ZEEZF7V2_Build_5fb813e43.hex.zip and all is working except haven't been able to arm since I have a GPS - and I cannot find a CLI cmd to ignore GPS.

TTGG4400:
So - I disabled the GPS - and it armed, and the motors behaved as they should when I moved it about in my hand.

nerdCopter:
@TTGG4400 , thank you!
here is the GPS unlock: https://github.com/emuflight/EmuFlight/wiki/GPS-Lock-Bypass

TTGG4400:
@nerdCopter  Thx for that workaround.  It's perfect.  Let's me wait for sats - or not.

TTGG4400:
It flies nicely too - just got in 2 min before the rains began again.  Thx for the SW.

Rubber80:
Buongiorno Ragazzi, come faccio per installarlo sulla mia zeez v2? lo scarico e lo carico in modalità locale?
è gia testato? scusate ma non sono molto bravo

[...]

PR #669, opened 2021-08-11
add ZeezF7 v2

Kaiowarez:
* ZeezF7 v2
* #592


nerdCopter (approved):
* positive test results

nerdCopter:
* merged to master.

@Kaiowarez , one request please: when committing, instead of for example, committing with comment `add target`, please be specific such as `add ZeezF7 v2 `, thanks!

PR #670, opened 2021-08-12
EmuBoost2.0 (should perhaps have a different name)

Quick-Flash:
This code is best described as a form of smart error based gain scheduling. Essentially this code does 2 things, boost/shrink Dterm and boost Pterm. The logic here is simple. There are 2 ways to improve responsiveness on a PID controller. 1 is to use more Pterm, and another is to use less Dterm. Pterm is what drives your movement and Dterm is what slows your movement (at least dterm from gyro measurement, i've worked around this for dterm from error). So what if we want the best of both worlds when it comes to improving responsiveness while not gaining the problems associated with high Pterm and a low Dterm??? Well the logic here is simple, when we want/need more response boost the Pterm and shrink the Dterm. Shrinking the Dterm gives one major advantage over just increasing Pterm. Dterm is a source of noise, and shrinking the Dterm will help alleviate some of that noise and clean up our motors during moves.

How and when does it do this? Well the code looks at the change in error (the derivative of error) and the total error. The current error and the change in error are scaled between 1 and -1 before being multiplied together. This means that error has to be increasing or decreasing rapidly and error has to somewhat large for the full effect to occur.

Here is the math for how the Pterm is boosted and how the Dterm is boosted/shrunk along with some pseudo code. scaledError is our scaled total error and our scaled change in error.
`pterm = pterm + pterm * (scaledError) * emuboost2;
dterm = dterm * (1 - scaledError);`
Pterm is only boosted when Dterm is shrunk, and Dterm can be both boosted or shrunk. In situations where the error is growing, Dterm will be shrunk. In situations where error is shrinking the Dterm is boosted to help deal with overshoot.

What the settings do and how to tune this.

- emuboost2: This is the percentage of how much Pterm is allowed to be boosted. A value of 100 means that 100% more Pterm can be added, or at max value your P gains will be twice as much. A value of 1 would add 1% to your P gains at max. A value of 0 disables EmuBoost2.
- emuboost2_filter: This is the cutoff frequency for the boosting/shrinking effect.
- emuboost2_cutoff: Once the error is greater than the emuboost2_cutoff the error scaler is at its max value. 
- emuboost2_expo: This creates a curve to go from no boosting to full boosting. A value of 0 is linear and a value of 100 is the most expo possible.

How to tune this stuff: EmuBoost2 is relatively easy to tune. If you want more responsiveness increase the EmuBoost2 value, if it feels to responsive decrease this value. EmuBoost2_filter is a bit tougher to tune. A low filter cutoff will lead to a smoother boost, but at the downside of more delay in the boosting. Lower this filter if you have excess noise. EmuBoost2_cutoff essentially places how hard you need a move to be before you get much effect from EmuBoost2. Small cutoff values (10) means the effect is active almost all the time and a high cutoff (150+) will mean its mainly active during flips and rolls. EmuBoost2_expo is essentially how aggressively you want this effect to come into play. A high expo value will mean that small movements near center stick do not really feel "boosted" as much as a lower value. For expo play with small vs high values and see what feels best. If your cutoff is already quite small, then the effect of changing your expo will be less pronounced. It may be worthwhile to play with different values for your cutoff and expo and tune both together for the feel you desire.


Quick-Flash:
@nerdCopter this should be tested

nerdCopter:
Closing this PR without merge. The EmuBoost 2.0 concept is preserved in issue #1277 for future reference.

nerdCopter:
AI Generated comment

Code-level analysis of this diff found two issues beyond needing tuning:

1. **Stale `axisLock` from reordered computation.** Pre-PR, `axisLock[axis]` was computed for all axes in its own pass before `scaledAxisPid[ROLL/PITCH/YAW]` consumed it. This PR moves `axisLock[axis] = pt1FilterApply(&axisLockLpf[axis], stickMovement[axis]) * axisLockMultiplier;` into the main per-axis loop, but `scaledAxisPid[*]` is still computed before that loop runs, and `axisLock` was converted from a local stack array to `static FAST_RAM_ZERO_INIT float axisLock[XYZ_AXIS_COUNT]`. Result: `scaledAxisPid` reads the previous PID-loop cycle's `axisLock` values every tick, one cycle stale, every flight.

2. **D-term scaling injects a setpoint-shaped value, not a scaled error.** `dtermErrorRate = currentPidSetpoint - gyroRate * dynamicDtermScaler[axis]` only scales the measurement (`gyroRate`), not `currentPidSetpoint`. When EmuBoost2 engages hard (`dynamicDtermScaler → 0`, triggered by large/fast-changing tracking error — flips, snaps, punch-outs), this expression collapses toward `currentPidSetpoint` itself (e.g. 800°/s during a flip) rather than toward a shrunk error term, injecting a large, non-error-shaped value directly into the D-term derivative at the same moment `pidData[axis].P` is boosted up to 2.5x (default `emuBoost2=150` → `1 + 1.5 * scaledError`). Boosting P and injecting a setpoint-scaled spike into D on the same trigger is a plausible cause for rough/noisy blackbox traces during test flights — most likely to appear during the exact fast-maneuver conditions EmuBoost2 is designed to engage on.

Both defects predate this analysis and were never fixed on this branch. See issue #1277 and PR #623 (earlier iteration, same two defects) for related discussion.

## Recommended fix

1. **`axisLock`:** revert to computing it in its own per-axis pass before `scaledAxisPid[ROLL/PITCH/YAW]`, matching current master — nothing else in this diff requires the per-axis interleaving.
2. **D-term setpoint leak:** compute `dtermErrorRate` unscaled (`currentPidSetpoint - gyroRate`, matching master), then apply `dynamicDtermScaler[axis]` to the finished D output instead: `pidData[axis].D *= dynamicDtermScaler[axis];`. Scaling the error before differentiating it corrupts `previousDtermErrorRate` for every subsequent frame; scaling the final D output avoids that entirely and matches this PR's own stated intent (`dterm = dterm * (1 - scaledError)`).

nerdCopter:
AI Generated comment

Comparison against the existing EmuBoost implementation in master (`src/main/flight/pid.c`), assuming both defects noted above are fixed:

**Existing EmuBoost (master):**
- P/I boost (`errorBoost`): `boostedErrorRate = errorRate * |errorRate| * errorMultiplier`, clamped by `errorBoostLimit`, added into both P and I. Pure magnitude-based quadratic — no memory, no filtering, no distinction between error growing vs shrinking.
- D boost (`dtermBoost`, default 0/off): same quadratic-magnitude approach applied to `dDelta`, and it adds to D rather than shrinking it.
- The two are independent, uncoupled mechanisms — nothing ties how much P is boosted to how much D is reduced. Both are delay-free (no filter stage).

**EmuBoost 2.0:**
- Gates on error magnitude × rate-of-change of error together — only engages hard when tracking error is both large and actively getting worse, distinguishing that from "large but already recovering." A more targeted trigger than master's magnitude-only quadratic.
- Explicitly couples the two effects: the same `scaledError` value simultaneously boosts P and shrinks D, vs. master's two independent, potentially inconsistent knobs.
- Costs: adds a PT1 filter stage (group delay master's boost doesn't have) plus expo shaping — 4 additional tunable parameters layered on top of, not replacing, the 2 that already exist.

**AI opinion:** the underlying idea is a legitimate refinement — rate-of-change gating and P/D coupling are conceptually better targeted than a magnitude-only quadratic. But it does not fill a gap; it is a second, overlapping mechanism for a problem `errorBoost`/`dtermBoost` already addresses, with its own filter delay and added parameter surface, and no demonstrated evidence it outperforms a well-tuned existing setup. Given one tested flight with poor blackbox results and no community interest across two 2021 submission attempts (#623, #670), the added complexity is not justified without a real head-to-head comparison against a properly-tuned `errorBoost`/`dtermBoost` configuration — none exists. Not recommended for further pursuit.

PR #675, opened 2021-08-18
Update serial-4way-if to v20005

4712:
Reservation of an MCU_ID area to simplificate adding of upcoming BLHeli_32 MCU

### DIFF
--- a/src/main/io/serial_4way.c
+++ b/src/main/io/serial_4way.c
@@ -77,7 +77,7 @@
 // *** change to adapt Revision
 #define SERIAL_4WAY_VER_MAIN 20
 #define SERIAL_4WAY_VER_SUB_1 (uint8_t) 0
-#define SERIAL_4WAY_VER_SUB_2 (uint8_t) 04
+#define SERIAL_4WAY_VER_SUB_2 (uint8_t) 05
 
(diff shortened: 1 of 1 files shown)


nerdCopter:
[EmuFlight_0.4.0_BETAFPVF405_Build_f4c9801d8__PR549.hex.zip](https://github.com/emuflight/EmuFlight/files/7008713/EmuFlight_0.4.0_BETAFPVF405_Build_f4c9801d8__PR549.hex.zip)

TYVM @4712!!! @SkylionFPV  is testing.

SkylionFPV:
I just tested this with the above hex and it's working perfectly now =]

Quick-Flash:
Lets just verify this works with jesc and blheli_s before we call it good.

4712:
@Quick-Flash there is no change for jesc and blheli_s, but verification is always good ;)

SkylionFPV:
@Quick-Flash I just tested with betafpv f405 v3 20a blheli_s aiofc with the same target as above and it connects to blheli_s suite, blheli_s configurator for jesc, and bluejays configurator just fine

Quick-Flash (approved):
Thanks for the PR

PR #676, opened 2021-08-18
Update horizon defaults

Quick-Flash:
Improvements to the defaults for horizon mode. Should give more users a slightly better time using horizon mode.


nerdCopter (approved):
`FailsafeFPV: It auto leveled until like over 50% pitch/roll and then I could flip over. It doesn't auto level when i'm upside down`
`QuickFlash: It shouldn't be auto leveling while upside down`

PR #688, opened 2021-08-22
Change imuf_q defaults

Quick-Flash:
Due to many users complaining of issues with this latest release and seeing a handful fix this by just raising their IMUF q, I'm pulling these changes. I think we should just replace the current 0.4.0 hex files with new ones that contain these defaults.


Quick-Flash:
Question I have is if its better to change the defaults or change the under the hood scaling as people will be used to the 3000 q they used in previous versions?

nerdCopter:
i have to think on it as well.
another option is add an alternative "better-defaults" preset... thoughts?

nerdCopter:
> Question I have is if its better to change the defaults or change the under the hood scaling as people will be used to the 3000 q they used in previous versions?

we actually spoke of changing the scale, but it never occurred.

nerdCopter:
> Question I have is if its better to change the defaults or change the under the hood scaling as people will be used to the 3000 q they used in previous versions?

after some thought and recall of old conversation, i feel scaling to historical range/values is proper

Quick-Flash:
> > Question I have is if its better to change the defaults or change the under the hood scaling as people will be used to the 3000 q they used in previous versions?
> 
> after some thought and recall of old conversation, i feel scaling to historical range/values is proper

The question is by now some have tuned with new scale (which from previous feedback we assumed was fine) so now making this change will maybe cause more issues if it's a scale change. Idk which way is better.

nerdCopter:
* i suppose since Q can go to 16000, i can live with it.
* i'll start merging.
* release notes did say "W & Q now have more effect"

nerdCopter (approved):

gretel:
> Question I have is if its better to change the defaults or change the under the hood scaling as people will be used to the 3000 q they used in previous versions?

release a fixed version which eventually would be 0.4.1 now.

PR #689, opened 2021-08-22
Making angle mode more responsive

Quick-Flash:
Project MockingBird has noted that the current angle mod implementation is not very responsive. This code change removes scaling on pitch and roll, which should in theory allow angle mode to go past your set max angle when pitch and roll are full. However it seems that event rarely or never occurs and allowing this results in a better flying experience.


nerdCopter (approved):
* approving in trust

nerdCopter:
>     * approving in trust

* should probably wait for PJC results before merge, @Quick-Flash

nerdCopter:
* will merge based on PJC feedback

PR #654, opened 2021-07-25
Support for Redpine SPI RX protocol from BF

pashky:
Any interest in Redpine RX protocol port (https://www.rcgroups.com/forums/showthread.php?3236043-Redpine-Lowest-latency-RC-protocol)? Not exactly mainstream feature, and probably not very relevant in the world where ELRS exists, but looks like some people asked for it in the past (#420) and there's already fleet of whoop AIO boards with built-in cc2500. 

Anyway, I was bored and got across it by pure chance, so here you go. I tested it briefly around home on MATEKF411RX (Tinyhawk), seems to work fine. RC latency is not what's holding back this particular quad, so can't comment on that part, but link was stable.

Unlike #444 I didn't even try to copy all BFs refactorings in SPI RX code that happened since 3.5. Not much point in doing that in 0.3.x branch. Instead I just hooked up older FrSkyXX and cc2500xx functions where appropriate following the style of other cc2500 protocols.

PR #699, opened 2021-08-29
Fix Redpine SPI protocol for CRAZYBEEF4RX

pashky:
Fixes issue reported in #654 where redpine was not compiled in for crazybee boards. Also added default case in spi init that in theory should just disable spi rx instead of bricking FC if protocol is not available for target (this bit needs testing but I have no quads with me right now).


nerdCopter:
`Sept 05: Beau Brewski: Yeah, it ended up working great! Looks like me and another guy on #654 reported it works`

nerdCopter (approved):
Thank you!

PR #701, opened 2021-08-31
Add 50 mW TX power level to CRSF protocol

avsaase:
This PR adds the 50 mW TX power level to the CRSF protocol. TBS is adding this power level in a future firmware version and they already added support for it in OpenTX. ExpressLRS uses 50 mW but cannot include it in the link statistics until the various projects that use CRSF support it, hence this PR.


gretel:
@avsaase thanks a lot!

nerdCopter (approved):
Thank you!

PR #705, opened 2021-10-04
SmartAudio support 5th power level

nerdCopter:
* ask requested a millions times
* needs real-world testing


BeauBrewski:
My bad, that needs switched where Tramp has 5th power level, not SmartAudio.

PR #715, opened 2021-10-24
Remove packed on accDev_t for dropping alignment warning

tstibor:
With beginning of GCC 9 a great deal of warnings of "unaligned pointer value"
are introduced. For dropping the warning:

```
taking address of packed member of 'struct accDev_s' may result in an
unaligned pointer value [-Waddress-of-packed-member]
```
simply remove the 'packing' similar to commit:
[5b09b397a74e76fff4be254b35576b1aea321368](https://github.com/emuflight/EmuFlight/commit/5b09b397a74e76fff4be254b35576b1aea321368)

The __attribute__((packed)) means the smallest possible alignment (1 byte),
usually used for implementing binary network protocols.


nerdCopter (approved):
* flew 8 packs; no problems

PR #716, opened 2021-10-25
Fix warning cast between incompatible function types

tstibor:
The function cast in macro PG_REGISTER_ARRAY_WITH_RESET_FN wants
(pgResetFunc)(void * /* base */)
rather than
(pgResetFunc)(void * /* base */, int /* size */);

This commit fixes the warning caused by incompatible function types.

Note: This fix is already available also in https://github.com/betaflight/betaflight/blob/daf202e8c365d14aeeb8a0942b7a0356a11d30d8/src/main/pg/pg.h#L44
and
https://github.com/betaflight/betaflight/blob/daf202e8c365d14aeeb8a0942b7a0356a11d30d8/src/main/pg/pg.c#L56


nerdCopter (approved):
* configured a quad; & used profiles
* flew 8 packs; no problems

PR #717, opened 2021-10-25
Fix warning statement may fall through

tstibor:
This commit fixes warning:
statement may fall through [-Wimplicit-fallthrough=]

This patch is also borrowed from betaflight to fix the above described problem:
https://github.com/betaflight/betaflight/blob/daf202e8c365d14aeeb8a0942b7a0356a11d30d8/src/main/rx/rx.c#L589


nerdCopter (approved):
* flew 8 packs; no problems

PR #719, opened 2021-10-29
Fix buffer is accessed out of bounds

tstibor:
Fix out of bounds buffer problems caused by employing incorrect sizes
VOL_LABEL_LEN != 12 in bs->volume_label
FILE_NAME_SHRT_LEN + FILE_NAME_EXTN_LEN != 8 in entry->name


nerdCopter (approved):
* balckboxed several flights while flying two occasions (8 packs); no problems

PR #724, opened 2021-11-15
Update peak frequency even if below mean energy

DzikuVx:
Current implementation on Dynamic Notches tries to track peaks only when they are above the noise floor.

This is done by comparing the peak value with a mean of all bins above min bin/frequency. However, this is not true, as the bin value mean is not a noise floor level.

In the case of some builds, the first peak (or 2 first peak combined) can have enough energy to move the mean high enough, so that the following peak fall below it, while still being well above the noise floor. As a result, Dynamic Notch stops tracking those lower energy peaks even if they are visible above the noise floor.

High energy peaks can saturate the peak detection algorithm.

This PR modifies this behavior by removing the peaks[p].value > sdftMeanSq condition. As a result, the frequency domain peak will be tracked even if its energy is below the mean value.

Bear in mind that previous steps detect on N strongest peaks. Tests revealed that 2nd and 3rd harmonics are still visible in the tracked frequency range. High-frequency notch attenuating low energy peak is still better than notch loosing track of a peak completely.


nerdCopter:
this is built to https://github.com/emuflight/dev-unstable/releases/tag/hex-2115 for testers

nerdCopter (approved):
Thank you

PR #729, opened 2021-11-23
Fix the kalman filter

Quick-Flash:
Remove a line in kalman.c to make it perform like its supposed to. Extra line was supposed to be removed, but I backported 1.0.0 changes incorrectly into 0.4.0, but this should improve flight performance a bit more :)


nerdCopter:
this is built to https://github.com/emuflight/dev-unstable/releases/tag/hex-2115 for testers

nerdCopter (approved):
* thank you

PR #751, opened 2022-02-09
fix dyn notch range

Quick-Flash:
Fix the dynamic notch range, see https://github.com/betaflight/betaflight/pull/11397


Quick-Flash:
will fix it differently later

PR #754, opened 2022-02-28
Kalman filter

Quick-Flash:
This pull request adds the kalman filter into 1.0.0, Please debug this with the debug mode kalman turned on. Looking to verify that the latency and attenuation is the same as Emu 0.4.0. To tune this filter a higher Q value will reduce filtering as well as a higher W.

Possible future improvements, allow the Q value to move with throttle. Reduce the range on the kalman Q AKA instead of a 100-16000 range reduce it to a 1-100 range. Don't know how we will want to do this yet. Add ability to tune the variance scale, perhaps this could be the method for tuning noise rather than q


Quick-Flash:

nerdCopter:

PR #755, opened 2022-02-28
Better gyro fusion

Quick-Flash:
This code looks at the variance over the last 8 samples and uses this to determine how to fuse dual gyros together. Gyros with more variance will be fused less while the gyro with less variance will be fused at a higher percent.

To test this code set your debug mode to fusion and set the gyro_to_use in the cli to be both_variance. Debug 0 is gyro1, debug 1 is gyro 2, debug 2 is the fused gyro data. It should be slightly cleaner than the other gyro signals.


nerdCopter:

PR #761, opened 2022-03-02
choose ptn filter order for rc filtering

Quick-Flash:
BF forces users to use pt3 filters for the rc filtering. This opens up the option to use pt1, pt2, or pt3 filters for the rc smoothing filters. The setting name is rc_smoothing_order in the cli to change this setting.


nerdCopter:

PR #762, opened 2022-03-02
Dynamic rc filtering

Quick-Flash:
The idea is to decrease rc filtering when you really bang your sticks. The reasoning is that if your racing you probably don't care a lot about fast movement, but slow movements you likely want to have filtered quite high, this would be a way to accomplish that.

TODO add logging variables for this, perhaps add it to the osd, and test this code.


nerdCopter:

PR #775, opened 2022-03-17
Import HDZero OSD Driver from iNav

### LINKED ISSUE #700: shark byte canvas mode
Enable OSD to be displayed through serial port 
Add OSD_displayport_device = MSP and displayport_msp_serial =  into CLI.

saidinesh5:
#### Notable changes:
* And added missing functions to bitarray.h/.c, msp_serial.h/.c
* Used baudRates[BAUD_115200] instead of
baudRates[portConfig->peripheral_baudrateIndex]

#### Additional changes:
* serial: serialPortFunction_e now needs 32 bits
* Updated msp_protocol minor version to 52

#### How to use:
In CLI, Set FUNCTION_HDZERO_OSD to the appropriate serial port:
Eg. Assuming to use serial port 4, enter the following CLI commands to
enable FUNCTION_HDZERO_OSD in EmuFlight Configurator's CLI:
```
serial 3 65536 115200 115200 0 115200
save
```
## REQUIRES HDZero firmware Rev 11062022 or later, otherwise older firmware will be a corruption issue in EmuFlight due to VTX-Tables "BackPack"
* https://www.hd-zero.com/document

#### Known quirks to be attended before ready for `master`:
* OSD VTX elements not synchronized with HDZero hardware settings [non-breaking issue].
* DJI expects OSD elements in the 0-31,0-31 range (unpacked), where this PR is 0-63,0-63 (unpacked). e.g. `osd_crosshairs_pos = 2253` not `osd_crosshairs_pos = 8589`. Desire to be backward-compat and/or intercept layer for DJI.
* Walksnail requires PAL in Configurator for text alignment. Desire fixed.
* Supports WTF.OS
* Support UNKNOWN for DJI 03 -- needs feedback on what works or what does not. expecting MSP is different.
* Needs to be updated and brought modern similar to iNav merges:
  - https://github.com/iNavFlight/inav/pulls?q=is%3Apr+hdzero+is%3Amerged
  - https://github.com/iNavFlight/inav/pulls?q=is%3Apr+is%3Amerged++hd+osd  (some not all)
<hr>
Closes #700


src/main/drivers/display_font_metadata.h
```diff
@@ -0,0 +1,24 @@
...
+#include <stdint.h>
+
+typedef struct osdCharacter_s osdCharacter_t;
+
+typedef struct displayFontMetadata_s {
+    uint8_t version;
+    uint16_t charCount;
+} displayFontMetadata_t;
+
+// 'I', 'N', 'A', 'V'
```
nerdCopter, line 13: @saidinesh5 , what is the purpose of this section? with "INAV"

src/main/interface/settings.c
```diff
@@ -355,7 +355,7 @@ static const char * const lookupTableRescueSanityType[] = {
 
 #ifdef USE_MAX7456
 static const char * const lookupTableVideoSystem[] = {
-    "AUTO", "PAL", "NTSC"
+    "AUTO", "PAL", "NTSC", "HDZERO"
```
nerdCopter, line 358: @saidinesh5 , @madchiller , @mricho ,
should we consider `HD720` here rather thn `HDZERO`? (in preparation of HD1080)

src/main/drivers/display_font_metadata.h
saidinesh5: I had to include this for availing `displayFontMetadata_t` this one function (which is used nowhere). Originally I wanted to keep this osd driver as close to the inav version as possible to minimize the headaches when updating, but I am beginning to think that we don't have to worry about that right now because we have a lot of code that is useless currently.

```
static bool getFontMetadata(displayFontMetadata_t *metadata, const displayPort_t *displayPort)
{
    UNUSED(displayPort);
    metadata->charCount = 512;
    metadata->version = FONT_VERSION;
    return true;
}
```

src/main/interface/settings.c
saidinesh5: This video system information is purely specific to HDZero (so as to signify that this OSD supports 50x18 characters) and is used only by the configurator to show the canvas of appropriate size.

nerdCopter:
* need to investigate "backpack" potentially the cause of MSP corruption

nerdCopter:
* needs to intercept MSP for any old HDZero firmware that supports backback.
* otherwise, seems that everything works with HDZero's latest firmware.
* would prefer more testers testing it.

saidinesh5:
I think we need to fix the OSD canvas size because DJI goggles only support SD resolution.
That's the reason we didn't merge this yet, because this is causing DJI osd to break.

nerdCopter:
@saidinesh5 , can you rebase on `master`?

nerdCopter:
https://github.com/iNavFlight/inav-configurator/blob/be88b5af8479d524089ef7a720c58b45c6026879/tabs/osd.js#L550

nerdCopter:
This needs work. I plan to take ownership. I really want to avoid MSP change for the uint32. betaflight stayed uint16.

src/main/drivers/display_font_metadata.h
nerdCopter: i believe, this will need to be EMUF for hdzero font `EMUF_000.bmp`. not validated yet.

PR #778, opened 2022-03-19
Update iterm rotation

Quick-Flash:
This updates iterm rotation to only effect pitch and roll. It is also more accurate than the previous math was. Big shoutout to RotorFlight and [pmattila](https://github.com/pmattila) for the code idea. 

Absolute control is removed and iterm rotation is set to on as default.


nerdCopter:

PR #783, opened 2022-03-29
Kalman Filter

nerdCopter:
* Cherry-pick @Quick-Flash code from #754

`This pull request adds the kalman filter into 1.0.0, Please debug this with the debug mode kalman turned on. Looking to verify that the latency and attenuation is the same as Emu 0.4.0. To tune this filter a higher Q value will reduce filtering as well as a higher W.`

`Possible future improvements, allow the Q value to move with throttle. Reduce the range on the kalman Q AKA instead of a 100-16000 range reduce it to a 1-100 range. Don't know how we will want to do this yet. Add ability to tune the variance scale, perhaps this could be the method for tuning noise rather than q`


nerdCopter:
i guess latest BF changed something.  SITL Simulator wont compile:
```
./src/main/common/kalman.c:22:10: fatal error: arm_math.h: No such file or directory
   22 | #include "arm_math.h"
      |          ^~~~~~~~~~~~
compilation terminated.
make[2]: *** [Makefile:442: obj/main/SITL/common/kalman.o] Error 1
```

nerdCopter:
AI Generated comment

PR #784, opened 2022-03-29
Better Gyro Fusion

nerdCopter:
* Cherry-Pick @Quick-Flash code
* exclude `SPRACINGH7CL` `SPRACINGH7EF`

`This code looks at the variance over the last 8 samples and uses this to determine how to fuse dual gyros together. Gyros with more variance will be fused less while the gyro with less variance will be fused at a higher percent.`

`To test this code set your debug mode to fusion and set the gyro_to_use in the cli to be both_variance. Debug 0 is gyro1, debug 1 is gyro 2, debug 2 is the fused gyro data. It should be slightly cleaner than the other gyro signals.`


nerdCopter:
* what happens if a gyro goes bad? 
* what happens if one is always bad vs going bad mid-flight. 
* can it automatically compensate for bad gyro or will things get worse?
* (i.e. does it need additional gyro checks or compensation-code)

Quick-Flash:
> * what happens if a gyro goes bad?
> 
>     * what happens if one is always bad vs going bad mid-flight.
> 
>     * can it automatically compensate for bad gyro or will things get worse?
> 
>     * (i.e. does it need additional gyro checks or compensation-code)

Comparing this to the current gyro fusion, this code will make the gyro that performs worse be ignored more than one that is still working right. So if one gyro went bad it would fly better with this code than it would without this code. However, it isn't good enough to totally ignore a bad gyro and will still perform better if the bad gyro is just disabled completely. Perhaps it could be modified so that if it detects its constantly leaning hard towards one gyro that it will just disable the other gyro and only rely on one. 
TLDR, it will work better than the current gyro fusion if one gyro goes bad, but it is still better to disable a bad gyro, but it may be possible to change that.

nerdCopter:
AI Generated comment

PR #785, opened 2022-03-29
Auto Notch

nerdCopter:
* Cherry-Pick @Quick-Flash code from #758

`Autonotch is code that automatically decides how much the notch filter is crossfaded with its input signal. By determining the amount of noise at the notch frequency we can set up a limit. As noise approaches that limit the notch filter is used more, as noise is low it isn't used. This will automatically phase out rpm and dynamic notch filters that to do not have much use. This should reduce delay and turn off the effect of notch filters that are doing nothing.`

`To test you have new rpm_noise_limit, and dyn_notch_noise_limit, both of which determine how much noise can pass through before the notch filter is fully turned on. A higher value allows more noise to pass through before the notch filter takes full effect, smaller values happen sooner.`

### DIFF
--- a/make/source.mk
+++ b/make/source.mk
@@ -229,6 +229,7 @@ SPEED_OPTIMISED_SRC := $(SPEED_OPTIMISED_SRC) \
             common/maths.c \
             common/sdft.c \
             common/typeconversion.c \
+            common/auto_notch.c \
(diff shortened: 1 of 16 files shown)


nerdCopter:
AI Generated comment

PR #797, opened 2022-04-14
dynamic notch on pidsum

Quick-Flash:
Going from gyro to pid adds a lot of nonlinear changes to the signal, therefore gyro filtering and pidsum filtering are different. Dynamic dterm, tpa, and more all have different effects on the pidsum. By running the dynamic notch on the pidsum instead of on the gyro we can better account for the effect that pids have on the signal. AKA this will filter PID, just like we do when filtering the gyro, but it will better filter for the specific ratio of PID and any extra dynamics like TPA, anti-gravity, etc. In other words this in theory should work a bit better than just filtering the gyro with a dynamic notch.


nerdCopter:
AI Generated comment

PR #798, opened 2022-04-15
allow negative ff-boost

Quick-Flash:
High ff brings overshoot and undesirable feel. This allows ff boost to be negative which might allow for even more accurate stick tracking and less overshoot, needs testing. Perhaps normal ff should also be allowed to be negative with positive boost. At that point boost and FF would need to be separated.


nerdCopter:
AI Generated comment

PR #807, opened 2022-05-11
Improve peak estimation

Quick-Flash:
Fix an issue where the old estimation can cause some very off very bad numbers. When the old denom would be near 0 it would create wildly inaccurate peak estimations, this fixes that.


nerdCopter:
what about the `#define DYN_NOTCH_SMOOTH_HZ        4` -->`8`?

nerdCopter:
AI Generated comment

PR #818, opened 2022-06-23
default 1.5khz logging

nerdCopter:
* default 1.5khz logging

### DIFF
--- a/src/main/blackbox/blackbox.c
+++ b/src/main/blackbox/blackbox.c
@@ -91,7 +91,7 @@
 PG_REGISTER_WITH_RESET_TEMPLATE(blackboxConfig_t, blackboxConfig, PG_BLACKBOX_CONFIG, 1);
 
 PG_RESET_TEMPLATE(blackboxConfig_t, blackboxConfig,
-                  .p_ratio = 32,
+                  .p_ratio = 48,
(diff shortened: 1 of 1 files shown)


nerdCopter:
i believe @madchiller would prefer 2k as default, thoughts?

madchiller:
> i believe @madchiller would prefer 2k as default, thoughts?

Not up to speed why we are changing default logging rates ..  however..  I do usually request 2k logs in general..  2k is best for most filter tuning, but 1k is all that’s needed for step responses…  maybe 1.5 is a happy default

nerdCopter:
> > i believe @madchiller would prefer 2k as default, thoughts?
> 
> Not up to speed why we are changing default logging rates .. however.. I do usually request 2k logs in general.. 2k is best for most filter tuning, but 1k is all that’s needed for step responses… maybe 1.5 is a happy default

it's definitely debate-able on why to change defaults.  pilots continuously submitting logs for review with only 1Khz and no gyro_scaled.  I'm just trying to ease the burden.  Honestly, an alternative is to refuse to review any log not up to standards.  If nothing else, very good to discuss.

PR #819 also involved here.

nerdCopter:
i think this one is acceptable and can be merged.

Quick-Flash (approved):

PR #833, opened 2022-07-23
set CRSF/ELRS SNR to uplink_SNR rather than downlink_SNR

nerdCopter:
* ELRS uses uplink_SNR (working)
* Tested on CRSF (working)

`
2022-07-23 [12:24 PM CDT] CapnBry: downlink_SNR is the wrong field to use. That is the SNR going from the receiver to the transmitter. Note how all your other fields use the uplink versions because you want to display TX to RX uplink info.
uplink_TX_Power we fill but only in wide switch mode or Fullres modes in 3.0, since those are the only modes that have space to send them (edited)
`


mricho (approved):
Works for me, with CRSF. shows SNR, it fluctuates like it should, smaller number when tx is further away, got larger number when walked back.

PR #854, opened 2022-08-06
fix mixer.c for when undefining USE_BRUSHED_ESC_AUTODETECT

nerdCopter:
* fix mixer.c for when using `#undef USE_BRUSHED_ESC_AUTODETECT`
* please validate logic; this compiles but untested.


Quick-Flash (approved):
Good Catch

PR #916, opened 2023-06-07
[Target] CRAZYBEEF4FR add  BMI270  ICM42688P

xxxvik-xakerxxx:
CRAZYBEEF4FR - ICM42688P; BMI270
testing only CRAZYBEEF4FR
closes https://github.com/emuflight/EmuFlight/issues/912


nerdCopter:
this should work just fine for both, but i notice that betaflight has separate target definitions for crazybeef4 series.
@BeauBrewski , should we make crazybee specific targets? i dont know if they are f411 or just f4

image

nerdCopter:
converted to `draft` to block merging because bmi270 is not merged to master yet.

nerdCopter (approved):

PR #933, opened 2023-10-24
ICM426XX - Disable AFSR

nerdCopter:
* ICM426XX - Disable AFSR
* Betaflight: Disable ICM426XX AFSR feature to prevent stalls
* Ardupilot: AP_InertialSensor: fix for ICM42688 stuck gyro issue
these undocumented bits in register 0x4d control the "adaptive full
scale range" mode of the ICM42688. The feature is enabled by default
but has a bug where it gives "stuck" gyro values for short periods
(between 1ms and 2ms):, leading to a significant gyro bias at longer
time scales, enough to in some cases cause a vehicle to crash if it is
unable to switch to an alternative IMU
* Thanks to all participants here: https://github.com/betaflight/betaflight/issues/12970
* Thanks to @DutchFPV for quick-turnaround testing
* Thanks to Betaflight devs communications

Co-authored-by: Andrew Tridgell <andrew@tridgell.net>
Co-authored-by: Steve Evans <Steve@SCEvans.com>

### DIFF
--- a/src/main/drivers/accgyro/accgyro_spi_icm426xx.c
+++ b/src/main/drivers/accgyro/accgyro_spi_icm426xx.c
@@ -289,7 +289,7 @@ void icm426xxGyroInit(gyroDev_t *gyro)
 
     spiBusWriteRegister(&gyro->bus, ICM426XX_RA_INT_SOURCE0, ICM426XX_UI_DRDY_INT1_EN_ENABLED);
 
-    uint8_t intConfig1Value = spiBusReadRegister(&gyro->bus, ICM426XX_RA_INT_CONFIG1);
+    uint8_t intConfig1Value = spiReadRegMsk(&gyro->bus, ICM426XX_RA_INT_CONFIG1);
     // Datasheet says: "User should change setting to 0 from default setting of 1, for proper INT1 and INT2 pin operation"
     intConfig1Value &= ~(1 << ICM426XX_INT_ASYNC_RESET_BIT);
     intConfig1Value |= (ICM426XX_INT_TPULSE_DURATION_8 | ICM426XX_INT_TDEASSERT_DISABLED);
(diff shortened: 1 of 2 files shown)


nerdCopter:
8k
debug_mode = GYRO_RAW
unsoldered hand-movements

[EMUF_BLACKBOX_LOG_20231024_103529-holding-wire.BBL.zip](https://github.com/emuflight/EmuFlight/files/13120413/EMUF_BLACKBOX_LOG_20231024_103529-holding-wire.BBL.zip)

nerdCopter:
i'm unsure what to use: `spiBusReadRegister` or `spiReadRegMsk`. both seem to work. but BBL shows some different axis behavior.
left=spiBusReadRegister ; right=spiReadRegMsk (debug, X,Y,Z seemingly re-arranged or maybe just scaled differently)
image


EmuFlight did not have `spiReadRegMsk` before ICM426xx, which seems to only "wait" before calling `spiBusReadRegister`.
https://github.com/emuflight/EmuFlight/blob/c55b0d708bcd9d4d127a8ae5523861bddbadf5c3/src/main/drivers/bus_spi.c#L275-L279

Quick-Flash (approved):
This looks like it should fix the issue, before we merge it we should see if we can replicate the issue on the bench and then show it not happening after this fix.

Quick-Flash:
> i'm unsure what to use: `spiBusReadRegister` or `spiReadRegMsk`. both seem to work. but BBL shows some different axis behavior. I don't have this FC mounted so i cannot flight-test :(
> 
> left=spiBusReadRegister ; right=spiReadRegMsk (debug, X,Y,Z seemingly re-arranged or maybe just scaled differently) image
> 
> EmuFlight did not have `spiReadRegMsk` before ICM426xx, which seems to only "wait" before calling `spiBusReadRegister`.
> 

Since you have an FC that works with this code, If your unsure what scaling or whatever it needs try using the gui to see what rotating the quad does. If you tilt the fc 90 degrees you should see the quad respond correctly to that and not take some extra time to correct, or see it overshoot. If thats the case then we know its working right.

nerdCopter:
@Quick-Flash , 90 tilts seem to work fine, with the caveat that our Configurator can become out-of-sync with physical position if moved too rapidly. (tested with mpu6000 and bmi270, this is still true, but maybe worse with icm42688p, both this PR and master)

master vs fix (where fix uses `spiReadRegMsk`)
obvious ticks left, not on right
image

Quick-Flash:
@nerdCopter can you test 90 degree pitch with `dcm_kp` and `dcm_ki` to zero? This will disable the accelerometer fusion. If 90 degree pitch without going to crazy works fine then I'll say it looks good to merge.

nerdCopter:
still does it, but sensors tab looks good. maybe i'll wait for flight-test. or maybe just know consider Configurator is obsolete on setup page render.

nerdCopter:
findings:
* flight test as good
* logging at 8k will cause undesired effect. flight-control is less responsive to stick-input.
* ICM4266p Configurator 3D render becomes unsynchronized in both EmuFlight and Betaflight. All gyro's seem to do but icm426xx is a bit more drastic.
* Sensors tab seems to be fine.
* i had concerns about using `spiReadRegMsk` versus `spiBusReadRegister` because they both had a bitwise OR operator (reg | 0x80), but it turns out that `reg | x80` is equal to `reg | 0x80 | 0x80`

PR #934, opened 2023-10-25
add Gyro LPF2 type to OSD

nerdCopter:
* add Gyro LPF2-type to OSD
* tested - OSD/Configurator synchronization good.
* moved gLPF1-type to top

image


Quick-Flash (approved):
LGTM

PR #935, opened 2023-10-25
fix/add USE_GYRO_LPF2 gating

nerdCopter:
- fix/add USE_GYRO_LPF2 gating
- tested OSD and Configurator; Configurator resets to zero's when refresh as expected.

image


nerdCopter:
request any reviewers to check that i only gated gyro_lowpass2 and associated functions, not any static-notch, etc.

Quick-Flash (approved):
LGTM

PR #936, opened 2023-10-26
revert some USE_GYRO_DATA_ANALYSE and USE_GYRO_LPF2 gating

nerdCopter:
* revert some USE_GYRO_DATA_ANALYSE and USE_GYRO_LPF2 gating because C compiler sucks
* why: because compiler reports `warning: embedding a directive within macro arguments is not portable`, which reports that `ifdef` inside macros may have unexpected results depending on compiler/devices.
* reverting these requires reverting variables. and thus revert MSP read of values. (but not MSP write of `0`'s)

### DIFF
--- a/src/main/flight/pid.c
+++ b/src/main/flight/pid.c
@@ -197,10 +197,8 @@ void resetPidProfile(pidProfile_t *pidProfile) {
     .dterm_ABG_half_life = 50,
     .emuGravityGain = 50,
     .angle_filter = 100,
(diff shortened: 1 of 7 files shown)


Quick-Flash (approved):
As discussed this helps remove potential issues of undefined behavior. Thanks!

nerdCopter:
fixed. tested on both unsoldered FC with `undef`'s and LOS tested on headless whoop with `undef`'s

PR #970, opened 2024-02-23
fix compiler warnings - SmartAudio CMS menu FREQ

nerdCopter:
FREQ POR ; NEEDS TESTING,


nerdCopter:
- confirmed OSD menu items for SA/`FREQ` SA/`POR FREQ` working.
- when setup for freq rather than band/channel, i somtimes get lost signal. both master and this PR. so no change in behavior

PR #974, opened 2024-02-23
fix compiler warnings - board.c - strncpy, allow trailing null (5th character)

nerdCopter:
* board.c strncpy allow trailing null (5th character)
* fixes:
```C
./src/main/pg/board.c:47:9: warning: 'strncpy' output truncated before terminating nul copying 4 bytes from a string of the same length [-Wstringop-truncation]
   47 |         strncpy(boardConfig->manufacturerId, TARGET_MANUFACTURER_IDENTIFIER, MAX_MANUFACTURER_ID_LENGTH);
```


nerdCopter:
tested several FC found zero ill-effects.

PR #994, opened 2024-03-28
default PID Denom 2 for F411

### LINKED ISSUE #915: F411 are often 100% cpu
* reports that F411 are often 100% fixable with pid-demon 2 or less filtering.
* suggest some logic near https://github.com/emuflight/EmuFlight/blob/c9cb676f1ed13222e80e7ee8dc00b65cb806b489/src/main/flight/pid.c#L87 to define pid-denom 2 `if defined (STM32F411xE)`

nerdCopter:
- resolves https://github.com/emuflight/EmuFlight/issues/915

### DIFF
--- a/src/main/flight/mixer.c
+++ b/src/main/flight/mixer.c
@@ -100,11 +100,7 @@ void pgResetFn_motorConfig(motorConfig_t *motorConfig) {
     {
         motorConfig->minthrottle = 1070;
         motorConfig->dev.motorPwmRate = BRUSHLESS_MOTORS_PWM_RATE;
-#if defined(STM32F411xE)
-        motorConfig->dev.motorPwmProtocol = PWM_TYPE_DSHOT300;
-#else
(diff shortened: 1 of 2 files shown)


nerdCopter:
@Quick-Flash , @BeauBrewski  what do you think about this quick-fix for F411 denom?

nerdCopter:
- tested on CRAZYBEEF4DX (F411) resulted **denom 2 and dshot300**
- tested SKYSTARSF405-AIO resulted denom 1 and dshot600
- tested PYDRODRONEF7 resulted denom 1 and dshot600
- tested HELIOSPRING resulted denom 1 and dshot600

Quick-Flash:
No need for dshot 300 as a default. If hardware is setup poorly and but any dshot is required it'll run slower than dshot600 as well. So 600 is a better default imo.

nerdCopter:
okay, thanks, reverted that part.

PR #1073, opened 2024-07-09
more descriptive USBD_PRODUCT_STRING when not defined in target.h

nerdCopter:
- before: 
image

- after:
image

- when defined in target.h:
image


nerdCopter:
i attempted to [stringify](https://www.guyrutenberg.com/2008/12/20/expanding-macros-into-string-constants-in-) BOARD_NAME, but no matter any variations and attempts i could not such it to work.
i tried with the already existing [`STR()`](https://github.com/emuflight/EmuFlight/blob/master/src/main/common/utils.h#L39-L40)  and my own home-grown versions, but nothing worked.

PR #1085, opened 2024-09-06
[github] update actions/upload-artifact and actions/download-artifact to v3

nerdCopter:


nerdCopter:
resulted in expected results. will forego v4 for now (maybe forever since Emu is :skull:).
image

Issue #1115, opened 2026-04-22
drivers/bus_i2c_busdev.h: header not self-contained (uses extDevice_t without including drivers/bus.h)


nerdCopter:
AI Generated Issue

---

## Problem

`src/main/drivers/bus_i2c_busdev.h` uses `extDevice_t` and `uint8_t` in its function prototypes but does not `#include "drivers/bus.h"` (which defines `extDevice_t`) or any stdint/platform header. The header compiles only if every caller happens to include `drivers/bus.h` **before** `drivers/bus_i2c_busdev.h`.

```c
// src/main/drivers/bus_i2c_busdev.h
#pragma once

bool i2cBusWriteRegister(const extDevice_t *dev, uint8_t reg, uint8_t data);
bool i2cBusReadRegisterBuffer(const extDevice_t *dev, uint8_t reg, uint8_t *data, uint8_t length);
uint8_t i2cBusReadRegister(const extDevice_t *dev, uint8_t reg);
```

`extDevice_t` is defined in `drivers/bus.h`. `uint8_t` comes from `<stdint.h>` (usually pulled in transitively). Neither is made available by this header directly.

## Current state (not a bug today)

The existing callers of this header (`src/main/drivers/bus_i2c_busdev.c`, the three I2C sensor init files) happen to include `drivers/bus.h` first, so the latent issue doesn't fire. Compilation is clean today.

It **did** fire transiently during the Stage J I2C bus-resource split refactor (see PR for context, stage branched from #1114): adding the include to `sensors/compass.c` in the file's existing include-block order (bus_i2c → bus_spi → bus) placed `bus_i2c_busdev.h` before `bus.h`, which produced:

```
./src/main/drivers/bus_i2c_busdev.h:23:31: error: unknown type name 'extDevice_t'
```

The working fix in the Stage J commit was to reorder the include block in `compass.c` so `bus.h` precedes `bus_i2c_busdev.h`. That workaround is brittle — any future include ordering change or new caller can trigger the same error.

This is a **hygiene/forward-compat** issue, not a functional bug.

## Upstream comparison

Betaflight 4.5-maintenance has the same latent issue in the equivalent header (`betaflight/src/main/drivers/bus_i2c_busdev.h` is also header-only with no includes). Fixing on our side introduces a minor delta vs BF 4.5; so does leaving it and working around it per-caller. Either is acceptable; the fix is more robust.

## Proposed fix

Make the header self-contained:

```c
[...]

PR #1114, opened 2026-04-22
refactor(bus): Stage L-drivers - spiSetBusInstance BF-convention

nerdCopter:
## Summary

Fourth in a series of refactors aligning EmuFlight's bus / `extDevice_t` architecture with Betaflight 4.5-maintenance. Builds on #1108 (naming), #1111 (type sweep), #1113 (SPI bus-resource split).

This PR introduces the Betaflight-convention `spiSetBusInstance(extDevice_t *dev, uint32_t device)` entry point (id-based, returns `bool`) and migrates six drivers to use it. EmuFlight's existing `spiBusSetInstance(extDevice_t *dev, SPI_TypeDef *)` is reimplemented as a thin forwarder — the remaining 12 call sites (`gyro.c`, `accgyro_mpu.c`) keep using the legacy `MACRO_SPI_INSTANCE` pattern for now and will migrate in a later stage.

### What changes

- **`src/main/drivers/bus_spi.h`** — declare `bool spiSetBusInstance(extDevice_t *, uint32_t)`; existing `void spiBusSetInstance(extDevice_t *, SPI_TypeDef *)` kept for transitional callers.
- **`src/main/drivers/bus_spi.c`** — implement `spiSetBusInstance` with BF's validation (`device == 0 || device > SPIDEV_COUNT → false`). Reimplement `spiBusSetInstance` as a thin forwarder (`spiDeviceByInstance → SPI_DEV_TO_CFG → spiSetBusInstance`), with a fallback path for `SPIINVALID` that preserves pre-Stage-L behaviour (set busType + inline instance, leave `dev->bus` NULL).
- **`src/main/sensors/barometer.c`**, **`src/main/sensors/compass.c`**, **`src/main/drivers/max7456.c`**, **`src/main/drivers/nbd7456.c`** — collapse `spiBusSetInstance(dev, spiInstanceByDevice(SPI_CFG_TO_DEV(config->X)))` round-trips into `spiSetBusInstance(dev, config->X)`.
- **`src/main/drivers/rx/rx_spi.c`** — collapse `if (!spibus) return false; spiBusSetInstance(...)` into `if (!spiSetBusInstance(dev, rxSpiConfig->spibus)) return false;` (new function rejects `device == 0`, replacing the manual guard).
- **`src/main/drivers/flash.c`** — collapse manual `spiInstanceByDevice` / NULL-check / `spiBusSetInstance` into `if (!spiSetBusInstance(dev, flashConfig->spiDevice)) return false;`.

[...]

PR #1113, opened 2026-04-21
refactor(bus): phase 3 - bus-resource split infrastructure

nerdCopter:
## Summary

Third in a series of refactors aligning EmuFlight's bus / `extDevice_t` architecture with Betaflight 4.5-maintenance. Overall goal: reduce the paste-and-tweak delta when backporting BF MCU target configs and IMU driver updates, so future chip support (ICM-426xx families, newer MPU variants, newer F7/H7 targets, etc.) can be brought over with mechanical changes rather than cross-cutting edits.

This PR adds the BF split-bus infrastructure **without** removing the old inline SPI storage. It is deliberately infrastructure-only: no access sites are migrated, no behavior changes, existing code paths stay authoritative.

- ``struct busDevice_s`` defined as a separate type in bus.h (holds ``busType`` + SPI peripheral instance)
- ``static busDevice_t spiBusDevice[SPIDEV_COUNT]`` global in bus_spi.c, populated at the end of ``spiInit()``
- ``busDevice_t *spiBusByDevice(SPIDevice)`` accessor
- ``busDevice_t *bus;`` back-pointer field added to ``extDevice_t``, wired by ``spiBusSetInstance()``
- Existing inline ``SPI_TypeDef *instance`` / ``SPI_HandleTypeDef *handle`` in ``extDevice_t``'s ``extSpi_s`` union remain present and authoritative for all SPI access

Net effect for review: BF driver code that writes ``dev->bus->busType_u.spi.instance`` now resolves (``bus`` points to a valid populated ``busDevice_t``), while EmuFlight code paths are untouched. Cost is ~4–8 bytes RAM per ``extDevice_t`` + ``SPIDEV_COUNT × ~12`` bytes for the globals.

## Why infrastructure-only?

An earlier attempt (Stage I.4 in this branch, preserved in history + reverted at the tip) migrated all SPI access to ``dev->bus->...`` and removed the inline fields. It compiled clean across six verification targets but **bricked TUNERCF405 at boot when flashed**. Root cause was not identified; revert was the responsible call. Keeping the I.4 commit + its revert in history lets a future PR localise the break with per-file re-application + flash cycles.

Phase 3 at the revert tip captures ~80% of the architectural-alignment value (BF-style bus globals + back-pointer exist and are wired) without re-triggering the brick path. The remaining field-removal step is a RAM optimization and will be revisited separately.

## Verification

**Build matrix** (``CCACHE_DISABLE=1``, zero new compiler warnings, zero errors):

- ``make test`` PASS
[...]

PR #1116, opened 2026-04-22
refactor(bus): Stage J - I2C bus-resource split (infra + 3 sites)

nerdCopter:
## Summary

Fifth in a series of refactors aligning EmuFlight's bus / `extDevice_t` architecture with Betaflight 4.5-maintenance. Builds on #1108 (naming), #1111 (type sweep), #1113 (SPI bus-resource split), #1114 (SPI init-site migration).

This PR introduces the Betaflight-convention `i2cBusSetInstance(extDevice_t *dev, uint32_t device)` entry point and migrates the three I2C init sites (baro, compass, MPU-slave-over-I2C) to use it. It is the I2C parallel of phase 3's SPI infra (#1113) plus Stage L-drivers' init-site migration (#1114), combined into one PR because the I2C call-site count is small.

### What changes

- **`src/main/drivers/bus_i2c_busdev.h`** — declare `bool i2cBusSetInstance(extDevice_t *, uint32_t)`. Matches BF 4.5 signature exactly.
- **`src/main/drivers/bus_i2c_busdev.c`** — implement `i2cBusSetInstance`: function-local `static busDevice_t i2cBus[I2CDEV_COUNT]` (matches BF layout — BF uses function-local for I2C and file-global for SPI). Validates `device` is in `1..I2CDEV_COUNT`; returns `false` out of range. Populates BOTH `dev->bus->*` (BF-parity) AND inline `dev->busType` + `dev->busType_u.i2c.device` (transitional — EmuFlight's existing `i2cBus*` accessors still read inline fields).
- **`src/main/sensors/barometer.c`** — replace `busType = BUS_TYPE_I2C; busType_u.i2c.device = I2C_CFG_TO_DEV(...)` writes with `i2cBusSetInstance(&dev->dev, barometerConfig()->baro_i2c_device)`. Address write remains inline.
- **`src/main/sensors/compass.c`** — same pattern. Also reorders the `bus_i2c_busdev.h` include to come after `bus.h` (header is not self-contained — uses `extDevice_t` without including `bus.h`; same latent issue as BF 4.5-maintenance; per the scope-boundary rule from #1111, fixed the caller, filed the header fix as #1115).
- **`src/main/drivers/accgyro/accgyro_mpu.c`** — same pattern for the I2C-over-MPU init path. `MPU_I2C_INSTANCE` is in 0-indexed DEV form (`I2CDEV_1` = 0), wrapped with `I2C_DEV_TO_CFG()` for the 1-indexed parameter.

**Intentionally unchanged:**

- `io/dashboard.c` (`dashBoardBus`) — never goes through `busRead/Write` dispatch (the OLED driver reads inline fields directly via `i2cWrite`), so migrating would add risk for no BF-paste value.
- Access-site migration of `i2cBusWriteRegister` / `i2cBusReadRegisterBuffer` / `i2cBusReadRegister` from inline `dev->busType_u.i2c.device` to `dev->bus->busType_u.i2c.device`. This is the I2C equivalent of Stage I.4's SPI field removal — same brick class — deferred until a reproducible hypothesis exists (see phase-3 post-mortem referenced in #1113).

**No target.h changes in this PR.**

## Verification

**Build matrix** (`CCACHE_DISABLE=1`, zero new compiler warnings, zero errors):

- `make test` PASS (38 binaries)
- `HELIOSPRING` (STM32F7, IMUF9001)
- `TUNERCF405` (STM32F405 — I.4 brick-victim gate)
- `SKYSTARSF405AIO` (STM32F405, BMI270)
- `PYRODRONEF7` (STM32F7, MPU6000)
- `MATEKF411` (STM32F411)
- `NBDHMBF4PRO` (STM32F405)
- `FOXEERF722V4` (STM32F722)
- `FOXEERF405` (STM32F405)

**Bench verification** (2026-04-22, pre-commit on content-equivalent tree):

- **HELIOSPRING** — PASS.
- **TUNERCF405** — PASS. Brick-victim gate cleared.
- **SKYSTARSF405AIO** — PASS.
- **PYRODRONEF7** — PASS.
- **FOXEERF405** — PASS.

Five-target bench across three MCU families (F4 / F7 / F7-IMUF) and two gyro families (MPU-series + BMI270) — same matrix as #1113 and #1114.

## Why this approach

Phase 3 (#1113) added the infrastructure for BF's split-bus architecture on the SPI side; Stage L-drivers (#1114) migrated the SPI init sites to the BF-convention `spiSetBusInstance`. This PR applies the same pattern to I2C — but because the I2C init-site count is small (3 sites, all `busType` + `.i2c.device` double-writes at init), the infra and the call-site migration are bundled into one PR rather than split across two as on the SPI side.

Keeping the inline `busType_u.i2c.device` writes alongside the new `dev->bus->` population means existing EmuFlight `i2cBus*` accessors (which still read inline) work unchanged, and BF-sourced driver code that reads `dev->bus->busType_u.i2c.device` also resolves. A future stage will migrate the access sites — deferred per the Stage I.4 risk class.

## Test plan

- [x] \`make test\`
- [x] Eight-target build matrix with zero warnings
- [x] TUNERCF405 bench (brick-victim gate)
- [x] HELIOSPRING bench
- [x] SKYSTARSF405AIO bench
- [x] PYRODRONEF7 bench
- [x] FOXEERF405 bench

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

## Release Notes

* **Refactor**
  * Streamlined I2C device initialization across sensor drivers for improved code consistency and maintainability. Internal infrastructure updates with no impact to end-user functionality.

### DIFF
--- a/src/main/drivers/accgyro/accgyro_mpu.c
+++ b/src/main/drivers/accgyro/accgyro_mpu.c
@@ -34,6 +34,7 @@
 
 #include "drivers/bus.h"
 #include "drivers/bus_i2c.h"
+#include "drivers/bus_i2c_busdev.h"
 #include "drivers/bus_spi.h"
 #include "drivers/exti.h"
 #include "drivers/io.h"
@@ -432,7 +433,7 @@ void mpuDetect(gyroDev_t *gyro) {
         gyro->dev.busType = BUS_TYPE_I2C;
     }
     if (gyro->dev.busType == BUS_TYPE_I2C) {
-        gyro->dev.busType_u.i2c.device = MPU_I2C_INSTANCE;
+        i2cBusSetInstance(&gyro->dev, I2C_DEV_TO_CFG(MPU_I2C_INSTANCE));
         gyro->dev.busType_u.i2c.address = MPU_ADDRESS;
         uint8_t sig = 0;
         bool ack = busReadRegisterBuffer(&gyro->dev, MPU_RA_WHO_AM_I, &sig, 1);
--- a/src/main/drivers/bus_i2c_busdev.c
+++ b/src/main/drivers/bus_i2c_busdev.c
@@ -42,4 +42,26 @@ uint8_t i2cBusReadRegister(const extDevice_t *dev, uint8_t reg) {
     i2cRead(dev->busType_u.i2c.device, dev->busType_u.i2c.address, reg, 1, &data);
     return data;
 }
+
+bool i2cBusSetInstance(extDevice_t *dev, uint32_t device)
+{
+    // I2C bus-resource array; function-local static matches BF 4.5-maintenance
+    // layout (bus_i2c_busdev.c:i2cBusSetInstance). Persistent across calls.
+    static busDevice_t i2cBus[I2CDEV_COUNT];
+
+    if ((device < 1) || (device > I2CDEV_COUNT)) {
+        return false;
+    }
+
+    dev->bus = &i2cBus[I2C_CFG_TO_DEV(device)];
+    dev->bus->busType = BUS_TYPE_I2C;
+    dev->bus->busType_u.i2c.device = I2C_CFG_TO_DEV(device);
+
+    // Transitional: keep inline fields authoritative until access-site
+    // migration lands. EmuFlight's i2cBus* accessors still read inline.
+    dev->busType = BUS_TYPE_I2C;
+    dev->busType_u.i2c.device = I2C_CFG_TO_DEV(device);
+
+    return true;
+}
 #endif
--- a/src/main/drivers/bus_i2c_busdev.h
+++ b/src/main/drivers/bus_i2c_busdev.h
@@ -23,3 +23,9 @@
 bool i2cBusWriteRegister(const extDevice_t *dev, uint8_t reg, uint8_t data);
 bool i2cBusReadRegisterBuffer(const extDevice_t *dev, uint8_t reg, uint8_t *data, uint8_t length);
 uint8_t i2cBusReadRegister(const extDevice_t *dev, uint8_t reg);
+// Associate a device with an I2C bus (BF 4.5-maintenance signature). device
+// is 1-indexed CFG form (1..I2CDEV_COUNT); returns false on out-of-range.
+// Populates dev->bus back-pointer AND inline dev->busType / busType_u.i2c.device
+// so both BF-sourced access paths and EmuFlight's existing inline accessors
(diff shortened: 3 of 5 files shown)


nerdCopter:
@coderabbitai — triage per the BF-alignment scope rule this refactor series follows (also applied to #1111, #1114): new-code findings are actionable only if Betaflight 4.5-maintenance does the equivalent; otherwise they create a future divergence to maintain.

I checked all three findings against BF 4.5-maintenance:

**1. Nitpick on `src/main/sensors/compass.c:127-131` (propagate `i2cBusSetInstance` failure)** — out of scope.

BF's `src/main/sensors/compass.c:176`:
```c
case BUS_TYPE_I2C:
    i2cBusSetInstance(dev, compassConfig()->mag_i2c_device);
    dev->busType_u.i2c.address = compassConfig()->mag_i2c_address;
    break;
```
BF does **not** check the return value; the SPI branch right below it does (`if (!spiSetBusInstance(...)) return false;`), exactly mirroring the asymmetry you flagged in our code. The claim that an early-return \"would match BF 4.5-m semantics\" is not accurate for I2C — adding it would diverge from BF.

**2. Nitpick on `src/main/drivers/accgyro/accgyro_mpu.c:430-458` (guard WHO_AM_I probe on I2C failure)** — out of scope.

Same reasoning. BF doesn't wrap its I2C probe path with a `if (i2cBusSetInstance(...))` guard, and EmuFlight's MPU-over-I2C detection falls through to the SPI probe on failure via the existing `busType` logic — behavior that's preserved by this PR.

**3. Minor on `src/main/drivers/bus_i2c_busdev.c:66` (document behavior change)** — out of scope.

BF's `bus_i2c_busdev.c:75` `i2cBusSetInstance` is structurally identical to ours (same bounds check, same silent early-return on `device < 1 || device > I2CDEV_COUNT`) and carries no such documentation. Adding EF-only prose to a BF-parity function creates maintenance drift. For reference: #1114's SPI-side return-check adoption was in scope because BF's `spiSetBusInstance` **callers** check the return; that test is what distinguishes in-scope from out-of-scope here.

Downstream, the semantic difference you called out (out-of-range now leaves `busType == BUS_TYPE_NONE` instead of writing `BUS_TYPE_I2C` with `device = I2CINVALID`) is not a regression: both paths fail I2C dispatch, just at slightly different layers. BF ships the same behavior.

Holding this PR at the current two commits. Will not apply any of the three suggestions in this stage.

PR #1168, opened 2026-05-09
fix(rx): JETI ExBus unconditionally use SERIAL_BIDIR

### LINKED ISSUE #704: Jeti EXBUS not working on HGLRC Zeus F745, but does on Betaflight
Jeti EXBUS receiver protocol doesn't seem to be working in EmuFlight.

I'm new to EmuFlight, setting it up for the first time.  Therefore, I'm not sure if this is an issue with EmuFlight firmware, or the specific board I'm using.  I suspect it is EmuFlight.  I'm able to flash to Betaflight, and everything works fine there.

* I have an HGLRC Zeus F745 20x20 FC with my receiver wired to TX on UART1 (TX1, which provides telemetry over EXBUS on Betaflight)
* The board is flashed with 0.4.0 with target HGLRCF722.
* I have 'Serial RX' turned on for UART1 in the Ports tab
* On the Configuration tab under Receiver, I have "Serial-based receiver(SPEKSAT, SBUS, SUMD)" selected
* For the Serial Provider, I have "JETIEXBUS" selected.

After Saving, rebooting/repowering, and going to the Receiver tab, I get no receiver output.

With this exact configuration under Betaflight, this all works, and is the standard setup for using Jeti Receivers on Betaflight.

I notice on the Ports tab, the Telemetry Output dropdown does not include Jeti, which seems to indicate to me that remnants of Jeti support from Betaflight are still in EmuFlight, but is no longer working.

nerdCopter:
AI Generated pull-request

Fixes #704

## Root Cause

JETI ExBus is an inherently half-duplex protocol. The serial port was opened with `SERIAL_BIDIR` only when `rxConfig->halfDuplex` was set by the user. Without it, the UART remained in full-duplex mode and the FC could not drive the TX line for telemetry responses.

## Changes (BF 4.5-maintenance parity)

**`src/main/rx/jetiexbus.c`**
- Always pass `SERIAL_BIDIR` to `openSerialPort` — removes the `rxConfig->halfDuplex` conditional (root-cause fix; matches BF `rx/jetiexbus.c`)
- Remove `serialSetMode(jetiExBusPort, MODE_RX)` after init — direction is hardware-managed with `SERIAL_BIDIR`
- Add per-frame `jetiExBusFrameMaxSize` tracking and overrun detection in the ISR to prevent buffer overwrite on noisy lines
- Use `microsISR()` / `cmpTimeUs()` for overflow-safe ISR timing
- Move `jetiExBusTimeLast = now` to after the gap-reset block (BF parity)
- Reuse already-captured `now` for `jetiTimeStampRequest` (removes second `micros()` call)
- Correct wiring comment: EX Bus connects to Serial TX, not RX

**`src/main/telemetry/jetiexbus.c`**
- Remove `serialSetMode(jetiExBusPort, MODE_TX)` and `serialSetMode(jetiExBusPort, MODE_RX)` from `handleJetiExBusTelemetry` — direction switching is not needed when the port is opened with `SERIAL_BIDIR`

## BF Equivalence

Diff verified against `betaflight/src/main/{rx,telemetry}/jetiexbus.c` on `4.5-maintenance`. The `SERIAL_BIDIR` change and removal of `serialSetMode` calls are direct backports.

## Coverage gaps

F446 and F7X5XG not exercised by CI (no targets in build set).

## Tier Classification

**Tier 2 — behavioral change** (device-init path, serial port options).  
Bench verification required on at least one F4 and one F7 target before merge.  
Build gate: zero warnings across all bench + compile targets — confirmed locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Jeti EX Bus receiver frame assembly timing accuracy for more reliable serial communication.
  * Simplified telemetry transmission flow by removing unnecessary serial port mode switching operations.
  * Enhanced serial port initialization for bidirectional communication support.

[Review Change Stack](https://app.coderabbit.ai/change-stack/emuflight/EmuFlight/pull/1168)

### DIFF
--- a/src/main/rx/jetiexbus.c
+++ b/src/main/rx/jetiexbus.c
@@ -33,9 +33,7 @@
  * Number of channels: 16
  *
  * Connect as follows:
- * Jeti EX Bus -> Serial RX (connect directly)
- * Serial TX -> Resistor(2k4) ->Serial RX
- * In jeti pdf it is different, but if the resistor breaks, the receiver continues to operate.
+ * Jeti EX Bus -> Serial TX (connect directly)
  *
  */
 
@@ -51,6 +49,7 @@
 
 #include "pg/rx.h"
 
+#include "common/time.h"
 #include "common/utils.h"
 
 #include "drivers/time.h"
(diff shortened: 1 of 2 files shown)


nerdCopter:
confirmed ELRS still works unaffected.

PR #1191, opened 2026-05-15
fix(bus): guard DMA HAL types against SIMULATOR_BUILD for SITL compile

nerdCopter:
AI Generated pull-request

## Summary

- `bus.h` guards around `DMA_InitTypeDef` / `LL_DMA_InitTypeDef` / `dmaChannelDescriptor_t` used `#ifndef UNIT_TEST` alone, which does not exclude SITL builds
- SITL defines `SIMULATOR_BUILD` (injected at `make/source.mk:183`), not `UNIT_TEST` — so the STM32 HAL types were exposed to the host build with no HAL headers available
- Regression introduced in #1121; `make SITL` has been broken on master since 2026-04-25
- Fix extends both guards to `#if !defined(UNIT_TEST) && !defined(SIMULATOR_BUILD)`, matching the existing pattern at `platform.h:25`

## Root cause

`make/source.mk:183` injects `-DSIMULATOR_BUILD` for SITL. `UNIT_TEST` is only defined by the unit-test Makefile path. The two defines are parallel but separate; both must be excluded when guarding against host/stub builds that lack STM32 HAL headers.

## Test plan

- [x] `make SITL` succeeds on this branch (verified locally)
- [x] Pattern matches `platform.h:25`: `#if !defined(UNIT_TEST) && !defined(SIMULATOR_BUILD)`
- [ ] CI ARM builds (targets-group-*) — no change to ARM paths; compile-only verification

## Summary by CodeRabbit

* **Chores**
  * Updated build configuration to exclude DMA-related components from simulator builds in addition to unit test builds.

[Review Change Stack](https://app.coderabbit.ai/change-stack/emuflight/EmuFlight/pull/1191)

### DIFF
--- a/src/main/drivers/bus.h
+++ b/src/main/drivers/bus.h
@@ -69,8 +69,8 @@ typedef struct busDevice_s {
     } busType_u;
     bool useDMA;
     uint8_t deviceCount;
-#ifndef UNIT_TEST
-    // dmaChannelDescriptor_t requires dma.h which is not available in unit test builds
+#if !defined(UNIT_TEST) && !defined(SIMULATOR_BUILD)
+    // dmaChannelDescriptor_t requires dma.h which is not available in unit test / SITL builds
     dmaChannelDescriptor_t *dmaTx;
     dmaChannelDescriptor_t *dmaRx;
 #if defined(STM32F7)
@@ -80,7 +80,7 @@ typedef struct busDevice_s {
     DMA_InitTypeDef    *initTx;
     DMA_InitTypeDef    *initRx;
 #endif
-#endif // UNIT_TEST
+#endif // !UNIT_TEST && !SIMULATOR_BUILD
     volatile struct busSegment_s *volatile curSegment;
     bool initSegment;
 } busDevice_t;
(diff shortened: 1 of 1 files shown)


nerdCopter:
AI Generated comment

@CodeRabbitAI — addressing the outside-diff finding on `bus.h` lines 27–29 (guard around `#include "drivers/dma.h"`).

**Finding is not applicable.** `make SITL` was verified to succeed on this branch both before and after re-examining this include. Here is why `dma.h` does not need to be excluded from `SIMULATOR_BUILD`:

`src/main/target/SITL/target.h` already provides host stubs for every type that `dma.h` uses unconditionally:

| Type | Stub location |
|------|---------------|
| `IRQn_Type` | `SITL/target.h:167` |
| `DMA_TypeDef` | `SITL/target.h:193` |
| `DMA_Channel_TypeDef` | `SITL/target.h:197` |

The only member that references `DMA_Stream_TypeDef` is inside `#if defined(STM32F4) || defined(STM32F7)` in `dma.h`, which is never active for a SITL host build.

The types that actually caused the original SITL compile failure — `DMA_InitTypeDef` and `LL_DMA_InitTypeDef` — are **not** in `dma.h`; they were in the `busDevice_t` and `extDevice_t` struct bodies in `bus.h` itself. Those are covered by this PR's fix (`#if !defined(UNIT_TEST) && !defined(SIMULATOR_BUILD)`).

Excluding `dma.h` from SITL would be unnecessary and could silently break SITL compilation units that legitimately reference `dmaChannelDescriptor_t` or other types via this transitive include.

Issue #1184, opened 2026-05-13
[tune] investigate stable-mode D-terms.


nerdCopter:
Observation: when testing dev-builds i noticed that acro-mode behaved exactly as expected, but when switching to stable-mode it was rather easy to induce a climb due to what i expect is the high-ish D-terms.

nerdCopter:
@coderabbitAI, investigate the codebase on why this may be.  "stable-mode" = Angle-mode in this description.
My configuration/tune is:  [EMUF_cli_HELIO_V1_20260503_130336.txt](https://github.com/user-attachments/files/27730198/EMUF_cli_HELIO_V1_20260503_130336.txt)
propose  solution that makes stable-mode a bit safer (even if it's simply a default set of stable-mode configuration values).  stable-mode is typically switched into during GPS RTH and i recall someone mentioning a similar issue that required regaining manual control to save the aircraft.

<img width="689" height="385" alt="Image" src="https://github.com/user-attachments/assets/8b6d43f6-df07-4682-9495-3f418e4052b8" />

nerdCopter:
@coderabbitAI, modify your analysis based on the .txt file only. the screenshot was not a basis, my mistake to paste a screenshot of defaults.  my tune is not defaults, but the stable-mode values in my .txt are likely still defaults.

nerdCopter:
@coderabbitAI, are the stable-mode PID different than acro-mode PIDS?  are the stable-mode PID exposed via MSP?  i dont think the EmuConfigurator dispalys any other PIDs that what was given in the screenshot of the second comment.

nerdCopter:
@coderabbitAI, 
1) regarding `i_level_low` and `i_level_high` -- ar there any legacy, unused PID MSP that can be-reused?  i think EmuFlight stable mode was changed long ago, but it seems that old pre-change (old `butterflight`/`betaflight` [circa version 3.5.7])  would have has exposed stable-mode PIDs to MSP.  If we can re-use an unused slot, this might solve the firmware/configure issue without damaging in-the-wild builds and not necessarily requires a re-work or re-release,

2) I also envision that since we know they are partially exposed, then i want to add a new sub-tab on the PID-tuning tab of EmuConfigurator and move all stable-mode PID/setting to it.

nerdCopter:
[...]

PR #1257, opened 2026-06-11
fix: use pure P-term for directFF in angle/horizon mode (rise/raise/fly-off bias)

nerdCopter:
AI Generated pull-request

## Summary

In ANGLE_MODE and HORIZON_MODE, \`directFeedForward\` was computed from \`currentPidSetpoint\` **after** D and F transients from the angle controller had been accumulated into the setpoint. During spring oscillations on roll/pitch sticks the angle controller's derivative contributions produce transient spikes in \`currentPidSetpoint\`, which amplified \`directFeedForward\` asymmetrically — causing motor floor-clipping on one side without equivalent ceiling-clipping on the other, and producing a net upward thrust bias (unintended climb).

**Fix (3 hunks in \`pid.c\`):**

- **Hunk 1** — declare \`static FAST_RAM_ZERO_INIT float angleModeP[2]\` alongside \`directFF\`, sized for ROLL and PITCH only (YAW never enters \`pidLevel()\`).
- **Hunk 2** — inside \`pidLevel()\`, capture the pure P-error blend (\`p_term_low + p_term_high\`) into \`angleModeP[axis]\` *before* the angle-controller D and F terms are added to \`currentPidSetpoint\`.
- **Hunk 3** — inside \`pidController()\`, when in ANGLE_MODE or HORIZON_MODE on roll/pitch (\`axis <= FD_PITCH\`), compute \`directFeedForward\` from \`angleModeP[axis]\` (clean tilt-proportional demand) rather than the full \`currentPidSetpoint\`. Acro mode and yaw are unchanged.

**Background — why \`directFeedForward\` exists in angle mode:**
\`directFF\` in roll/pitch is intentionally active only while angle/horizon mode is engaged — it is zeroed for acro. The gain is governed by CLI parameters \`df_angle_low\` / \`df_angle_high\`, which blend based on \`errorAnglePercent\` (strongest near level, fading toward max tilt). The effective multiplier is \`DIRECT_FF_SCALE × df_angle_low = 0.005 × df_angle_low\` — so the default of 70 yields a 0.35× gain, not 0.70×. This provides faster setpoint tracking in angle mode and is deliberate codebase behaviour.

**CLI parameter reference** — internal blackbox field `levelPIDLOW` packs P/D/F as `"P,D,F"`; the I-slot is repurposed to store `df_angle_low`. The corresponding CLI settings are:

| Blackbox field | CLI parameter |
|---|---|
| `levelPIDLOW.P` | `p_angle_low` |
| `levelPIDLOW.D` | `d_angle_low` |
| `levelPIDLOW.F` (third value) | `f_angle` |
| `levelPIDLOW.I` (repurposed) | `df_angle_low` |

The rest of this description uses the CLI parameter names (`p_angle_low`, `df_angle_low`) since those are what a pilot actually sets.

## Hardware Test Results (HELIOSPRING — F405)

Three-state blackbox comparison using stick spring oscillation maneuver:

| State | `p_angle_low` | `df_angle_low` | Fix applied | Peak P0 | Simultaneous floor+ceil clip | Observed result |
|---|---|---|---|---|---|---|
| Original defaults | 100 | 70 | no | −748 | sustained episodes | aggressive ascent |
| Softer tune only | 70 | 40 | no | <−527 | 0 rows | less aggressive, minimal ascent |
| Fix + softer tune | 70 | 40 | yes | −527 | 36 rows (0.05%) | noticeably more controllable/tolerable |

**Root cause of ascent:** `p_angle_low=100` (internally `levelPIDLOW.P`) generates rate setpoints that push P0=−748 in the acro PID loop, driving one motor to 2047 (ceiling) and another to 163 (floor) simultaneously. Motor headroom is asymmetric around hover (~620): floor = 457, ceiling = 1427. When both clip together, average thrust rises ~70% above hover. `directFeedForward` contributes ~12% of P0 at peak — this fix removes the D-contamination from that contribution, which is correct and measurably improves saturation onset.

**Residual ascent after fix + softer tune** is present but significantly reduced. After motor saturation is effectively eliminated (0.05% simultaneous clipping), residual climb in angle mode likely originates from `throttle_boost` or `emu_gravity` interaction — outside the scope of this fix.

## Tuning guidance — motor saturation in angle mode

The defaults (`df_angle_low = 70`, `p_angle_low = 100`) are sized for small whoops (1–2 inch). On larger frames, simultaneous motor floor+ceiling clipping is the mechanism — not the gains themselves. Lowering either CLI setting reduces the rate demand and keeps motors out of saturation.

| Frame size | `df_angle_low` | `p_angle_low` | Notes |
|---|---|---|---|
| 1–2" whoops | 70 (default) | 100 (default) | Motor headroom large relative to PID demand |
| 3–4" | 40–50 | 70–80 | |
| 5–6" freestyle | 25–35 | 55–65 | |
| 7–10"+ | 20–30 | 50–60 | GPS Rescue: prioritise stability; `p_angle_low`=60–80, `df_angle_low`=40–50 |

Reducing `p_angle_low` softens the self-levelling snap — tune to taste.

Partially addresses #1184
Reference: https://github.com/emuflight/EmuFlight/issues/1184#issuecomment-4681039700

## Test plan

- [x] Compile-gate: `make HELIOSPRING FOXEERF722V4 STELLARH7DEV`
- [x] Unit tests: `make test`
- [x] Hardware: HELIOSPRING — three-state blackbox comparison confirms measurable reduction in motor saturation and ascent severity
- [x] Acro mode: directFF behaviour unchanged in rate/acro flight

### DIFF
--- a/src/main/flight/pid.c
+++ b/src/main/flight/pid.c
@@ -353,6 +353,7 @@ typedef struct pidCoefficient_s {
 
 static FAST_RAM_ZERO_INIT pidCoefficient_t pidCoefficient[XYZ_AXIS_COUNT];
 static FAST_RAM_ZERO_INIT float directFF[3];
+static FAST_RAM_ZERO_INIT float angleModeP[2];
 static FAST_RAM_ZERO_INIT float maxVelocity[XYZ_AXIS_COUNT];
 static FAST_RAM_ZERO_INIT float feathered_pids;
 static FAST_RAM_ZERO_INIT float setPointPTransition[XYZ_AXIS_COUNT];
@@ -540,6 +541,7 @@ static float pidLevel(int axis, const pidProfile_t *pidProfile, const rollAndPit
     attitudePrevious[axis] = getAngleModeAngles(axis);
 
     currentPidSetpoint = p_term_low + p_term_high;
+    angleModeP[axis] = p_term_low + p_term_high;
     currentPidSetpoint += d_term_low + d_term_high;
     currentPidSetpoint += f_term_low;
 
@@ -853,7 +855,12 @@ void pidController(const pidProfile_t *pidProfile, const rollAndPitchTrims_t *an
         // this feedforward is literally setpoint * feedforward
         // since yaw acts different this will only work for yaw
         // actually this also works in angle mode so pitch and roll have angle mode values
-        float directFeedForward = currentPidSetpoint * directFF[axis];
+        float directFeedForward;
+        if ((FLIGHT_MODE(ANGLE_MODE) || FLIGHT_MODE(HORIZON_MODE)) && axis <= FD_PITCH) {
+            directFeedForward = angleModeP[axis] * directFF[axis];
+        } else {
+            directFeedForward = currentPidSetpoint * directFF[axis];
+        }
 
 #ifdef USE_YAW_SPIN_RECOVERY
         if (gyroYawSpinDetected()) {


nerdCopter:
AI Generated comment

Reverting to draft — hardware testing confirms spring oscillation ascent persists with this fix applied. Removing D/F terms from `directFeedForward` in angle/horizon mode is not sufficient to mitigate the climb. Further analysis and a deeper fix are needed before this is ready for review.

nerdCopter:
AI Generated comment

## Follow-up: fix scope does not cover GPS_RESCUE_MODE (RTH)

Verified against current master `src/main/flight/pid.c`:

- `GPS_RESCUE_MODE` calls `pidLevel()` directly (line ~719), and `pidLevel()` unconditionally recomputes `directFF[axis]` from `df_angle_low`/`df_angle_high` (line ~559-560) — RTH genuinely uses `p_angle_low`, `d_angle_low`, and `df_angle_low`, same as ANGLE_MODE/HORIZON_MODE.
- The pure-P-term fix in this PR only gates the `directFeedForward` *source* (`angleModeP[axis]` vs `currentPidSetpoint`) on `(FLIGHT_MODE(ANGLE_MODE) || FLIGHT_MODE(HORIZON_MODE)) && axis <= FD_PITCH` (line ~859). `GPS_RESCUE_MODE` falls into the `else` branch and still uses the un-fixed, D/F-contaminated `currentPidSetpoint` as the directFF multiplier.

So RTH is structurally exposed to the same directFF computation this PR changes for angle/horizon — it just doesn't get the fix.

## Possible shared root cause with reported RTH flyoff/rise

There is at least one user report of unintended climb/flyoff when GPS Rescue (RTH) activates. Per the hardware test data already on this PR, the dominant saturation mechanism for the original bug was **not** the directFF D/F-contamination itself (only ~12% of P0 at peak) — it was the rate-setpoint magnitude from `pidLevel()`'s P-term (`p_angle_low`), which can drive simultaneous motor floor+ceiling clipping given the asymmetric headroom around hover.

`src/main/flight/gps_rescue.c` commands a sustained pitch angle up to `gpsRescueConfig()->angle` (default **32°**) during the fly-home phase (`rescueState.intent.maxAngleDeg = gpsRescueConfig()->angle`), fed through the same `pidLevel()` / `p_angle_low` path. A sustained large-angle command combined with a wind-correction transient could plausibly reach a comparable P0 magnitude to the stick-spring-oscillation case, via the same motor-headroom-asymmetry mechanism — independent of the directFF path this PR addresses.

This is a hypothesis, not a confirmed finding — no hardware test exists for the GPS Rescue case specifically. Flagging for whoever picks up the deeper root-cause investigation already noted as an open item on this PR (residual ascent after this fix + softer tune, likely `throttle_boost`/`emu_gravity`-related).

nerdCopter:
AI Generated comment

## Follow-up: mechanism confirmed for GPS Rescue (RTH), not just gate-structure speculation

Read the full `pidLevel()` body in master `pid.c`. GPS Rescue's angle demand is not isolated from the D/F-transient mechanism this PR targets — it goes through the identical computation:

```c
angle += gpsRescueAngle[axis] / 100;   // GPS Rescue's commanded angle folds into the SAME "angle" var as stick input
...
float errorAngle = angle - currentAngle;
...
d_term_low  = (1 - errorAnglePercent) * (attitudePrevious[axis] - getAngleModeAngles(axis)) * D_angle_low;
d_term_high = errorAnglePercent       * (attitudePrevious[axis] - getAngleModeAngles(axis)) * D_angle_high;
...
f_term_low = (angle - previousAngle[axis]) * F_angle / dT;   // derivative of the setpoint itself
```

`d_term` is the derivative of the craft's actual attitude change; `f_term` is the derivative of the setpoint itself. Neither distinguishes stick-sourced demand from GPS-Rescue-sourced demand — both respond to whatever is driving `angle`, and `gpsRescueAngle[axis]` drives it the same way stick deflection does.

`gps_rescue.c` increments its angle demand in small bounded steps per control loop rather than an abrupt stick-spring release, so the transient is normally smaller/slower. But under a fast correction (wind gust, a jump in GPS position, mode-entry from an off-nominal attitude), the D/F transient magnitude could approach the stick case — and RTH is excluded from this PR's gate (`(FLIGHT_MODE(ANGLE_MODE) || FLIGHT_MODE(HORIZON_MODE)) && axis <= FD_PITCH`, `pid.c` ~L859), so it never gets the pure-P-term fix regardless of magnitude.

Upgrading prior note from "plausible" to "mechanism confirmed by code; magnitude during real RTH flight unconfirmed — no hardware test exists for this case."

nerdCopter:
AI Generated comment

## Cross-reference: independent validation of #1257's tuning values

CodeRabbitAI independently validated the frame-size tuning table (derived from this PR's hardware test data) against current `pid.c`/master when it was added to the EmuConfigurator UI:

https://github.com/emuflight/EmuConfigurator/pull/648#issuecomment-5072251721

Confirms: 1-2" whoop defaults match `PID_LEVEL_LOW` reset values (`p_angle_low=100`, `df_angle_low=70`), `DIRECT_FF_SCALE=0.005f` produces the 0.35x effective multiplier (not 0.70x), and the 3-4"/5-6"/7-10"+ progression is internally consistent with the values documented on this PR.

PR #1278, opened 2026-06-18
fix: DShot beacon fires even when battery alarm owns beeper

nerdCopter:
AI Generated pull-request

## Summary

\`BEEPER_RX_SET\` (manual beeper AUX switch, priority 9) is lower priority
than \`BEEPER_BAT_CRIT_LOW\` (6) and \`BEEPER_BAT_LOW\` (7) in the beeper
priority table. When a low-battery alarm is active it wins arbitration and
\`currentBeeperEntry->mode\` is never \`BEEPER_RX_SET\`, so the DShot motor
beacon never fires — a downed model with a dead battery cannot be found by
its motors even with the beeper switch held on.

\`BEEPER_RX_LOST\` (priority 1) was already immune to this. Only \`RX_SET\`
was affected.

**Fix**: add \`dshotBeaconPendingMode\` — a static set in \`beeper()\` whenever
\`BEEPER_RX_SET\` or \`BEEPER_RX_LOST\` is called, **before** the priority
arbitration early-return, independently of which entry wins buzzer priority.
\`beeperUpdate()\` reads and clears this each cycle; the DShot motor-tone block
uses it instead of \`currentBeeperEntry->mode\`. \`beeperSilence()\` also clears
it so an explicit silence cannot leave a stale beacon pending.

The physical buzzer still plays the highest-priority alarm (battery warning
continues to sound). Only the motor beacon output is decoupled from buzzer
priority. All existing guards are unchanged: \`!areMotorsRunning()\`, 1.2 s
disarm guard, \`!isTryingToArm()\`, \`dshotBeaconOffFlags\`.

BF 4.5-maintenance has the same priority-crowding behaviour (RX_SET=10,
BAT_CRIT_LOW=7, BAT_LOW=8) with no equivalent fix.

## Test plan
- [x] Build passes: \`HELIOSPRING\` (F4, DShot)
- [x] Unit tests pass: \`make test\` — 8 cases in \`beeper_dshot_beacon_unittest\`
  covering the key regression, both alarm variants, all three safety guards,
  and \`beeperSilence()\` clearing the pending mode
- [ ] Hardware verified on HELIOSPRING: DShot motor beacon fires while
      battery alarm is active and beeper AUX switch is on

## Release Notes

* **New Features**
  * Motor beacon coordination: DShot motor beacon tones now emit independently from the buzzer system, ensuring motor feedback remains available even when high-priority alarms override the buzzer during critical alerts and alarm conditions.

* **Tests**
  * Added unit tests validating motor beacon behavior under alarm arbitration, including safety guards and edge-case handling.

### DIFF
--- a/src/main/io/beeper.c
+++ b/src/main/io/beeper.c
@@ -270,17 +270,19 @@ void beeper(beeperMode_e mode) {
         }
         break;
     }
+    // Capture beacon request before the priority-loss early-return so the DShot motor tone
+    // fires even when a higher-priority alarm (e.g. BAT_CRIT_LOW) owns the physical buzzer.
+#ifdef USE_DSHOT
+    if (mode == BEEPER_RX_SET || mode == BEEPER_RX_LOST) {
+        dshotBeaconPendingMode = mode;
+    }
+#endif
     if (!selectedCandidate) {
         return;
     }
     currentBeeperEntry = selectedCandidate;
     beeperPos = 0;
     beeperNextToggleTime = 0;
-#ifdef USE_DSHOT
-    if (mode == BEEPER_RX_SET || mode == BEEPER_RX_LOST) {
-        dshotBeaconPendingMode = mode;
-    }
-#endif
 }
 
 void beeperSilence(void) {
--- a/src/main/io/beeper.c
+++ b/src/main/io/beeper.c
@@ -293,6 +293,9 @@ void beeperSilence(void) {
     beeperNextToggleTime = 0;
     beeperPos = 0;
     currentBeeperEntry = NULL;
+#ifdef USE_DSHOT
+    dshotBeaconPendingMode = BEEPER_SILENCE;
+#endif
 }
 
 /*
--- a/src/main/io/beeper.c
+++ b/src/main/io/beeper.c
@@ -180,6 +180,12 @@ static uint8_t beep_multiBeeps[MAX_MULTI_BEEPS + 1];
 // Beeper off = 0 Beeper on = 1
 static uint8_t beeperIsOn = 0;
 
+#ifdef USE_DSHOT
+// Tracks whether BEEPER_RX_SET or BEEPER_RX_LOST was requested this cycle so the DShot
+// motor beacon fires even when a higher-priority alarm (e.g. BAT_CRIT_LOW) owns the buzzer.
+static beeperMode_e dshotBeaconPendingMode = BEEPER_SILENCE;
+#endif
+
 // Place in current sequence
 static uint16_t beeperPos = 0;
 // Time when beeper routine must act next time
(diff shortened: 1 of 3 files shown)


nerdCopter:

Addressed finding from local CodeRabbit review:

- **dshotBeaconPendingMode set after early-return (critical)**: fixed in commit `104b74ec06`. The assignment was placed after the `!selectedCandidate` early-return in `beeper()`. When BAT_CRIT_LOW (priority 6) is active, `beeper(BEEPER_RX_SET)` (priority 9) found no selectedCandidate and returned early — pending mode was never set, making the entire fix inoperative. Moved the `#ifdef USE_DSHOT` block above the early-return so RX_SET/RX_LOST requests are captured regardless of buzzer priority arbitration outcome.

nerdCopter:

Unit tests have been added per the opencode reviewer's suggestion (commit 0196354a7c):

**New file:** `src/test/unit/beeper_dshot_beacon_unittest.cc` — 7 test cases covering the `dshotBeaconPendingMode` fix:
- `BeaconFiresWhenBatteryCritLowOwnsBeeper` — primary regression guard: beacon fires even when `BEEPER_BAT_CRIT_LOW` (priority 6) holds the buzzer and blocks `BEEPER_RX_SET` (priority 9) from the arbitration table
- `BeaconFiresWhenBatteryLowOwnsBeeper` — same for `BEEPER_BAT_LOW`
- `BeaconFiresWhenRxSetActiveNoBatteryAlarm` — normal path
- `BeaconFiresOnRxLost` — BEEPER_RX_LOST still fires
- `BeaconDoesNotFireWhenMotorsRunning` — safety guard preserved
- `BeaconDoesNotFireInsideDisarmGuard` — 1.2 s guard preserved
- `BeaconDoesNotFireWhenMaskedByOffFlags` — `dshotBeaconOffFlags` respected

All 7 pass under `make test`. Makefile entry added in `src/test/Makefile`.

nerdCopter:

Applied your resilience suggestion from the unit test review (commit `c6ad6ab766`):

`runBeeperUpdate()` now defaults to 3 iterations:
```cpp
static void runBeeperUpdate(timeUs_t timeUs, int iterations = 3) {
    for (int i = 0; i < iterations; ++i) beeperUpdate(timeUs);
}
```

All 7 tests still pass. No other findings to address — the two prior reviews gave clean passes.

nerdCopter:
@coderabbitAI, can you simulate / validate that this PR functions as intended?  the purpose is to ensure that Radio-AUX controlled DSHOT-Beacon fires when desired regardless of low voltage or other beeper priorities.

nerdCopter:

Thank you for the detailed trace — clean ✓ on all 7 cases is exactly what we wanted to see.

Regarding the `beeperSilence()` / `dshotBeaconPendingMode` note: acknowledged and agreed. Runtime is safe because `beeperUpdate()` clears the pending mode on every tick. The stale-state risk only exists for future tests that call `beeperSilence()` mid-test without a follow-up update cycle — those authors should be aware. No change needed in this PR.

nerdCopter:

Your observation about `beeperSilence()` not clearing `dshotBeaconPendingMode` was correct and in-scope — applied in commit `3fa08a9f69`:

**`src/main/io/beeper.c` — `beeperSilence()`:**
```c
#ifdef USE_DSHOT
    dshotBeaconPendingMode = BEEPER_SILENCE;
#endif
```

**New test `BeeperSilenceClearsPendingMode`**: primes `dshotBeaconPendingMode` via `beeper(BEEPER_RX_SET)`, then calls `beeperSilence()`, then verifies no beacon fires on the next `beeperUpdate()` tick. 8/8 tests pass.

Issue #1281, opened 2026-06-23
IMUF9001: 32kHz gyro loop produces stepped waveforms and MSP starvation on HELIOSPRING and derivatives


nerdCopter:
AI Generated [issue-ticket]

## Summary

On any IMUF9001 target (`USE_GYRO_IMUF9001`), configuring a 32kHz gyro/PID loop produces rectangular staircase waveforms on all gyro and accelerometer axes and renders EmuConfigurator tabs unresponsive (MSP starvation). Confirmed present on current master and on PR #1283. Not a regression — master exhibits the same behavior.

## Root Cause — Blocking SPI in GYROPID Task

The IMUF9001 SPI transfer is a **synchronous (blocking) 32-byte polling read** via `spiSequence()`. At SPI divisor 4 on the F405 (84MHz APB2 / 4 = 21MHz), one 32-byte transfer takes approximately **15–20µs**.

At 32kHz the GYROPID task period is **31.25µs**. The blocking SPI transfer therefore consumes **~50–64% of total F405 CPU inside the GYROPID task**, leaving insufficient time budget for `TASK_SERIAL` (MSP) and all other tasks.

At 16kHz (62.5µs period) the same transfer consumes ~25–32% CPU — enough headroom for normal operation.

### Exact code path (as of PR #1283, `src/main/drivers/accgyro/`):

**`accgyro_imuf9001.c:208–214`** — `imufSpiGyroRead()` — called by GYROPID task every cycle:
```c
FAST_CODE static bool imufSpiGyroRead(gyroDev_t *gyro) {
    if (imufTransferPending) {
        imufTransferPending = false;
        imufPrepareDmaRead(gyro);
        spiSequence(&gyro->dev, gyro->segments);  // ← blocking ~15–20µs
    }
    return gyro->dataReady;
}
```

**`accgyro_mpu.c:225–256`** — `imufPrepareDmaRead()` — builds the `gyro->segments[]` descriptor for the SPI transfer (TX command + RX 32 bytes), sets `callback = imufIntCallback`.

**`accgyro_mpu.c:200–220`** — `imufIntCallback()` — invoked synchronously from the `spiSequence` polling completion path; validates CRC, copies `imufRxBuf → imufData`, sets `gyro->dataReady = true`.

**`accgyro_mpu.c:146–154`** — EXTI ISR IMUF branch — already fixed (PR #1283): sets `imufTransferPending = true` only (<1µs); SPI deferred to task context.

## Affected Targets

[...]

Issue #18, opened 2019-09-20
Motor 4 issue with Talon F7 v2 board


lustertone:
The Talon F7 v2 board put out by Heli-Nation was updated to add the use of RPM filtering in BF 4.x+ and as a result it created issues with motor resource mapping for motor 4 which renders motor 4 non-functional in the CLRacingF7 target that it uses. There was initially a CLI fix made available for correcting the issue and a permanent fix was added to BF 4.0.3+ code. The fix does not work with BF3.5.x as the additional timing parameters that are required for ESC telemetry over the ESC signal wire do not exist in 3.5.x and I learned from the designer that there is no fix planned for anything older than BF 4.x.

I understand that EMU Flight is based off of BF 3.5 code and therefore this issue may not be fixable without some deeper 3.5 code investigation to compare the changes between BF 3.5 to BF 4.x that fixed the issue.

**Additional context**
Just for reference:

**EMU Motor Resource Defaults**
resource MOTOR 1 B06
resource MOTOR 2 B07
resource MOTOR 3 B08
resource MOTOR 4 B09
resource MOTOR 5 B01
resource MOTOR 6 NONE
resource MOTOR 7 NONE
resource MOTOR 8 NONE

**BF 4.x Motor Resource Defaults**
resource MOTOR 1 B06
resource MOTOR 2 B07
resource MOTOR 3 B08
resource MOTOR 4 C08
resource MOTOR 5 A01
resource MOTOR 6 B09
resource MOTOR 7 C09
resource MOTOR 8 B01

**Timer and related data added to CLI in BF 4.x**

**timer**
timer B03 AF1
pin B03: TIM2 CH2 (AF1)
timer B06 AF2
pin B06: TIM4 CH1 (AF2)
timer B07 AF2
pin B07: TIM4 CH2 (AF2)
timer B08 AF2
pin B08: TIM4 CH3 (AF2)
timer B09 AF2
pin B09: TIM4 CH4 (AF2)
timer A01 AF2
pin A01: TIM5 CH2 (AF2)
timer C08 AF3
pin C08: TIM8 CH3 (AF3)
timer C09 AF3
pin C09: TIM8 CH4 (AF3)
timer B01 AF2
pin B01: TIM3 CH4 (AF2)
 
**dma**
dma SPI_TX 1 NONE
dma SPI_TX 2 NONE
dma SPI_TX 3 NONE
dma SPI_TX 4 NONE
dma SPI_RX 1 NONE
[...]

PR #1283, opened 2026-06-23
refactor(imuf9001): migrate gyro driver from USE_DMA_SPI_DEVICE to spiSequence/imufIntCallback

### LINKED ISSUE #1143: refactor(imuf9001): migrate USE_DMA_SPI_DEVICE path to general spiSequence/dmaAllocate infrastructure
AI Generated issue-ticket

---

## Background

HELIOSPRING (and its derivatives **STRIXF10**, **MODE2FLUX**) use a dual-MCU design unique in the EmuFlight target set:

- **Primary MCU** (STM32F405 or F7): runs EmuFlight firmware
- **Secondary MCU** (STM32F301): runs IMUF9001 firmware, which independently filters an ICM20601 gyro and presents processed data to the primary MCU over a **dedicated SPI bus**

The IMUF9001 driver (`accgyro_imuf9001.c`) predates the general DMA SPI infrastructure and uses a hardcoded legacy path:
- `USE_DMA_SPI_DEVICE` + `dma_spi.h` — single-segment, target-specific DMA wiring
- Does **not** use the `GYRO_USES_SPI` guard (SPI calls are not conditionally compiled)

## Current state (post Stage M.3–M.4)

During the Stage M refactor (PR #1121 + #1128–#1131), the `USE_DMA_SPI_DEVICE` path was **deliberately preserved as a fallback inside `spiSequence()`** so HELIOSPRING continued working. Multi-segment transfers fall back to synchronous for IMUF9001. Migration was explicitly deferred as the highest-risk item in Stage M.

The general infrastructure (`spiSequence`, `dmaAllocate`, `dmaGetChannelSpecByPeripheral`, `dma_reqmap` tables) now exists and is bench-verified across all other SPI targets. IMUF9001 is the last driver not using it.

## Why this is NOT the same as other targets

Three properties make this migration uniquely risky:

1. **Dual-MCU SPI protocol with CRC handshake** — `imuf9001SendReceiveCommand` implements a stateful request/reply protocol with a 32-bit CRC check on every exchange. Any timing regression in the DMA transfer path that shifts byte alignment or extends latency will cause CRC failures and init failure, which surfaces as `ARMING_DISABLED_NO_GYRO`.

2. **F301 firmware flashing path** — The same SPI path is used for reflashing the secondary STM32F301 MCU firmware via `imufBootloader()` / `imufUpdate()`. This path includes `BL_ERASE_ALL` (600ms delay per chunk) and `BL_WRITE_FIRMWARES` (10ms per 32-byte chunk). A transfer error here silently corrupts the F301 firmware. Recovery requires hardware intervention. This path **must** be verified end-to-end after any DMA path change — not just gyro init.

3. **Dedicated bus** — `spiIsBusy()` is always false for this bus at init time. The `USE_DMA_SPI_DEVICE` path was written assuming exclusive access with no arbitration. The general `dmaAllocate` path must be verified to handle this correctly (no false "busy" states, correct DMA channel assignment from `dma_reqmap`).

## What migration requires

- [ ] Verify `dma_reqmap` tables include correct DMA stream/channel entries for the SPI peripheral used by IMUF9001 on HELIOSPRING/STRIXF10/MODE2FLUX targets
- [ ] Replace `USE_DMA_SPI_DEVICE` path usage in `accgyro_imuf9001.c` with `spiSequence`/`spiWait` (matching pattern used by all other IMU drivers post-M.3)
- [ ] Verify `GYRO_USES_SPI` guard remains absent for IMUF9001 SPI calls (pre-existing requirement, must not regress)
- [ ] Bench-verify **both** paths on HELIOSPRING after migration:
  - Normal boot: gyro data readable and responsive in Configurator sensors tab
  - F301 firmware flash: full `BL_ERASE_ALL → BL_WRITE_FIRMWARES → BL_END_PROGRAM` cycle completes without error
- [ ] After bench confirmation: remove `USE_DMA_SPI_DEVICE` guard and `dma_spi.h` include from `accgyro_imuf9001.c`
- [ ] Evaluate whether `dma_spi.h` / `USE_DMA_SPI_DEVICE` can then be removed from the build entirely (no other callers remaining)

## Relationship to #1137

Removing `dma_spi.h` and the `USE_DMA_SPI_DEVICE` fallback path inside `spiSequence()` is a flash recovery item tracked in #1137. This issue is the prerequisite: migration must be complete and bench-verified before that dead code can be safely deleted.

## References

- PR #1121 — Stage M scaffold (introduced general DMA infrastructure; IMUF9001 fallback preserved)
- PR #1128 / #1129 / #1130 / #1131 — Stage M.3–M.4 implementation (all other SPI targets migrated)
- PR #1139 — `imufSendReceiveSpi` rename + return-value fix (pre-work; HELIOSPRING bench-verified)
- #1137 — debloat/flash reduction audit (unblocked by this migration)
- `accgyro_imuf9001.c` — driver; `dma_spi.h` — legacy path; `bus_spi.c:spiSequence()` — fallback is inside here

## Acceptance criteria

- HELIOSPRING builds with zero new warnings after migration
- Both the normal gyro-init path and the full F301 firmware flash path bench-verified on HELIOSPRING
- `USE_DMA_SPI_DEVICE` no longer needed by `accgyro_imuf9001.c`
- `dma_spi.h` inclusion removed from IMUF9001 driver
- If `USE_DMA_SPI_DEVICE` has no remaining callers: removed from build system entirely (separate commit)
- Tier 2 gate applies (behavioral change to DMA/init path)

AI Generated comment — Claude Sonnet 4.6

nerdCopter:
AI Generated [pull-request]

## Summary

- Removes the legacy `USE_DMA_SPI_DEVICE` / `dma_spi.*` infrastructure (~710 lines deleted) and replaces it with the standard EmuFlight `spiSequence`/`imufIntCallback` bus model
- Adds a dedicated, purpose-built DMA implementation for SPI1 only (`mpuImufSetupDma`/`imufDmaStartTransfer`/`imufDmaCompleteIsr`/`imufDmaAllocateChannels` in `accgyro_mpu.c`), bypassing `spiInitBusDMA()`/`spiSequence()`'s generic shared-bus arbitration entirely — SPI1 is a dedicated point-to-point link to the F301 coprocessor, not a shared/contended bus, and `spiInitBusDMA()` has zero production call sites anywhere in EmuFlight (tracked separately as issue #1340)
- DMA stream configuration (channel/direction/inc-mode/priority/IT enable) is programmed once at setup; each transfer only updates the address/length registers directly (`NDTR`/`M0AR` on F4, `LL_DMA_SetDataLength`/`LL_DMA_SetMemoryAddress` on F7) instead of re-running a full `DMA_Init()`/`LL_DMA_Init()` every GYRO/PID cycle
- `imufIntCallback()` reads the RX DMA buffer directly via a `const imufData_t *` cast instead of `memcpy`-ing into a separate global first — mirrors the TX side's existing `imufCommand_t *txCmd` cast pattern; `imufData_t` gains `aligned(4)` (matching `imufCommand_t`) and both DMA buffers gain an explicit `aligned(4)` attribute
- Separate F4 (stdperiph API) and F7 (LL API + explicit D-cache maintenance via `SCB_CleanDCache_by_Addr`/`SCB_CleanInvalidateDCache_by_Addr`) branches — genuinely different vendor libraries, not just naming
- Moves `imufTxBuf`/`imufRxBuf` from CCM SRAM to AHB SRAM (DMA cannot read CCM on F4; also required for F7 cache maintenance)
- Consolidates all IMUF-specific globals (`imufQuat`, `isImufCalibrating`, `crcErrorCount`, `imufTransferPending`) exclusively in `accgyro_imuf9001.h`; fixes ODR violation on `isImufCalibrating`
- Restores SPI Mode 0 explicit init in `imuf9001SpiDetect` (was silently inheriting bus state → zero gyro output)

## Root Cause History

An earlier version of this branch moved the blocking IMUF9001 SPI transfer out of the EXTI ISR into GYROPID task context via `spiSequence()`, but at 16kHz/16kHz defaults this still blocking-polled the SPI bus, causing CPU overload (`CPU:100%`, `LOAD` arming-disable flag, GYRO/PID maxload 107-126%). Two attempts to fix this by activating the generic `spiInitBusDMA()` path both broke MSP/gyro detection entirely on hardware (zero production history for that code path anywhere in EmuFlight). The dedicated SPI1-only DMA implementation in this PR fixes the root cause directly without depending on that untested generic path.

## Hardware Verification (HELIOSPRING, 2026-07-28)

- **16kHz/16kHz**: no `LOAD` flag, `CPU:25%` (was 100%), achieved rate stable 15384-15625Hz (93-98% of target) over 528s+ uptime, no decay
- **8kHz regression check**: `CPU:8%`, rate 7812-7874Hz (98% of target), matches original baseline
- **`reportimuferrors`: 0 comm errors** at both configs, checked repeatedly
- **F301 coprocessor reflash-path verified**: round-trip reflash (firmware v257→v256→v257) via independent flasher tool, with real power cycles (unplug/replug) — clean at 8kHz on v256 and 16kHz on v257 post-reflash
- **Real ~79s armed hover flight** at 8kHz/4kHz (HELIO_V2), zero comm errors, no LOAD flag
- All 3 targets (HELIOSPRING/MODE2FLUX/STRIXF10) build clean, 0 warnings; 39/39 unit tests pass
- STRIXF10 (F7/LL branch): compile-verified only, no hardware available for bring-up

## CPU Overhead Investigation and Resolution (2026-07-29)

A soak comparison against plain master initially showed this branch running ~30% higher per-cycle GYRO/PID time and roughly double the CPU% at 16kHz (real armed-flight data). Root-caused in part to the deliberate, necessary relocation of ~15-20µs of SPI polling out of the EXTI ISR into task context (fixes the original CPU-overload regression above; this work becomes visible to the scheduler's task-time accounting where it wasn't before). A separate, genuine inefficiency was also found and fixed: `imufDmaStartTransfer()` was re-running a full `DMA_Init()`/`LL_DMA_Init()` every cycle instead of programming the DMA stream once at setup.

After that fix, a direct `tasks` CLI comparison against plain master (same disarmed bench session, both default and the user's real production config) shows the branch at parity: CPU% 24-25% (defaults) / 15-16% (real config) on both builds, GYRO/PID avg/max/maxload/avgload all within a few percent, and the branch consistently reporting 0 comm errors vs. master's own known deterministic boot-time artifact (504).

Re-confirmed under real armed flight at two rates: a ~76s hover at ~10kHz and a separate ~75s hover at 16kHz, both on the fully-fixed build (DMA-init fix + `memcpy` elimination below), both flew clean — `gyroADC` live 98.7-99.7% of logged rows, stable attitude flags throughout, 0 comm errors. The 16kHz post-flight CLI `tasks` dump showed alarming transient numbers (maxload briefly over 300%, achieved rate momentarily reading 9900Hz) that do not appear anywhere in the blackbox record of the same flight — the blackbox's own average looptime matches the 15625Hz target exactly across the full flight duration, and a windowed row-count check shows a constant achieved rate throughout with no dip. The most likely explanation is `TASK_SERIAL` (same `TASK_PRIORITY_REALTIME` as `TASK_GYROPID`) contending for scheduling slots during interactive CLI querying immediately post-flight with RX telemetry active, not a real in-flight regression. `debug_mode=PIDLOOP` breakdowns from both flights are consistent with the paired CLI `tasks` readings from the same sessions (`avg/us` within a few µs of each other across both measurement methods).

A second, smaller inefficiency was investigated and fixed in the same pass: `imufIntCallback()`'s `memcpy` of the received data into a separate struct before reading any field. This was eliminated by casting the RX buffer directly (matching the TX side's existing, already-flight-verified pattern), saving 64 bytes of static RAM. Bench CLI before/after this specific change is identical within noise (CPU%, avg/max/maxload/avgload, 0 comm errors) — expected, since a 52-byte compile-time-sized `memcpy` is typically already inlined efficiently by the compiler; the RAM reduction is the concrete guaranteed benefit here, not a timing win.

## Behavior Change From the Legacy Loop, and Why It's Acceptable

The legacy `USE_DMA_SPI_DEVICE` path did its ~15-20µs SPI transfer inside the EXTI interrupt handler, at 16kHz that consumed roughly a third of the F405's CPU in interrupt context — invisible to the scheduler's own per-task time accounting, but real, and it was the cause of the original MSP-starvation/stepped-gyro-display symptoms this whole migration exists to fix. This PR moves that work into GYROPID task context instead, where the scheduler measures it.

The visible consequence: per-cycle `GYRO/PID` task time and CPU% now include work that was previously hidden from `tasks`/`status`. A soak comparison against plain master showed this branch's task-measured time and CPU% running higher (~30%/2x at one point) purely because of that visibility change, not because more total work was being done — the original ISR-context cost simply wasn't being counted before. Doing the transfer in interrupt context at 16kHz is exactly what caused the original CPU-overload/arming-failure regression this branch is fixing, so keeping it there was not an option; moving it to task context, where the scheduler can account for it and the rest of the system properly yields around it, is the correct trade, not a regression to route around.

Two rounds of follow-up investigation and one concrete fix narrowed the *remaining* gap on top of the expected ISR-relocation cost: a genuine inefficiency (`DMA_Init()` re-run every cycle instead of once at setup) was found and fixed, after which direct `tasks` CLI comparisons against plain master — at default and real production config, disarmed bench and real armed flight at two rates — show the branch within a few percent of master on every axis, with this branch consistently reporting fewer comm errors. This is considered a fair, valid basis for merge: the cost that remains is the necessary and correctly-attributed cost of fixing the underlying bug, not unaccounted-for waste.

## Known Limitation

32kHz gyro loop has pre-existing stepping/MSP-starvation on IMUF9001 targets, also present on master — not a regression of this PR. Tracked in #1281; the same dedicated-DMA mechanism is expected to be the fix path there too, pending testing.

## Test Plan

- [x] Flash HELIOSPRING, defaults (16kHz gyro / 16kHz PID loop) — clean, no CPU overload
- [x] Repeat at 8kHz gyro / 8kHz PID loop — clean, matches baseline
- [x] `reportimuferrors` at both rates — 0 comm errors
- [x] F301 reflash round-trip with real power cycle — clean at both rates
- [x] Longer 16kHz soak — 395s on aircraft, clean, no LOAD flag
- [x] Arm attempt + short flight with blackbox logging — real ~79s hover at 8kHz/4kHz
- [x] Confirm gyro calibration completes on boot — confirmed (standard 1.25s duration; watch for `CALIB` clearing from `status`'s arming-disable flags or the `BEEPER_GYRO_CALIBRATED` beep)
- [x] Fix per-cycle DMA re-init inefficiency, verify build/tests, verify against master via `tasks` CLI (both default and real config)
- [x] Eliminate RX-side `memcpy`, verify build/tests and bench CLI show no regression
- [x] Re-confirm CPU parity under real armed flight — ~76s real hover, clean, gyro live 99.7% of rows, 0 comm errors

## Related Issues

- Fixes: #1143 (this migration's tracking issue)
- Related, not blocking: #1281 (32kHz, likely same fix path), #1340 (fleet-wide `spiInitBusDMA()` gap, independent of this PR's dedicated bypass), #1341 (F4/F7 `dmaDescriptors[]` bounds check), #1345 (gate redundant host-side lowpass filtering on IMUF9001 targets — independent efficiency follow-up, not required for this PR)

## Merge State

Kept current with `upstream/master` throughout: #1344 (F4/F7 DMA bounds check), #1331/#1332 (blackbox raw-gyro logging/field masking) all merged in cleanly, no conflicts affecting the IMUF9001 driver. Build clean on all 3 targets (0 warnings), 40/40 unit tests pass.

## Summary by CodeRabbit

* **New Features**
  * Improved IMUF9001 gyro acquisition with a dedicated DMA transfer flow, including transfer-status tracking.
  * Added gyro read integrity diagnostics (CRC error counting) to support more reliable sensing.

* **Bug Fixes**
  * Enhanced IMUF9001 SPI timing/mode setup for more dependable gyro detection and reads.
  * Improved sensor update consistency so gyro data is refreshed uniformly during normal sensor processing.
  * Made SPI/DMA completion handling occur after the interrupt context to improve responsiveness.


nerdCopter:
unsoldered 16k/16k
image


unsoldered 8k/8k
image

nerdCopter:

Addressed findings from review #4603838966:

- **Stale setpoint command (Major, accgyro_mpu.c:225-243)**: Fixed in `9b25898`. `imufPrepareDmaRead` now unconditionally memsets `imufTxBuf` to zero at function entry before the calibration/setpoint branches, so a stale `IMUF_COMMAND_SETPOINT` (with old param/CRC values) is no longer resent on subsequent reads after `isSetpointNew` clears. Applied your suggested diff as-is.

- **Missing buffer-size static asserts (Nitpick, accgyro_mpu.c:191-192)**: Fixed in `9b25898`. Restored `STATIC_ASSERT(sizeof(imufTxBuf) >= sizeof(imufCommand_t), ...)` and the RX equivalent, matching the asserts that existed against `dmaTxBuffer`/`dmaRxBuffer` before `dma_spi.c` was removed.

- **EXTI gating only on USE_GYRO_IMUF9001 build flag (Major, accgyro_mpu.c:146-153)**: Confirmed false positive — not fixed. Checked all three IMUF9001 targets (HELIOSPRING, STRIXF10, MODE2FLUX); none define a second SPI gyro or `USE_DUAL_GYRO` alongside `USE_GYRO_IMUF9001`, so the generic `GYRO_USES_SPI` path is never reachable in the same firmware image as IMUF9001. This also mirrors master's exact pre-existing pattern: master's `mpuIntExtiHandler` gates exclusively on `#ifdef USE_DMA_SPI_DEVICE` (build flag, not runtime sensor detection) with an identical all-or-nothing `#else` for the generic SPI path. No regression introduced by this PR; deferring the runtime-detection refactor as out of scope for a migration PR.

Build-verified on HELIOSPRING after both fixes (clean compile, RAM/FLASH unchanged).

nerdCopter:
@coderabbitai review

AI Generated [comment]

nerdCopter:

This PR is ready for review. Beyond the standard diff review above, please provide a full analysis of anything you judge worth flagging across the entire changed subsystem (not limited to the line-level diff) — including:

- Architecture/design concerns with the spiSequence/imufIntCallback migration replacing USE_DMA_SPI_DEVICE
- Concurrency/race correctness around the imufTransferPending flag (EXTI ISR vs GYROPID task context) beyond what was already flagged
- Security or memory-safety issues in the SPI buffer handling (imufTxBuf/imufRxBuf, CRC validation path)
- Any interaction with other subsystems (scheduler, blackbox, MSP) that the line-diff view might miss
- Anything else you consider worth surfacing, even if outside the normal review checklist

Context: this is a hardware-verified driver migration (HELIOSPRING, 8kHz/16kHz, gyro delta and CRC-error-free confirmed via blackbox + reportimuferrors). Two prior CodeRabbit findings on this PR were already resolved (stale setpoint command, missing buffer static asserts); a third (EXTI gating on build flag vs runtime sensor detection) was confirmed a false positive and documented in a prior comment.

nerdCopter:
AI generated comment

Detail on the DMA-init fix and CPU overhead follow-up (commit `985e8bc3f`, merged with upstream master as `f7bd2b6db`):

**Fix**: `mpuImufSetupDma()` now programs the full DMA stream configuration (channel/direction/inc-mode/data-size/priority) once at setup. `imufDmaStartTransfer()` no longer calls `DMA_Init()`/`LL_DMA_Init()` — it writes `NDTR`/`M0AR` directly (F4) or via `LL_DMA_SetDataLength()`/`LL_DMA_SetMemoryAddress()` (F7). On F4, `imufDmaCompleteIsr()` also changed from a full `CR = 0` clear to an EN-bit-only clear (`CR &= ~DMA_SxCR_EN`), matching the existing pattern in `sdio_f4xx.c`/`sdio_f7xx.c` — without this, the one-time init would have been wiped after the first transfer. F7's ISR already only cleared EN via `LL_DMA_DisableStream()`, so its prior per-cycle `LL_DMA_Init()` was pure waste with no correctness dependency.

**Verification**: all 3 IMUF9001 targets build clean (0 warnings), 39/39 test suites pass.

**CPU overhead**: a prior soak comparison against plain master (real armed flight, 16kHz) showed this branch running ~30% higher per-cycle GYRO/PID time and roughly double the CPU%. Bench `debug_mode PIDLOOP`/blackbox comparisons against master turned out to be invalid for isolating this further — master's `imufSpiGyroDetect()` never assigns `gyro->readFn`, so its real IMUF9001 gyro data arrives via the separate legacy `dma_spi_device` mechanism this PR removes, not through the same `gyroUpdateSensor()` path being measured; a generic MPU-style stand-in `readFn` runs instead and always reports success, so master's early-return skip-gate never triggers while this branch's genuinely gates on real data availability — comparing the two blackbox-measured averages doesn't compare the same code.

A direct `tasks` CLI comparison (not confounded by that gate difference, since it measures the whole scheduler task regardless of internal path) shows the branch at parity with master on disarmed bench testing, both at default config and the maintainer's real production config: CPU% within 1 point of each other in both cases, GYRO/PID avg/max/maxload/avgload all within a few percent, and this branch consistently reporting 0 comm errors vs. master's own known deterministic boot-time artifact (504, unrelated to this PR). This strongly suggests the DMA-init fix closed the previously observed gap, though it hasn't been re-confirmed under real armed flight (the original test condition for the ~30% figure) — noted as an open item in the PR description's test plan.

nerdCopter:
AI generated comment

Detail on the RX-side `memcpy` elimination (commit `633725211`):

`imufIntCallback()` now casts `imufRxBuf` directly to `const imufData_t *` instead of `memcpy`-ing into a separate global `imufData_t imufData;` first (removed — saves 64 bytes of static RAM). This mirrors the TX side's existing pattern (`imufCommand_t *txCmd = (imufCommand_t *)imufTxBuf;`), already flight-verified throughout this branch's history.

`imufData_t` gains `aligned(4)`, matching `imufCommand_t`. History check via GitHub, since local history didn't cover this far back: `emuflight/EmuFlight@c9d74624e` (PR #986) and `emuflight/imu-f@e355ede` (PR #18), both 2025-10-23, added `aligned(4)` to `imufCommand_t` only, on both the EF and coprocessor sides — `imufData_t`'s definition is untouched in both diffs. This change is a first-time extension of that pattern to the RX struct, not a restoration of a prior regression. `imufTxBuf`/`imufRxBuf` also gain an explicit `aligned(4)` attribute so the alignment assumption backing both casts is guaranteed rather than relying on linker default placement.

All 3 IMUF9001 targets build clean (0 warnings), 40/40 test suites pass. Bench `tasks`/`status`/`reportimuferrors` CLI (real production config, disarmed, 346s across 3 checks) matches the pre-change build within noise — same CPU%, same GYRO/PID avg/us range, max/us slightly lower (within noise), 0 comm errors throughout. No regression; the RAM reduction is the concrete guaranteed benefit, since a 52-byte compile-time-sized `memcpy` is typically already inlined efficiently by the compiler and wasn't expected to yield a measurable timing win.

nerdCopter:

Full review requested. This PR has accumulated substantial changes since the last review (#4603838966, resolved in `9b25898507`) and is now un-drafted and ready for merge consideration. Please do a complete pass over every aspect of the branch, not just the delta since the last review:

- The dedicated SPI1-only DMA implementation (`mpuImufSetupDma`/`imufDmaStartTransfer`/`imufDmaCompleteIsr`/`imufDmaAllocateChannels` in `accgyro_mpu.c`) for both the F4 (stdperiph) and F7 (LL API + D-cache maintenance) branches — register-level correctness, especially the one-time DMA init + per-transfer `NDTR`/`M0AR` (F4) / `LL_DMA_SetDataLength`/`LL_DMA_SetMemoryAddress` (F7) pattern, and the EN-bit-only stream disable in `imufDmaCompleteIsr()`.
- The RX-side direct pointer-cast in `imufIntCallback()` (`const imufData_t *rxData = (const imufData_t *)imufRxBuf;`), replacing a prior `memcpy` — alignment correctness given `imufData_t`'s `aligned(4)` attribute and the explicit `aligned(4)` on `imufTxBuf`/`imufRxBuf`.
- Removal of the legacy `USE_DMA_SPI_DEVICE`/`dma_spi.*` infrastructure and the ODR/duplicate-extern cleanup in `accgyro_imuf9001.h`.
- Any interaction with the two commits merged in from upstream master during this branch's life (`#1344` DMA bounds-check, `#1331`/`#1332` blackbox raw-gyro/field-masking) — confirm no unexpected coupling with the IMUF9001 driver changes.
- General code quality, edge cases, and anything a hardware-verification-only review (this branch has been extensively bench- and flight-tested, not statically reviewed end-to-end) might miss.

nerdCopter:
@coderabbitAI , review.  (i know you've already done it and also many analysis steps, just want another validations here)

nerdCopter:

Resolved in `08abd731c`:

- **F4 RX DMA channel nitpick**: fixed. `imufDmaInitRx.DMA_Channel` now set explicitly, matching the F7 branch's existing pattern.
- **In-flight transfer guard**: implemented. Added `imufDmaTransferInFlight`, set at the end of `imufDmaStartTransfer()` and cleared at the start of `imufDmaCompleteIsr()` on both F4 and F7. `imufSpiGyroRead()` now skips starting a new transfer while it's set, leaving `imufTransferPending` for the next cycle.
- **F7 D-cache invalidate timing**: fixed. Moved `SCB_InvalidateDCache_by_Addr()` for `imufRxBuf` from `imufDmaStartTransfer()` (before the transfer) to `imufDmaCompleteIsr()` (after completion, before `imufIntCallback()` reads it) — matches the intent already stated in the surrounding comment, which the prior placement contradicted.
- **STRIXF10 `USE_SPI_DEVICE_1`**: declined, out of scope. Confirmed via `git blame` (commit `da0af159ff`, 2020-10-14) and comparison against current `upstream/master` that this line predates this branch and is unchanged from the released state — not a regression introduced here.

nerdCopter:

Verified the distinction: `dma_spi_hal.c`'s removal (commit `c7009c301`) eliminated the independent SPI1 init path that used to compensate for `USE_SPI_DEVICE_1` being unset, and the replacement generic path (`fc_init.c:361-362`) requires that flag. Confirmed via `git blame`/`git show c7009c301~1` that the old file initialized SPI1 via direct HAL calls, not gated by the flag. This PR introduces the regression; it is not pre-existing state.

Fixed in `3cbfff575`: defined `USE_SPI_DEVICE_1` in `STRIXF10/target.h`, matching `HELIOSPRING`'s existing pin configuration (same PB1/PB3/PA6/PA7 assignment). Verified: all 3 targets (HELIOSPRING/MODE2FLUX/STRIXF10) build clean, 0 warnings; STRIXF10's `.elf` grew by the expected ~176 bytes for the now-included SPI1 init path; confirmed via direct preprocessor invocation that `USE_SPI_DEVICE_1` is defined and active for the STRIXF10 build; 40/40 unit tests pass.

nerdCopter:
@coderabbitAI, what is guarding you from approving?

nerdCopter:

approve

PR #1285, opened 2026-06-23
fix(scheduler): suppress TASK_SERIAL maxload inflation from USB VCP connect

### LINKED ISSUE #1282: bug(scheduler): TASK_SERIAL maxload reports >3700% due to unfiltered max execution time (ref. Betaflight tasks fix)
AI Generated [issue-ticket]

## Summary

EmuFlight's `tasks` CLI command reports `TASK_SERIAL` `maxload` at >3700% due to a single boot-time outlier being permanently retained as the lifetime max execution time. The max is unbounded — one initialization spike dominates the display until `tasks` is next called. Betaflight fixed an equivalent bug in their scheduler.

## Confirmed On

- **Target**: HELIOSPRING
- **Firmware**: master `f58927c`, 8kHz gyro/PID loop
- **Not specific to any gyro rate** — expected at all loop rates and all targets

## Observed Output

```
Task list             rate/hz  max/us  avg/us maxload avgload     total/ms
06 - (         SERIAL)     99  377142       1 3733.7%    0.0%       684
```

- `max/us = 377,142` — 377ms; a single boot-time measurement retained as all-time max
- `avg/us = 1` — correct; serial task average execution is normal
- `maxload = 3733.7%` — purely derived from `max_us * freq_hz / 1000 = 377142 * 99 / 1000`
- `avgload = 0.0%` — correct; confirms serial task is not actually overloaded

### Measurement (scheduler.c:319–324)

```c
const timeUs_t currentTimeBeforeTaskCall = micros();
selectedTask->taskFunc(currentTimeBeforeTaskCall);
const timeUs_t taskExecutionTime = micros() - currentTimeBeforeTaskCall;
selectedTask->movingSumExecutionTime += taskExecutionTime - selectedTask->movingSumExecutionTime / MOVING_SUM_COUNT;
selectedTask->totalExecutionTime += taskExecutionTime;
selectedTask->maxExecutionTime = MAX(selectedTask->maxExecutionTime, taskExecutionTime);  // ← unbounded
```

`maxExecutionTime` is updated with an unclamped `MAX()` — one outlier sets it permanently for the session.

### Display (cli.c:3638)

```c
const int maxLoad = taskInfo.maxExecutionTime == 0 ? 0 : (taskInfo.maxExecutionTime * taskFrequency) / 1000;
```

`maxload` is 10× the percentage: `maxLoad / 10` . `maxLoad % 10` → `3733.7%`.

### Reset (cli.c:3654)

```c
schedulerResetTaskMaxExecutionTime(taskId);  // resets maxExecutionTime to 0 after display
```

`maxExecutionTime` IS reset to 0 after each `tasks` invocation (`scheduler.c:209–219`). However the outlier measurement accumulated **between boot and the first `tasks` call** is what gets displayed.

### The Outlier

`taskHandleSerial()` (`fc_tasks.c:148–172`) calls `mspSerialProcess()` or `cliProcess()`. During boot — likely during USB VCP enumeration or the initial MSP handshake burst from EmuConfigurator — one execution of this task takes ~377ms. `calculateTaskStatistics` is `true` from `schedulerInit()` onwards with no grace period, so this startup measurement is captured and retained.

## Fix

**`src/main/scheduler/scheduler.c:324`** — clamp `maxExecutionTime` to the task's `desiredPeriod` (or a multiple thereof) to exclude physically impossible measurements:

```c
// Current (unbounded):
selectedTask->maxExecutionTime = MAX(selectedTask->maxExecutionTime, taskExecutionTime);

// Fix option A — cap to task period:
selectedTask->maxExecutionTime = MAX(selectedTask->maxExecutionTime, MIN(taskExecutionTime, selectedTask->desiredPeriod));

// Fix option B — delay statistics collection (disable calculateTaskStatistics until steady-state)
// Set calculateTaskStatistics = false during fc_init, re-enable after tasksInit() completes.
```

Option A is simpler and self-correcting. Option B matches the Betaflight approach more closely (they also added a startup grace period). Both may be appropriate.

The `desiredPeriod` for `TASK_SERIAL` at 100Hz is 10,000µs — any single-execution measurement above that is an initialization artifact and should be discarded.

## Related

- #1281 — separate issue: IMUF9001 raw CPU starvation at 32kHz (different root cause; different fix)

nerdCopter:
AI Generated pull-request

Fixes #1282

## Summary

On initial USB Configurator connect, `TASK_SERIAL` `maxExecutionTime` was permanently set to ~374ms (maxload ~3706%).

**Root cause:** At the moment the host OS opens the CDC serial port, the host USB CDC driver (`cdc_acm` / `usbser.sys` / `AppleUSBCDC`) has not yet begun polling the USB IN endpoint. On the embedded side, `Handle_USBAsynchXfer` (STM32 USB device firmware, `usbd_cdc_core.c:732`) arms the IN endpoint and sets `USB_Tx_State = BUSY`. Because no host ACK arrives, `USB_Tx_State` stays BUSY. `CDC_Send_DATA` (`usbd_cdc_vcp.c:168`) waits up to 2ms for `USB_Tx_State` to clear and returns 0. `usbVcpFlush` (`serial_usb_vcp.c`) retries for the full 50ms `USB_TIMEOUT` per MSP response write. The burst of MSP responses sent at connect time accumulates to ~354ms of blocking in a single TASK_SERIAL tick, permanently recording that as `maxExecutionTime`.

**Fix:**
- Port `schedulerIgnoreTaskExecTime()` / `schedulerIgnoreTaskStateTime()` / `schedulerIgnoreTaskExecRate()` / `schedulerGetIgnoreTaskExecTime()` framework from Betaflight 4.5 into `scheduler.c` / `scheduler.h`. Flag is reset before each task call and checked after; when set, `maxExecutionTime` is not updated for that tick.
- Call `schedulerIgnoreTaskExecTime()` inside `usbVcpFlush()` and `usbVcpWriteBuf()` (`serial_usb_vcp.c`) when `CDC_Send_DATA` cannot drain the full buffer in one call — i.e., when CDC TX backpressure is detected. Fires only on blocking writes; fast writes are unaffected. This suppression point is shared by both platform USB backends (`vcpf4` on F4, `vcp_hal` on F7/H7), since both expose the same bounded-timeout/partial-return contract via `CDC_Send_DATA`.
- Reset `TASK_SERIAL` `maxExecutionTime` on USB VCP connect edge (`fc_tasks.c`) so post-connect steady-state is reflected rather than any pre-connect accumulation.
- Also apply `schedulerIgnoreTaskExecTime()` at known slow-path sites: `fc/config.c` EEPROM write and PID profile switch, `interface/msp.c` `MSP_EEPROM_WRITE`, `flight/imu.c` acc-disabled branch, `io/beeper.c` early-return path.
- Add unit-test stubs for the new scheduler symbols in `beeper_dshot_beacon_unittest.cc` and `flight_imu_unittest.cc`.

**Flight safety:** fix is entirely in the USB serial write path (FC→Configurator). No change to gyro/PID, TASK_RX, failsafe, arming, or any RX protocol (SBUS, CRSF, ELRS all use UART DMA/interrupts). The trigger condition (USB host connect) cannot occur mid-flight, so flight testing cannot add evidence for or against this change; hardware verification is done via live `tasks` CLI observation across a USB connect event instead.

### `DEBUG_RC_SMOOTHING_RATE` — serial RX frame timing / regression (recommended for ELRS/CRSF)

Monitors RX frame interval at the `fc_rc.c` layer — protocol-agnostic. Confirms UART RX timing is unaffected by any SERIAL task changes.

| Channel | Value |
|---------|-------|
| 0 | `currentRxRefreshRate` (µs between RX frames — ELRS 500Hz expects ~2000µs steady) |
| 1 | `rcSmoothingData.training.count` (smoothing training step count) |
| 2 | `rcSmoothingData.averageFrameTimeUs` (calculated average frame time) |
| 3 | `sampleState` (guard time active) |

### `DEBUG_RC_INTERPOLATION` — RC command values and interpolation

| Channel | Value |
|---------|-------|
| 0 | `rcCommand[ROLL]` |
| 1 | `currentRxRefreshRate / 1000` (ms) |
| 2 | `rcInterpolationStepCount` |
| 3 | `setpointRate[ROLL]` |

### `DEBUG_RX_SIGNAL_LOSS` — RX signal health (protocol-agnostic)

| Channel | Value |
|---------|-------|
| 0 | `rxSignalReceived` (1 = valid signal) |
| 1 | `rxIsInFailsafeMode` (1 = failsafe active) |
| 3 | `rcData[THROTTLE]` |

### `DEBUG_USB` — VCP connect state (for reproducing/diagnosing this fix)

| Channel | Value |
|---------|-------|
| 0 | `usbCableIsInserted()` |
| 1 | `usbVcpIsConnected()` |

Set via CLI: `set debug_mode = RC_SMOOTHING_RATE` (flight regression) or `set debug_mode = USB` (VCP diagnosis) → `save`.

## Test plan

- [x] Unit tests pass
- [x] Build passes — 19+ targets across F4/F7, all CI groups green
- [x] Hardware verified on FOXEERF405 (F4, `vcpf4` USB backend; ELRS EP1 receiver attached, radio off): SERIAL `max/us=243`, `maxload=2.4%`; RX max/us=60 @ 32Hz unaffected (was 374,476µs / 3706% pre-fix)
- [x] Hardware verified on HELIOSPRING (F4, `vcpf4` USB backend; bare unsoldered): SERIAL `max/us=235`, `maxload=2.3%`
- [x] Hardware verified on FOXEERF722V4 (F7, `vcp_hal` USB backend; bare unsoldered): SERIAL `max/us=271`, `maxload=2.6%` — confirms fix on the other platform USB backend
- [x] CodeRabbit review addressed: removed overly-broad suppress calls from `msp_serial.c`

## Summary by CodeRabbit

* **Improvements**
  * Improved USB communication responsiveness by tightening transmission wait times and handling backpressure more efficiently.
  * Improved accuracy of scheduler and performance-load reporting by excluding unavoidable delays from timing measurements.
  * Prevented USB connection initialization and routine device operations from artificially inflating execution-time statistics.
  * Improved timing behavior for configuration saves, PID profile changes, EEPROM operations, sensor updates, and beeper activity.
* **Tests**
  * Updated unit-test support for the scheduler timing improvements.


nerdCopter:
@coderabbitAI review, but additionally analyze if this is the "proper" and "correct" resolution to the issue.

nerdCopter:

Addressed findings from review #4563611525:

**msp_serial.c:433-446 / 549-558 — schedulerIgnoreTaskExecTime in command dispatcher** — Actionable ✅ Fixed in `a1c78a8ff5`

CodeRabbit is correct. Both calls were redundant and overly broad — they suppressed maxload for every MSP command dispatch, not just slow-path operations. The real fix is already at the correct layer: `usbVcpFlush` and `usbVcpWriteBuf` in `serial_usb_vcp.c` only suppress when `CDC_Send_DATA` cannot drain the full buffer in one call (backpressure detected). Removed both `schedulerIgnoreTaskExecTime()` calls and the now-orphaned `#include "scheduler/scheduler.h"` from `msp_serial.c`.

**Docstring coverage 26.67% < 80%** — False positive

This is C firmware. The 80% docstring threshold is a CRabbit configuration artifact that does not apply to a C codebase — there is no docstring convention in EmuFlight or Betaflight firmware. No change needed.

Issue #1275, opened 2026-06-17
feat(rc_smoothing): 1€ (One Euro) adaptive RC filter — re-evaluate output filter order and open new PR


nerdCopter:
AI Generated issue-ticket

Low-priority feature request tracking the follow-up to the 1€ One Euro adaptive RC smoothing filter (closed PR #1247, closed PR #1291).

## Background

PR #1247 implemented a 1€ adaptive low-pass filter as an RC smoothing option — cutoff rises
with stick velocity and returns to fc_min at rest. It was closed pending an unresolved design
decision on output filter order.

PR #1291 implemented the dual-stage adaptive+fixed cascade (Option 2 below) and was itself
closed 2026-07-07 in favor of continued parallel evaluation across five design-variant branches.

Full design context and the original three options are documented in the PR #1247 closing comment:
https://github.com/emuflight/EmuFlight/pull/1247#issuecomment-4733912146

## Design Variants Evaluated

Five design-variant branches were built and compared in parallel:

- **Option 1 — single PT1 output** (`feat/1euro-single`)
- **Option 2 — fixed + adaptive PT1 cascade** (`feat/2euro`, ex-PR #1291)
- **Option 3 — double PT1 / PT2 output** (`feat/1euro-double`)
- **Option 4 — pilot-facing single/dual stage runtime toggle** (`feat/1euro-stage2`) — superseded
- **Option 5 — independent peer filter selections, no toggle** (`feat/1euro-2euro`) — **selected**

## Selected Design — Option 5 (`feat/1euro-2euro`)

`rc_smoothing_input_type` becomes a flat set of four peer selections: `PT1`, `1EURO`
(single-stage adaptive), `PT2`, `2EURO` (dual-stage adaptive). The legacy `PT3` slot is dropped —
confirmed negligible real-world EmuFlight usage, and upstream Betaflight itself retired the
entire PT1/PT2/PT3/BIQUAD selector model (PR #14683) in favor of a different auto-smoothing
architecture. `1EURO`/`2EURO` are independent choices with no runtime toggle and no hidden
config flag, avoiding the default-vs-label mismatch risk of the Option 4 toggle design.

All four 2EURO-era filter parameters (fc_min, fc_max, fc_d, beta) are fully internal/computed —
zero pilot tuning required in the normal case. `rc_smoothing_type` default remains
`INTERPOLATION`; explicit opt-in (`rc_smoothing_type = FILTER`) is required to use `1EURO`/`2EURO`.

[...]

PR #1291, opened 2026-07-04
2EURO: dual-stage (1€ adaptive + static fixed) RC smoothing filter

nerdCopter:
AI Generated pull-request

## 🧪 Calling for flight testers — full-stick freestyle needed

Hover-only validation is done and clean (two sessions, see below). What's still needed before this can leave DRAFT: **full-stick freestyle flight testing** — hard flips, fast rolls, aggressive acro — on your own hardware.

If you can test:
1. Flash this branch.
2. Set the **exact CLI parameters below** — copy/paste all of them, even the ones that may already match your defaults. (A prior flight test on this feature silently ran the wrong filter for an entire session because one required parameter was missing from a saved diff — don't rely on assumed defaults, set them explicitly.)
3. Fly full-stick freestyle/acro. Report back in a PR comment:
   - Does it feel over-filtered / mushy / delayed, or fine, at full-stick inputs?
   - Any perceptible difference vs. your current filter (PT1/PT2/PT3/INTERPOLATION)?
   - **Attach your BBL blackbox log** (or a link to it) so it can be analyzed alongside your feedback.

### CLI Setup — copy/paste exactly

```
set rc_smoothing_type = FILTER
set rc_smoothing_input_type = 2EURO
set debug_mode = RC_SMOOTHING
set rc_smoothing_debug_axis = ROLL
save
```

- `rc_smoothing_type = FILTER` — **required**. Without this, no smoothing filter (including 2EURO) runs at all; the FC falls back to INTERPOLATION regardless of the setting below.
- `rc_smoothing_input_type = 2EURO` — **required**. Selects the new filter. This is the one setting most likely to silently revert if you restore an old saved diff/backup after flashing — always re-set it explicitly.
- `debug_mode = RC_SMOOTHING` — needed so your blackbox log captures the filter's internal state (raw input, filtered output, adaptive cutoff, link-rate training) for analysis. Without it, your BBL log can still be used for general PID/gyro tracking feedback, but not for filter-internals analysis.
- `rc_smoothing_debug_axis = ROLL` — which axis populates the debug slots above. ROLL is the convention used in all prior test sessions for this PR, for consistency across logs.

No other 2EURO-related parameters exist to set — the filter is fully zero-config (see below).

### Troubleshooting — if your settings don't stick / look reset after flashing

This PR bumps `PG_RX_CONFIG` from `2` (current master) to `3`. On first boot after flashing, the FC detects that version change and does an **automatic full reset of `rxConfig_t`** — every RX-related setting, not just the 2EURO ones — back to firmware defaults. This is expected behavior, not a bug or a failed save.

What this means for you:
- **Flashing this branch for the first time**: after your first boot, re-apply the CLI Setup block above and `save` again. Any pre-existing RX-related settings (channel map, RSSI config, etc.) will have been reset to firmware defaults along with the smoothing fields.
- **Upgrading from an earlier test build of this same branch**: development builds used `PG_RX_CONFIG` values up to `5` before being cleaned up to a proper single `2→3` bump — so you'll get one more full `rxConfig_t` reset on this build too. Re-apply CLI Setup + `save` after flashing, same as a first-time flash.

If settings still don't appear to save after that:
1. `defaults nosave` — loads clean firmware defaults into RAM without writing EEPROM yet, so you can inspect before committing.
2. Re-apply the CLI Setup block above.
3. `save` explicitly — don't rely on an implicit save from a reboot or USB disconnect.
4. If it's still wrong, plain `defaults` (no `nosave`) forces a full reset **and** auto-saves + reboots — re-apply CLI Setup + `save` once more afterward.

This is general PG-version-bump behavior, not specific to 2EURO — worth knowing any time settings unexpectedly reset after an EmuFlight upgrade.

## Summary

Introduces **2EURO**, a dual-stage RC smoothing filter derived from the [1€ filter](https://gery.casiez.net/1euro/) (Casiez et al., CHI 2012). Stage 1 is the unmodified 1€ adaptive PT1 (cutoff rises with stick velocity). Stage 2 is a new, static PT1 cascaded after it at a fixed cutoff (`fc_fixed = 2 × fc_min`), providing an always-on quantization floor that stage 1 alone cannot: stage 1's own cutoff opens precisely during the fast inputs where RC-frame steps are largest, so a single adaptive PT1 (previously tested as `feat/1euro-single`) let 37.1% of a quantization step through during fast flips — confirmed unacceptable in blackbox testing.

```
raw RC input → [Stage 1: 1€ adaptive PT1] → [Stage 2: static PT1, fc_fixed=2×fc_min] → rcCommand
```

**Fully zero-config — no pilot-tunable filter parameters exist.** Every value the filter needs is auto-calculated from data the firmware already has:
- `fc_min` — auto-scales with detected RC link rate (`rx_hz/12`, floored at 6 Hz, with no fixed upper Hz value — the ratio to Nyquist stays mathematically consistent across the entire link-rate range, not just up to some cutoff rate)
- `fc_d` (derivative filter cutoff) — auto-scales with link rate (`rx_hz/19`)
- `fc_fixed` — always `2 × fc_min`, recomputed whenever fc_min changes
- **beta** (speed coefficient) — derived from the active rate profile (rcRates/rates for roll+pitch), recalculated on every rate-profile-change event

None of these have a CLI, MSP, or configurator field — there is nothing to configure beyond selecting `rc_smoothing_type = FILTER` and `rc_smoothing_input_type = 2EURO` (see CLI Setup above). (An earlier revision of this PR left `fc_min`/`fc_max`/`deriv_hz` as CLI-only edge-case overrides while `beta` had none — an inconsistency, since beta was the one parameter pilots were previously told they might need to hand-tune. Resolved by removing all three rather than adding a beta override back; see "Design rationale" below.)

## Design rationale (condensed — full math in `CONTEXT_2euro.md`)

- **Stage order is fixed**: adaptive PT1 must run first. The 1€ algorithm computes stick velocity from stage 1's own filtered state as the previous-sample reference; running the static stage first would make the derivative measure the velocity of an already-lagged signal, delaying cutoff-opening at stick onset.
- **`fc_fixed = 2×fc_min`** keeps the Nyquist-to-cutoff ratio constant (6:1) across the full link-rate range — verified mathematically consistent (~94.8% attenuation) from ~72 Hz through at least 2400 Hz, covering every real RC protocol (SBUS 50 Hz through ELRS 1000 Hz full-rate). An earlier revision hard-clamped `fc_min` to a 40 Hz ceiling that only reproduced this ratio exactly at 500 Hz; above that the ratio grew past 6:1 (over-attenuating). Fixed by removing the fixed ceiling — the only remaining bound is a 200 Hz safety constant that engages only for implausibly fast RX rates (>2400 Hz).
- **Auto-beta recalculation is wired through `initRcProcessing()`**, which fires on all 8 rate-profile-change paths (aux switch, stick combo, MSP, CLI, config load, OSD profile switch, OSD rate-value-edit save) — not just at boot or arm.
- **A prior version of this branch had a real bug**: stage 1/stage 2 filter-gain (`k`) values were computed using the RC-frame-rate `dT` but applied every PID loop, giving an effective cutoff ~8× higher than intended. Fixed by splitting the struct into `dT_inv` (RC-frame rate, used by the derivative estimator) and `pid_dT_inv` (PID rate, used by both signal-path PT1 stages). Confirmed fixed by real hardware flight data (below): observed `fc_min` floor exactly matches the expected value.
- **A second real bug, found via independent CodeRabbit analysis and confirmed against flight data, now confirmed fixed on a second flight session**: the adaptive stage only recomputed its derivative/cutoff when the raw stick value *changed* (`input != lastInput`). If the stick moved fast and then held perfectly steady (no further raw-value change across several RX frames), the elevated post-move cutoff froze indefinitely instead of decaying back toward `fc_min`. Fixed by gating on a genuine new-RX-frame signal (`isRXDataNew`, already computed elsewhere in `fc_rc.c` but not previously threaded into the filter) instead of value equality. Note: this bug always erred in the *stuck-open* (less filtering) direction, not stuck-closed — it cannot explain over-filtering/mushiness, and the flight data shows none, in either session.
- **Whoop-specific beta tuning remains an open question** — a common assumption is that whoops want a lower beta (gentler adaptive response), but the current formula is driven by rate-profile aggressiveness, not airframe size: a whoop pilot flying gentle rates would get a *higher* auto-beta from this formula, not lower. This needs a formula-level fix if it matters in practice, not a manual override (deliberately not added — see zero-config rationale above).
- **Minor**: `fc_min` was truncated to the next-lower integer Hz instead of rounded (e.g. `500/12=41.67` became `41`, not `42`). Switched to `lrintf()` rounding.
- **Minor**: `PG_RX_CONFIG` had accumulated dev-time version churn (`2→3→4→3→4→5`) across this branch's iteration history, including one legitimate re-bump that landed on a version number (`4`) already used earlier for a different struct layout (that earlier use had itself already been reverted once — a real instance of a dev-bump getting overwritten). No actual EEPROM corruption occurred since every flashed/saved test build always differed from master's version, but shipping the accumulated number made no sense. Reset to a clean `2→3` (master+1). See "Troubleshooting" above for what this means if you're re-flashing over an earlier test build.

## Hardware flight validation

**Session 1 (2026-07-04)** — FOXEERF722V4, 250 Hz ELRS, 3 backyard hover flights (2 acro-style, 1 stable+spring stick-release oscillation test), flown *before* the two bugs above were found:

| Metric | Acro 1 | Stable+Spring | Acro 2 |
|---|---|---|---|
| fc_min floor observed (expected 20.8 Hz) | 20.0 Hz | 20.0 Hz | 20.0 Hz |
| Adaptive cutoff range | 61.1 Hz | 152.7 Hz | 56.2 Hz |
| Setpoint→gyro system lag (whole control chain) | 17.7 ms | 18.6 ms | 19.0 ms |
| Raw/filtered tracking correlation | 0.9986 | 0.992 | 0.9982 |
| Frozen-cutoff bug exposure (pre-fix) | 0% | 22 truly-frozen episodes | 2 truly-frozen episodes |

Verdict: no over-filtering, mushiness, or excess delay attributable to 2EURO, including from the frozen-cutoff bug (wrong direction to cause mushiness).

**Session 2 (2026-07-05)** — same hardware, firmware including both bug fixes, 3 more hover flights:

| Metric | Acro 1 | Stable+Spring | Acro 2 |
|---|---|---|---|
| fc_min floor observed (rounding fix confirmed) | 21.0 Hz | 21.0 Hz | 21.0 Hz |
| Setpoint→gyro system lag | 18.3 ms | 17.9 ms | 19.0 ms |
| Frozen-cutoff bug — truly-frozen episodes | 0 | **0** (was 22) | 0 (was 2) |

**Fix confirmed in the field**: traced one held-stick episode end-to-end (stick pegged at max for 150ms) showing the adaptive cutoff correctly decaying 79.9→23.0 Hz across the hold instead of freezing. Lag is unchanged from session 1 — no regression.

Both sessions are hover-only — **full-stick freestyle is the missing piece**, which is what this PR needs testers for.

## ⚠️ Testing defaults are live — must change before merge

This branch currently defaults to `rc_smoothing_type = FILTER` and `rc_smoothing_input_type = 2EURO` (`src/main/pg/rx.c`, marked `// TESTING DEFAULT — DO NOT MERGE`) so flight testing needs zero CLI setup out of the box. **Mergeable defaults are `INTERPOLATION` / `PT2`** (current master parity) and must be restored in the same commit that removes DRAFT status. (Still set the CLI parameters above explicitly when testing — don't rely on the branch default in case you're testing against a diff/config that predates it.)

## Test plan

- [x] Unit tests pass; all target builds (Bench/Compile/Unique sets) pass clean
- [x] First hardware hover validation (FOXEERF722V4, 250 Hz ELRS) — no over-filtering/mushiness/delay found
- [x] Frozen-cutoff bug found (independent CodeRabbit analysis + flight-data confirmation) and fixed
- [x] Re-flown with the frozen-cutoff fix — confirmed 0 frozen episodes (was 24), no regression
- [ ] **Full-stick freestyle flight (hard flips, fc_max behavior, onset feel) — community testers needed, see call-out above**
- [ ] Comparison flight against the sibling design-variant branches (not part of this PR): `feat/1euro-single` (single-stage, already confirmed unacceptable under fast input — see above), `feat/1euro-double` (double-stage, open question), and `feat/1euro-stage2` (this design with a pilot-facing `rc_smoothing_1euro_stage2` toggle to switch to single-stage without reflashing). All three now share this branch's bug fixes and zero-config auto-beta design, so the comparison is apples-to-apples
- [ ] Restore mergeable defaults (`INTERPOLATION`/`PT2`) before removing DRAFT
- [ ] Pilot opinion on full-stick feel

Closes/references: emuflight/EmuFlight#1275 (tracking issue). Does not reopen the earlier closed draft (#1247) — this is a fresh PR after the design decision and initial hardware validation.


nerdCopter:

This is a draft PR (automatic pass was skipped for that reason) but I'd like your independent take on it now, on demand. Please don't just summarize the PR description back to me — I want you to actually walk the diff and tell me what you genuinely think, including anything that contradicts or is unverified by the description's claims.

Specifically:

1. **The new dual-stage filter design** (`src/main/common/filter.c`, `filter.h`): stage 1 is an adaptive PT1 (cutoff rises with `fc_min + beta*|derivative|`), stage 2 is a static PT1 cascaded after it at `fc_fixed = 2 * fc_min`. Does the cascade math check out? Is stage 1 running before stage 2 actually the correct order given how the derivative estimate is computed from stage 1's own filtered state?

2. **`rcSmoothingSetFilterCutoffs()` in `src/main/fc/fc_rc.c`**: `fc_min` now scales as `rx_hz/12`, floored at 6 Hz, bounded only by a `RC_SMOOTHING_2EURO_FC_MAX` constant (200 Hz) that should only ever engage for RX rates above ~2400 Hz. Check the actual arithmetic and the `constrainf()` call — does it behave the way that's claimed for realistic RX rates (50–1000 Hz) and does the ceiling actually only bite where claimed?

3. **The PG version bump and field removal** (`src/main/pg/rx.h`, `rx.c`, `src/main/interface/settings.c`): three `rxConfig_t` fields were deleted entirely (not just hidden from CLI). Confirm there's no dangling reference anywhere else in the tree (MSP, configurator-facing code, other filters) that still expects those fields to exist, and that the PG version bump is actually necessary/sufficient for the struct change.

4. **Auto-beta recalculation** (`initRcProcessing()` path): is the claim that beta gets recomputed on every rate-profile-change event actually true end-to-end, or are there call paths that could leave it stale?

5. Anything else that looks off — magic numbers without justification, integer truncation issues, thread-safety/ISR-context concerns, missing null checks, anything you'd flag in a firmware PR like this regardless of what the description says.

Genuinely critical feedback is welcome — this is pre-merge flight-test code and I'd rather find problems now.

nerdCopter:

Thanks for walking the actual diff instead of the PR text. Went through all 4 points against current code (not just the excerpts above) — here's the resolution status on each:

**1. Adaptive-cutoff freeze on value-equality gating (your "biggest issue" in #1)** — Confirmed real, fixed in `583841496f`. `twoEuroFilterApply()` now gates the derivative/cutoff recompute on `newSample` — an actual new-RX-frame signal (`isRXDataNew`, `fc_rc.c:84`) — instead of `input != lastInput` (current: `filter.c:63`). Field-validated on 2026-07-05: traced a held-stick episode end-to-end and confirmed the cutoff decays 79.9→23.0 Hz across a 150ms hold instead of freezing. A detector re-run across two hover sessions found 24 truly-frozen episodes pre-fix, 0 post-fix.

**2. Stale `x_filter` after profile/RX-rate change (your caveat in #4)** — Resolved as a side effect of #1, not a separate change. `twoEuroFilterUpdate()` still only refreshes `d_filter`/`x_filter_fixed`/stored params immediately (`filter.c:32-46`), but `x_filter`'s own `k` now recomputes on the very next `newSample=true` call — i.e. within one RX frame — because that gate no longer depends on the stick value actually changing. No further code change was needed once the newSample gating landed.

**3. `inputCutoffFrequency` truncation (#2)** — Confirmed and fixed: `rcSmoothingSetFilterCutoffs()` now does `(uint16_t)lrintf(constrainf(rx_hz / 12.0f, 6.0f, RC_SMOOTHING_2EURO_FC_MAX))` (`fc_rc.c:355`) instead of a bare truncating cast, so 250 Hz RX → 21 Hz (rounded), not 20 Hz. Field-confirmed in the 2026-07-05 flight data. Kept the field as `uint16_t` rather than widening to `uint32_t` — it's shared storage with PT1/PT2/PT3's cutoff, sub-Hz precision doesn't change any conclusion for this filter, and widening would be an unrelated, broader-scoped change than this PR should make.

**4. Enum reuse — BIQUAD/PT4 values reassigned to 2EURO (your caveat in #3)** — Confirmed, and intentional: BIQUAD and PT4 are being dropped from this codebase, so reusing their old enum values is deliberate, not incidental. No MSP payload data for the old fields is read or interpreted by design — verified `msp.c:1077-1079`/`2227-2229` write/read placeholder bytes only. The PG_RX_CONFIG bump is 4→5 specifically for the field-removal commit (`de4cd98fd4`); vs. master (currently at 2) that's the 2→5 your script saw, across this branch's several PG-touching commits. Old configurator/tooling assuming `1=BIQUAD`/`4=PT4` is an accepted risk on this experimental branch, not a bug we're fixing.

Net: your #1 (freeze) and #2 (truncation) were real bugs, now fixed and field-validated on real hardware. The profile-staleness point resolved as a consequence of the same fix rather than needing its own patch. The enum-reuse point is a confirmed, intentional design tradeoff.

nerdCopter:
AI Generated comment

Closing this draft. Two additional design-variant branches have been created since this PR was opened, exploring the same 2EURO dual-stage design from different angles:

- `feat/1euro-stage2` — added a pilot-facing `rc_smoothing_1euro_stage2` toggle to switch between single-stage and dual-stage at runtime without reflashing. Now superseded by the branch below.
- `feat/1euro-2euro` — replaces that toggle with `1EURO`/`2EURO` as independent, directly selectable `rc_smoothing_input_type` values (same relationship as `PT1`/`PT2`), reusing the `PT3` slot after confirming it has negligible real-world adoption in both Betaflight's and EmuFlight's own history. This is the current leading design-variant branch.

None of the branches have full-stick freestyle flight validation yet — only `feat/2euro` has been flown, hover-only. Will reopen a fresh PR referencing tracking issue #1275 once a design decision is made and flight-tested.

PR #1306, opened 2026-07-13
Fix CRAZYFLIE2/CRAZYFLIE2BQ: clamp numAuxChannels to prevent OOB read/write

### LINKED ISSUE #1303: CRAZYFLIE2/serialrx.c: unbounded numAuxChannels loop can write past channelData bounds
AI Generated issue-ticket

## Summary
`crtpCommanderCPPMEmuPacket_t.hdr.numAuxChannels` is a 4-bit bitfield (representable range 0-15), but the code that consumes it never validates it against `CRTP_CPPM_EMU_MAX_AUX_CHANNELS` (10) — the actual size of both `channelAux[]` and the implicit bound on `channelData[]`. A value of 11-15 causes an out-of-bounds read from `channelAux[]` and an out-of-bounds write to the static `channelData[]` array.

## Location
https://github.com/emuflight/EmuFlight/blob/5a74854dec5312b19d1cdcf91ec6714f4924fe4b/src/main/target/CRAZYFLIE2/serialrx.c#L106-L107

```c
uint8_t i;
for (i = 0; i < crtpCppmPacket->hdr.numAuxChannels; i++) {
    channelData[i + 4] = crtpCppmPacket->channelAux[i];
}
```

Struct definition:
https://github.com/emuflight/EmuFlight/blob/5a74854dec5312b19d1cdcf91ec6714f4924fe4b/src/main/target/CRAZYFLIE2/crtp.h#L72-L82

```c
#define CRTP_CPPM_EMU_MAX_AUX_CHANNELS 10
typedef struct crtpCommanderCPPMEmuPacket_s {
    struct {
        uint8_t numAuxChannels : 4;   // Set to 0 through CRTP_CPPM_EMU_MAX_AUX_CHANNELS
        uint8_t reserved : 4;
    } hdr;
    ...
    uint16_t channelAux[CRTP_CPPM_EMU_MAX_AUX_CHANNELS];
} __attribute__((packed)) crtpCommanderCPPMEmuPacket_t;
```

`channelData` is declared as `static uint32_t channelData[SUPPORTED_CHANNEL_COUNT]` where `SUPPORTED_CHANNEL_COUNT = 4 + CRTP_CPPM_EMU_MAX_AUX_CHANNELS = 14` (valid indices 0-13). A `numAuxChannels` value of 15 writes `channelData[4]` through `channelData[18]`, five elements past the end of the array, and reads `channelAux[10]` through `channelAux[14]`, past the end of the 10-element field in the packed struct.

## Attack surface
`numAuxChannels` is part of an incoming CRTP packet, delivered via the Crazyflie's onboard nRF51 radio module over UART/syslink (see the file's own header comment) — i.e. it originates from external radio data, not a value the flight controller computes itself. Nothing clamps or rejects an out-of-range value before the loop uses it.

## Scope
- Identical code exists in `src/main/target/CRAZYFLIE2/serialrx.c` and `src/main/target/CRAZYFLIE2BQ/serialrx.c` (both boards affected equally).
- Confirmed present on current `master` (`5a74854dec`) — pre-existing, unrelated to the atomic-target-split work in PR #1302.
- Confirmed there is no Betaflight equivalent to check against: Betaflight has no Crazyflie target or CPPM-emulation packet handling at all (checked `4.5-maintenance` branch and the Betaflight target-configs reference repo — no matches in either).

## Suggested fix
Clamp `numAuxChannels` to `CRTP_CPPM_EMU_MAX_AUX_CHANNELS` (or reject the packet) before the loop runs.

nerdCopter:
AI Generated pull-request

## Summary
`crtpCommanderCPPMEmuPacket_t.hdr.numAuxChannels` is a 4-bit bitfield (representable range 0-15) taken directly from an incoming CRTP packet delivered over UART/syslink from the Crazyflie's onboard nRF51 radio module. The loop that consumes it never validated the value against `CRTP_CPPM_EMU_MAX_AUX_CHANNELS` (10), the actual size of both `channelAux[]` and the implicit bound on the static `channelData[]` array. A value of 11-15 caused an out-of-bounds read from `channelAux[]` and an out-of-bounds write to `channelData[]`.

Fix clamps `numAuxChannels` to `CRTP_CPPM_EMU_MAX_AUX_CHANNELS` before the loop runs. TAER (thrust/roll/pitch/yaw) channels are written unconditionally earlier in the same function, so clamping (rather than rejecting the whole packet) preserves that data plus any valid aux channels.

Identical fix applied to both `src/main/target/CRAZYFLIE2/serialrx.c` and `src/main/target/CRAZYFLIE2BQ/serialrx.c`, which carry byte-identical copies of the affected function.

Closes #1303

## Test plan
- [x] Build passes: `make CRAZYFLIE2 CRAZYFLIE2BQ` — both succeed, no warnings
- [x] Unit tests: `make clean_test test` — all 39 suites pass
- [x] Hardware verified: n/a — neither CRAZYFLIE2 nor CRAZYFLIE2BQ is a hardware-verify target

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved stability and security when processing malformed CPPM emulation packets.
  - Auxiliary channel data is now safely limited to the supported maximum, preventing invalid memory access.


nerdCopter:
@coderabbitai review

PR #1321, opened 2026-07-24
build: add -fmerge-all-constants to LTO optimisation flags

nerdCopter:
AI Generated pull-request

## Summary
- Closes the one remaining LTO flag gap between EmuFlight and Betaflight 4.5-maintenance: adds
  `-fmerge-all-constants` to `OPTIMISATION_BASE` in `Makefile:137`. `-flto`/`-fuse-linker-plugin`/
  `-ffast-math` were already present since 2017.
- Verified fresh under the GCC 13.3.1 toolchain from #1286, not the GCC9 toolchain on current
  master, since GCC13 already required per-function optimize overrides for codegen/timing
  regressions — old GCC9 measurements don't carry over.

## Blocked on #1286
This PR is stacked on #1286 (build: upgrade xpack arm-none-eabi-gcc 9.3.1 → 13.3.1) and must not
be merged before it. The parent branch only exists on the fork, so GitHub can't set it as this
PR's base against upstream — the diff below includes #1286's full changeset in addition to this
PR's actual new change, which is only `Makefile:137` (adding `-fmerge-all-constants`). Everything
else in the diff is #1286, already under separate review there.

## Verification
- 33 unique targets checked (23-target bench/compile/unique sweep + 5 H7 family + 5 F405-composition
  recheck): all shrink flash under GCC13, zero growth, zero new warnings.
- ITCM sub-region (F722-class, 16KB hard cap) independently checked on 8 boards including
  AIRBOTF7HDV (fleet's tightest board, 98.19% used): zero delta from this flag.
- RAM unchanged or slightly improved on every target checked.
- CodeRabbit: one finding (`-fmerge-all-constants` non-conforming constant-merge risk) resolved
  BF-parity-justified — Betaflight 4.5-maintenance ships this exact flag in production; grepped
  `src/main` for the specific risk pattern, no genuine instance found.
- Build-time impact: inconclusive — repeat measurement showed the signal was within the baseline's
  own run-to-run noise; not a confirmed regression.

## Test plan
- [x] `make clean_test && make test` — unit tests pass
- [x] Bench/Compile/Unique target sweep — all build clean, 0 warnings
- [ ] Hardware loop-timing verification of the combined configuration (`GCC_OPTIMIZE_SIZE`
      overrides + `-fmerge-all-constants` + GCC13 together) — not yet flown; #1286's own hover
      test covered `GCC_OPTIMIZE_SIZE` alone

## Summary by CodeRabbit

* **Performance**
  * Improved application build optimization, which may result in smaller binaries and enhanced runtime performance.


nerdCopter:
@coderabbitai review

PR #1324, opened 2026-07-24
fix(pid): unify directFF pure-P-term gate with pidLevel() call site

### LINKED ISSUE #1323: directFF pure-P-term fix (#1257) doesn't cover GPS_RESCUE_MODE — duplicated flight-mode gate has drifted
AI Generated issue-ticket

## Problem

`src/main/flight/pid.c` currently gates which flight modes call `pidLevel()` and which
directFeedForward source to use with two independently-maintained conditionals that have
drifted out of sync (verified against master @ 20214a15158bc080c75f63596c25f16a383c9d46):

Call site (decides *whether* `pidLevel()` runs, ~L718-723):
```c
if (axis == FD_YAW) {
} else if (FLIGHT_MODE(GPS_RESCUE_MODE)) {
    currentPidSetpoint = pidLevel(axis, pidProfile, angleTrim, currentPidSetpoint);
} else if ((FLIGHT_MODE(ANGLE_MODE) || FLIGHT_MODE(HORIZON_MODE)) && !FLIGHT_MODE(NFE_RACE_MODE)) {
    currentPidSetpoint = pidLevel(axis, pidProfile, angleTrim, currentPidSetpoint);
} else if ((FLIGHT_MODE(ANGLE_MODE) || FLIGHT_MODE(HORIZON_MODE)) && FLIGHT_MODE(NFE_RACE_MODE) && (axis != FD_PITCH)) {
    currentPidSetpoint = pidLevel(axis, pidProfile, angleTrim, currentPidSetpoint);
}
```

directFF-source site (decides *which value* directFeedForward multiplies, ~L859, added in #1257):
```c
if ((FLIGHT_MODE(ANGLE_MODE) || FLIGHT_MODE(HORIZON_MODE)) && axis <= FD_PITCH) {
    directFeedForward = angleModeP[axis] * directFF[axis];   // pure-P, fixed
} else {
    directFeedForward = currentPidSetpoint * directFF[axis]; // pre-fix, D/F-contaminated
}
```

`GPS_RESCUE_MODE` is present in the first gate but absent from the second. `pidLevel()`
still runs for GPS Rescue and still populates `directFF[axis]` from `df_angle_low`/
`df_angle_high` unconditionally — GPS Rescue is not excluded from using directFF, only
from the pure-P-term fix that #1257 applies to it.

## Why this matters

Inside `pidLevel()`, GPS Rescue's angle command feeds the identical computation stick
input uses:

```c
angle += gpsRescueAngle[axis] / 100;   // GPS Rescue's demand folds into the same "angle" var
...
d_term_low  = (1 - errorAnglePercent) * (attitudePrevious[axis] - getAngleModeAngles(axis)) * D_angle_low;
d_term_high = errorAnglePercent       * (attitudePrevious[axis] - getAngleModeAngles(axis)) * D_angle_high;
...
f_term_low = (angle - previousAngle[axis]) * F_angle / dT;
```

Neither `d_term` (derivative of actual attitude change) nor `f_term` (derivative of the
setpoint) distinguishes a stick-driven setpoint change from a GPS-Rescue-driven one. Under
a fast GPS Rescue correction (wind gust, GPS position jump, mode-entry from an off-nominal
attitude), the same D/F-transient-into-directFF mechanism #1257 fixes for angle/horizon
mode can occur during GPS Rescue — and never gets the fix, regardless of magnitude.

There is at least one user report of unintended climb/flyoff when GPS Rescue activates
that may share this root cause. This is a hypothesis, not a confirmed finding — no
hardware test exists for the GPS Rescue case specifically.

## Proposed fix — remove the duplicated gate, not just patch it

Adding `|| FLIGHT_MODE(GPS_RESCUE_MODE)` to the L859 condition would close this specific
gap, but leaves two independently-maintained conditionals that can drift again for the
next mode that calls `pidLevel()` (exactly how this gap was introduced). Prefer deriving
the L859 choice from a single flag set at the point `pidLevel()` actually runs (e.g. a
per-axis bool written inside `pidLevel()`, read at L859) so there is one source of truth
instead of two gates that must be kept in sync by hand. This also simplifies the code by
removing a duplicated flight-mode check.

## Scope

Out of scope for #1257 (that PR is unrelated in disposition — hardware-verified
ineffective for its own target case, per discussion on that PR). This is a distinct gap
in directFF's flight-mode gating, independent of whether #1257 itself is merged, reworked,
or closed.

Related: #1257, #1184

nerdCopter:
AI Generated pull-request

## Summary

Resolves #1323. `pid.c` gated which flight modes call `pidLevel()` (call site, ~L717-724 —
includes `GPS_RESCUE_MODE`) and which directFF source to use (~L859, added in #1257 —
`ANGLE_MODE`/`HORIZON_MODE` only) with two independently-maintained conditionals that had
drifted: GPS Rescue populated `directFF[axis]` via `pidLevel()` but never got routed through the
pure-P-term fix from #1257, silently falling back to the pre-fix D/F-contaminated
`currentPidSetpoint`-based calc.

## Changes (`src/main/flight/pid.c`, 4 lines added, 1 changed)

- Added `static FAST_RAM_ZERO_INIT bool pidLevelActive[2];` next to `angleModeP[2]` (same sizing
  — `pidLevel()` never runs for YAW).
- Reset `pidLevelActive[axis] = false` each cycle alongside the existing `directFF[axis] = 0.0f`
  reset (~L713-716).
- Set `pidLevelActive[axis] = true` inside `pidLevel()` itself (~L544, next to the existing
  `angleModeP[axis]` capture) — single point of truth, regardless of which of the three
  call-site branches invoked it.
- L859 now reads `axis <= FD_PITCH && pidLevelActive[axis]` instead of re-deriving
  `(FLIGHT_MODE(ANGLE_MODE) || FLIGHT_MODE(HORIZON_MODE)) && axis <= FD_PITCH`. `axis <= FD_PITCH`
  kept for array-bounds safety (same reasoning as #1257).

Call-site conditional (~L717-724) intentionally left untouched — collapsing its three branches
into one is a separate simplification not required to close this gap.

Net effect: `GPS_RESCUE_MODE` gets the same pure-P-term directFF treatment as
`ANGLE_MODE`/`HORIZON_MODE`, and any future flight mode calling `pidLevel()` inherits the fix
automatically with no second gate to remember to update.

## Test plan

- [x] Unit tests: `make test` — 0 failures
- [x] Compile-gate real-aircraft targets: `make HELIOSPRING FOXEERF722V4 STELLARH7DEV` — 3
      succeeded, 0 failed
- [ ] Hardware flight test — not performed. This is PID control logic, not a
      flash/SPI/gyro/DMA/USB driver change, so mandatory hardware-verification doesn't apply;
      the RTH-flyoff magnitude in real flight remains unconfirmed per #1323 (this closes the
      structural gate gap, not a hardware-validated behavior change)

Related: #1257, #1184, #1323

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved roll/pitch direct-feedforward scaling by using whether leveling control actually ran during the current PID cycle, instead of depending on flight-mode conditions.
  * Updated handling when non-YAW direct feedforward is disabled to ensure consistent roll/pitch behavior.

* **Tests**
  * Added regression coverage confirming GPS RESCUE uses the same PID-level direct-feedforward path as ANGLE mode.
  * Added coverage confirming NFE RACE engages PID-level only for roll (with pitch matching expected behavior).

### DIFF
--- a/src/main/flight/pid.c
+++ b/src/main/flight/pid.c
@@ -354,6 +354,7 @@ typedef struct pidCoefficient_s {
 static FAST_RAM_ZERO_INIT pidCoefficient_t pidCoefficient[XYZ_AXIS_COUNT];
 static FAST_RAM_ZERO_INIT float directFF[3];
 static FAST_RAM_ZERO_INIT float angleModeP[2];
+static FAST_RAM_ZERO_INIT bool pidLevelActive[2]; // true when pidLevel() ran for this axis this cycle, regardless of which flight mode triggered it
 static FAST_RAM_ZERO_INIT float maxVelocity[XYZ_AXIS_COUNT];
 static FAST_RAM_ZERO_INIT float feathered_pids;
 static FAST_RAM_ZERO_INIT float setPointPTransition[XYZ_AXIS_COUNT];
@@ -542,6 +543,7 @@ static float pidLevel(int axis, const pidProfile_t *pidProfile, const rollAndPit
 
     currentPidSetpoint = p_term_low + p_term_high;
     angleModeP[axis] = p_term_low + p_term_high;
+    pidLevelActive[axis] = true;
     currentPidSetpoint += d_term_low + d_term_high;
     currentPidSetpoint += f_term_low;
 
@@ -712,6 +714,7 @@ void pidController(const pidProfile_t *pidProfile, const rollAndPitchTrims_t *an
         // disable directFF for pitch and roll while not in angle
         if (axis != FD_YAW) {
             directFF[axis] = 0.0f;
+            pidLevelActive[axis] = false;
         }
 
         if (axis == FD_YAW) {
@@ -856,7 +859,7 @@ void pidController(const pidProfile_t *pidProfile, const rollAndPitchTrims_t *an
         // since yaw acts different this will only work for yaw
         // actually this also works in angle mode so pitch and roll have angle mode values
         float directFeedForward;
-        if ((FLIGHT_MODE(ANGLE_MODE) || FLIGHT_MODE(HORIZON_MODE)) && axis <= FD_PITCH) {
+        if (axis <= FD_PITCH && pidLevelActive[axis]) {
             directFeedForward = angleModeP[axis] * directFF[axis];
         } else {
             directFeedForward = currentPidSetpoint * directFF[axis];
--- a/src/test/unit/pid_unittest.cc
+++ b/src/test/unit/pid_unittest.cc
@@ -442,12 +442,12 @@ TEST(pidControllerTest, testPidHorizon) {
     EXPECT_FLOAT_EQ(0, pidData[FD_YAW].I);
 }
 
-// Identical priming + step sequence under the given flight mode so pidLevel()'s internal state starts equal for any mode compared.
-void primeAndStepUnderFlightMode(flightModeFlags_e flightMode) {
+// Identical priming + step sequence under the given flight mode mask (0 = plain rate/acro) so pidLevel()'s internal state starts equal for any mode compared.
+void primeAndStepUnderFlightMode(int flightModeMask) {
     resetTest();
     ENABLE_ARMING_FLAG(ARMED);
     pidStabilisationState(PID_STABILISATION_ON);
-    enableFlightMode(flightMode);
+    enableFlightMode(static_cast<flightModeFlags_e>(flightModeMask));
 
     // Default test profile has PID_LEVEL_LOW.I=0, which zeroes directFF entirely.
     pidProfile->pid[PID_LEVEL_LOW].I = 70;
@@ -468,7 +468,7 @@ void primeAndStepUnderFlightMode(flightModeFlags_e flightMode) {
     attitude.values.pitch = -550;
     pidController(pidProfile, &rollAndPitchTrims, currentTestTime());
 
-    disableFlightMode(flightMode);
+    disableFlightMode(static_cast<flightModeFlags_e>(flightModeMask));
 }
 
 TEST(pidControllerTest, testGpsRescueUsesPidLevelDirectFF) {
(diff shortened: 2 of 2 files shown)


nerdCopter:
AI Generated comment

Addressed CodeRabbit finding from review 4775922042 (nitpick: add regression coverage for per-axis direct-feedforward selection):

- Added `testGpsRescueUsesPidLevelDirectFF` — verifies GPS_RESCUE_MODE selects the same pure-P-term directFF path as ANGLE_MODE via `pidLevelActive[axis]`, rather than falling back to the pre-#1257 D/F-contaminated path. Confirmed the test fails against the pre-fix `ANGLE_MODE`/`HORIZON_MODE`-only gate (temporarily reverted locally, ~7.8 unit divergence vs ~0.03 residual noise with the fix in place).
- Added `testNfeRaceModeOnlyEngagesPidLevelOnRoll` — verifies NFE_RACE_MODE's pitch exclusion from the `pidLevel()` call site still produces plain rate/acro output with no cross-axis state leakage.

Both in `src/test/unit/pid_unittest.cc`, commit efc5e1cd19. `make test` passes (12/12 in pid_unittest).

nerdCopter:
@coderabbitAI review . is this ready for merge?

nerdCopter:
AI Generated comment

Addressed local CodeRabbit CLI finding (major): testNfeRaceModeOnlyEngagesPidLevelOnRoll ran with the default PID_LEVEL_LOW.I=0, which zeroes directFF entirely - the roll-vs-rate comparison only reflected the angle-mode P/D/F formula switch, not the directFF path this PR actually changes.

Generalized primeAndStepUnderFlightMode() to accept a flight-mode bitmask (0 = plain rate/acro baseline) and reused it for both the NFE and baseline runs in this test, matching the non-zero PID_LEVEL_LOW.I setup already used in testGpsRescueUsesPidLevelDirectFF. Commit 54fbb4ab55.

`coderabbit review --agent --base upstream/master -c AGENTS.md` (local CLI) on the full branch diff: 0 findings. `make test`: 12/12 pass in pid_unittest.

nerdCopter:
@coderabbitai review

nerdCopter:
@coderabbitAI, AI generated comment

Please analyse HEAD at 54fbb4a for completeness, thoroughness, and codebase/merge safety — specifically src/main/flight/pid.c and src/test/unit/pid_unittest.cc. Confirm no regressions against upstream/master and that the new pidLevelActive per-axis gating is sound.

Issue #1254, opened 2026-06-10
feat(blackbox): output field masking via CLI (BF 4.5 parity)


nerdCopter:
AI Generated issue-ticket

Betaflight 4.5-maintenance implements a \`blackbox_disable_*\` CLI setting system (\`BLACKBOX_DISABLE\` bitmask) allowing users to selectively suppress individual blackbox fields (motors, gyro, ACC, debug, setpoint, RC commands, etc.) without disabling blackbox entirely. EmuFlight has no equivalent.

**Reference files in BF 4.5-maintenance:**
- \`src/main/blackbox/blackbox.c\` — mask checks before each field write
- \`src/main/cli/settings.c\` — \`blackbox_disable_*\` CLI entries
- \`src/main/msp/msp.c\` — MSP handler (see note below)

**Scope for this PR:** CLI + PG storage + blackbox write-path mask checks only.
MSP handler for \`blackbox_disable_*\` is deferred to the MSP freeze PR — see #1250.

Issue #1250, opened 2026-06-09
feat(msp): MSP 1.55.0 freeze tracker — all pending MSP changes for EmuFlight 0.5.0


nerdCopter:
AI Generated issue-ticket

## MSP 1.55.0 Freeze Tracker — EmuFlight 0.5.0

MSP protocol changes are deferred until feature-freeze. Each feature PR implements CLI/PG storage now; all MSP handlers land in a single pre-release PR targeting API version `1.55.0`.

Companion tracker to https://github.com/emuflight/EmuConfigurator/issues/654

---

## Pending MSP work

| Feature | PR / Issue | MSP work deferred |
|---|---|---|
| Rate profile count in `MSP_STATUS_EX` | #1250 (original scope) | `MSP_STATUS_EX` field |
| HDZero VTX configurator read/set + tab rename | #1252 | VTX band/channel/power/pit MSP commands; configurator UI |
| Blackbox output field masking | #1254 | `blackbox_disable_*` MSP handler |
| Altitude limit warning | #1260 | `MSP_POSITION_CONFIG` / `MSP_SET_POSITION_CONFIG` |

## Out of scope

- Motor output reordering (#1255) — CLI-only; EmuFlight does not implement MSP2; no MSP planned
- HDZero firmware-to-VTX serial forwarding in #1252 — does **not** require new EF MSP commands; not subject to freeze

---

### #1250 — Rate profile count (`MSP_STATUS_EX`)

Rate profile count is currently hardcoded to 6 in EmuConfigurator. PR #1249 introduced a build-time `EXTRA_FLAGS` override, but the Configurator cannot reflect overridden counts without MSP support. BF already transmits `CONTROL_RATE_PROFILE_COUNT` in `MSP_STATUS_EX` (added at API 1.46); EF never ported that field.

**Firmware:**
- Append `CONTROL_RATE_PROFILE_COUNT` (u8) to `MSP_STATUS_EX` response after arming disable flags
- Bump `API_VERSION_MINOR` to 55
- Decide default: 3 (small-flash parity) or 4 (BF 4.5-maintenance parity)

**EmuConfigurator:**
- Add `CONFIG.numRateProfiles` default to `fc.js`
- Read from MSP in `MSPHelper.js` when `apiVersion >= 1.55.0`
- Use in `loadRateProfilesList()` (`pid_tuning.js`) instead of hardcoded 6

[...]

PR #1331, opened 2026-07-24
feat(blackbox): log unfiltered/raw gyro data like Betaflight

### LINKED ISSUE #1251: feat(blackbox): log unfiltered/raw gyro data like Betaflight
AI Generated issue-ticket

EmuFlight currently only logs \`gyroADCf\` (the fully-filtered gyro) in \`src/main/blackbox/blackbox.c:914\`. Betaflight 4.5-maintenance additionally logs the pre-filter raw gyro sample as a separate blackbox field, giving tuners visibility into raw sensor noise before any filtering is applied.

**Request:** Add an unfiltered gyro blackbox field mirroring the BF 4.5-maintenance implementation.

**Reference:** \`betaflight/src/main/blackbox/blackbox.c\` — raw gyro field alongside \`gyroADC\`.

No MSP changes required.

nerdCopter:
AI Generated pull-request

Closes #1251.

## Summary

Adds an unfiltered/raw gyro blackbox field, `gyroUnfilt`, matching Betaflight 4.5-maintenance's field name, encoding, and gating exactly: **always logged by default, no separate enable flag** (BF has no such flag either — `CONDITION(GYROUNFILT)` in BF resolves purely to the field-mask check, nothing else). EF currently only logs `gyroADC` (post-filter, despite the legacy field name) in the blackbox log; this adds the pre-filter sample as an unconditional second field so tuners can inspect raw sensor noise before any filtering is applied, matching BF's default behavior for every non-IMUF9001 target.

This is a logging-only addition. No behavior change to flight control — the new field is captured alongside the existing filtered gyro capture in `loadMainState()` and does not alter any control-loop computation.

## Exact identifiers introduced (for #1254 to reference)

- Blackbox field name (as it appears in the log header): **`gyroUnfilt`** (3 entries: axis 0/1/2, exactly matching BF's field name)
- Sensor-side struct member: **`gyro.gyroADC[axis]`** (raw/unfiltered, `src/main/sensors/gyro.h`) — distinct from the existing `gyro.gyroADCf[axis]` (filtered)
- Blackbox capture struct member: **`blackboxMainState_t.gyroUnfilt[axis]`** (`src/main/blackbox/blackbox.c`)

**Default state: always on, unconditional (`CONDITION(ALWAYS)`) — matches BF exactly, matches how EF's own `gyroADC` field is already declared.** No PG storage, no CLI setting for this PR — issue #1254 (PR #1332)'s field-masking mechanism is the only intended way to disable it, exactly like BF.

## What changed

- **`src/main/sensors/gyro.h`**: added `float gyroADC[XYZ_AXIS_COUNT]` to `gyro_t`. EF's global `gyro_t` previously had no unfiltered member at all (only `gyroADCf`) — per-device raw values already existed on `gyroDev_t` but were never propagated to a global unfiltered field. This is sensor-side plumbing that did not previously exist in EF, unlike BF where both raw and filtered were already globally available.
- **`src/main/sensors/gyro.c`**: populate `gyro.gyroADC[...]` in all four `gyroUpdate()` combine branches (single-gyro, dual-gyro-1, dual-gyro-2, dual-gyro-averaged), mirroring the existing `gyroADCf` propagation lines exactly. **Also**: on `USE_GYRO_IMUF9001` targets (e.g. HELIOSPRING), the coprocessor protocol exposes only one filtered sample per axis — no separate raw channel exists at all. `gyroDev.gyroADC[]` mirrors `gyroDev.gyroADCf[]` on this platform (the same underlying value `DEBUG_GYRO_SCALED` reads), rather than staying at a permanently stale/zero value. Documented in-line. See "Real-hardware verification" below for how closely this tracks `DEBUG_GYRO_SCALED` in practice.
- **`src/main/blackbox/blackbox.c`**:
  - New `gyroUnfilt` field-definition table rows (3 axes), `CONDITION(ALWAYS)` — same as the existing `gyroADC` (filtered) field, same encoding/predictor scheme: `SIGNED_VB` / `PREDICT(0)` for I-frames, `SIGNED_VB` / `PREDICT(AVERAGE_2)` for P-frames.
  - `blackboxMainState_t.gyroUnfilt[XYZ_AXIS_COUNT]` history struct member.
  - I-frame and P-frame writes are unconditional, matching `gyroADC`'s own write calls immediately above them.
  - Capture in `loadMainState()`: `blackboxCurrent->gyroUnfilt[i] = lrintf(gyro.gyroADC[i]);`

No PG/CLI changes — `blackboxConfig_t` and `PG_BLACKBOX_CONFIG` are unchanged from `upstream/master` (no version bump needed).

## Field-masking (#1254) relationship

BF gates `gyroUnfilt` purely via its `fields_disabled_mask` bitmask (`FLIGHT_LOG_FIELD_SELECT_GYROUNFILT` + `blackbox_disable_gyrounfilt`) — verified directly against BF 4.5-maintenance source: `case CONDITION(GYROUNFILT): return isFieldEnabled(FIELD_SELECT(GYROUNFILT));`, nothing else. EF doesn't have that masking mechanism yet — **issue #1254 (PR #1332)** ports it as a separate, already-scoped follow-up. Once both branches merge, `gyroUnfilt`'s field-def rows need the same `CONDITION(ALWAYS)` → maskable-condition migration PR #1332 already performs for `PID`/`RC_COMMANDS`/`SETPOINT`/`GYRO`, with the condition body being BF's own exact snippet above. Until then, this PR stands alone and gyroUnfilt is simply always on, matching BF's actual default behavior on non-IMUF9001 hardware.

**Update 2026-07-30**: this follow-up landed — `gyroUnfilt` is now masked via `blackbox_disable_gyrounfilt` (part of merging `upstream/master` into `feat/blackbox-field-masking`, PR #1332).

## MSP Freeze Policy

MSP deferred -> #1250. No MSP command or response was added or modified.

## Log-size impact

Unconditional: **adds 3 `int16` fields per frame** (`gyroUnfilt[0..2]`), `SIGNED_VB` (zigzag + 7-bit variable-byte) encoded — the same encoding class already used for the existing filtered `gyroADC` field. Analytically (1 byte for |value| <= 63, 2 bytes for |value| <= 8191, 3 bytes beyond that): I-frames add roughly 3-9 bytes total across 3 axes depending on gyro rate magnitude; P-frames (average-of-2 predictor, residual-encoded) typically add less per axis but can run somewhat higher than the already-filtered `gyroADC` field's residual, since raw samples carry unfiltered sensor noise the predictor can't smooth away. This is an analytical estimate, not a measured flight-log capture — no SITL/hardware flight-log session was run in this environment to produce actual byte-per-frame numbers. This log-size cost is now unconditional (applies to every user, matching BF), not opt-in.

## Build/test verification

- `make clean && make test`: all 39 unit test suites passed, 0 failures, including `test_blackbox_unittest` and `test_blackbox_encoding_unittest`.
- Built with `CCACHE_DISABLE=1` (warning gate): `HELIOSPRING`, `FOXEERF722V4`, `STELLARH7DEV`, `APEXF7` (the 4 real-aircraft hardware targets) plus `TUNERCF405`, `SKYSTARSF405AIO` (bench targets) — 6 succeeded, 0 failed, zero compiler warnings.
- Not hardware-verified in flight at PR submission time — see "Real-hardware verification" below for post-merge results.

## Real-hardware verification (added 2026-07-30)

`gyroUnfilt` vs `debug_mode=6` (`DEBUG_GYRO_SCALED`) compared row-by-row on two real captures:

- **Non-IMUF9001 (FOXEERF722V4)**: 123,328/123,328 rows exact match across all 3 axes — byte-identical.
- **IMUF9001 (HELIOSPRING)**: not byte-identical. A hand-shake capture (build `d1f9388`) shows small mismatches — 0.45% of rows (roll), 2.05% (pitch), 2.98% (yaw), all within ±1-6 raw units on values in the hundreds/thousands. Root cause: `DEBUG_GYRO_SCALED` is set from two separate call sites in `gyroUpdateSensor()` for this platform (an early unconditional one, and a second one inside `filterGyroDebug()`), with the `gyroUnfilt` mirror landing between them. If a new IMUF9001 coprocessor sample arrives over SPI/DMA in that window — more likely at higher angular rates — the two debug-set calls and the mirror can observe slightly different underlying samples. Mismatch rate scaling with axis (roll < pitch < yaw) in this capture is consistent with a timing race rather than a logic error. Not a functional concern for this PR (both fields still report real, current gyro data on this platform, matching within a few LSB), but the two fields should not be assumed byte-identical on IMUF9001 targets specifically.

## Revision history

- Initial version made this field opt-in via a dedicated `blackbox_record_gyro_unfilt` CLI boolean, default OFF — diverged from BF, which has no such flag and logs unconditionally by default. Corrected per maintainer feedback to match BF exactly (`CONDITION(ALWAYS)`, no PG/CLI surface at all).
- CodeRabbit review (GitHub bot) found a real bug: `gyroDev.gyroADC[]` was never populated on `USE_GYRO_IMUF9001` targets. Fixed — see "What changed" above.

- Q&A review (2026-07-29) found that commit `ce1b4a16fc`'s IMUF9001 mirror-fix, needed for `gyroUnfilt` to work on HELIOSPRING, shared a variable with a pre-existing debug line (`gyro_filter_impl.h:22`, `DEBUG_SET(DEBUG_KALMAN, 0, ...)`, unrelated to blackbox, predates this branch). This is a regression introduced by this PR's own commit, confined to this branch only (`master` is unaffected — that debug line reads a variable that stays zero-init on `master`'s IMUF9001 path). Fixed directly by guarding the line with `#ifndef USE_GYRO_IMUF9001`, restoring the exact value (`0`) IMUF9001 targets have on `master` today. `make test`: 39/39, no regressions.
- 2026-07-30: real HELIOSPRING hand-shake capture — see "Real-hardware verification" above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved blackbox logging reliability by ensuring unfiltered gyro data is consistently recorded.
  * Corrected gyro data scaling and synchronization across supported sensor configurations.
  * Prevented stale or misleading debug values when using IMUF9001 gyro hardware.
* **Refactor**
  * Simplified unfiltered gyro logging so it is always included in blackbox records.


nerdCopter:
AI generated comment

@CodeRabbitAI review

Local `coderabbit review --agent` could not complete during implementation — hit persistent free-tier rate-limiting across an attempt and a retry. Change adds an opt-in (default OFF) `gyroUnfilt` blackbox field via `blackbox_record_gyro_unfilt` CLI setting, plus new sensor-side raw-gyro plumbing (`gyro.gyroADC[axis]`) that didn't previously exist in this codebase. No MSP surface touched. `make test` 39/39, 6 target builds (4 real-aircraft + 2 bench) all clean. Requesting a full review since no automated review has run on this PR yet.

nerdCopter:
AI generated comment

@CodeRabbitAI Fixed in commit ce1b4a16fc. Confirmed the gap: the USE_GYRO_IMUF9001 branch of gyroUpdateSensor() only updated gyroADCf[], never gyroADC[] — verified via the IMUF9001 driver (accgyro_mpu.c) that the coprocessor protocol exposes only one filtered sample per axis, no separate unfiltered channel. Added a mirror assignment (gyroADC[axis] = gyroADCf[axis]) in the IMUF9001 branch so gyroUnfilt reflects current data instead of a permanently stale/zero value on HELIOSPRING and other IMUF9001 targets. Documented in-line that this means gyroUnfilt == gyroADC (filtered) specifically on this platform, since no truly raw sample is available. make test (39/39) and HELIOSPRING build both verified clean after the fix.

PR #1348, opened 2026-08-01
Activate spiInitBusDMA() for F4/F7 targets (partial #1340)

### LINKED ISSUE #1340: SPI DMA never activated on any target: spiInitBusDMA() ported from Betaflight but never called
AI Generated [issue-ticket]

## Summary

`spiInitBusDMA()` (the only function that sets `bus->useDMA = true`, enabling the DMA branch in `spiSequence()`/`spiSequenceStart()`) has **zero call sites anywhere in EmuFlight, for any target, confirmed on master**. Every SPI transfer on every board — flash chips, OSD, RX modules, gyros — has always used blocking-polling (`spiTransfer()`), never real DMA, since this generic bus infrastructure was introduced. This is not IMUF9001-specific; it affects the entire fleet.

Discovered while root-causing a 16kHz CPU-overload issue on HELIOSPRING (see #1281, #1143, PR #1283) — activating `spiInitBusDMA()` broke real hardware twice for reasons that needed hardware-in-the-loop debugging to characterize. Investigating further surfaced that this was never a HELIOSPRING-specific gap at all.

## Confirmed against Betaflight 4.5-maintenance

Checked the upstream reference implementation directly (`~/SYNC/nerdCopter-GIT/worktrees/4.5-maintenance`, `betaflight/betaflight` 4.5-maintenance branch):

- `spiInitBusDMA()` is defined and called in `src/main/fc/init.c`, gated by two mutually-exclusive flags: `USE_SPI_DMA_ENABLE_EARLY` (called right after `setArmingDisabled(ARMING_DISABLED_BOOT_GRACE_TIME)`, **before** `motorPostInit()`/`motorEnable()`) or `USE_SPI_DMA_ENABLE_LATE` (called after motor init, only if EARLY isn't defined).
- **Both `src/main/target/STM32F405/target.h` and `src/main/target/STM32F7X2/target.h`** (HELIOSPRING's and STRIXF10's exact chip families) define `USE_SPI_DMA_ENABLE_EARLY`. This is a per-MCU-family setting in Betaflight, applied generically to every board of that chip, not a per-board opt-in.

## Why this isn't a simple copy-paste for EmuFlight

EmuFlight's `fc_init.c` has a materially different boot structure:

- `motorDevInit()` is called at `fc_init.c:334` — **before** `setArmingDisabled(ARMING_DISABLED_BOOT_GRACE_TIME)` at `fc_init.c:673`. There is no separate `motorPostInit()`/`motorEnable()` step later in the sequence to place a DMA call before, unlike Betaflight's split. EmuFlight only has the one early motor init call.
- EmuFlight has no equivalent of Betaflight's shared per-MCU-family `target/STM32F405/target.h` / `target/STM32F7X2/target.h` headers that get automatically applied to every board of that chip. `src/main/target/STM32F7X2/` in EmuFlight is a standalone board target (own `target.c`/`.h`/`.mk`), not a shared family header included by other F7X2 boards.
- Consequence: even if `spiInitBusDMA()` is added to EmuFlight's `fc_init.c`, motors will have already claimed their DMA1 stream(s) well before any reasonable insertion point, unlike Betaflight's EARLY ordering where SPI gets first pick. This was directly observed while investigating #1281/#1143: SPI2 (Ext SPI) and SPI3 (Flash/OSD) both lose their preferred DMA1 stream option to motor conflicts (Motor1→Stream4, Motor2→Stream5, Motor3→Stream7 all claimed at `fc_init.c:334`, before any spiInitBusDMA() insertion point) and silently fall back to polling regardless of where the call goes.

## What needs deciding, not assumed

- Whether EmuFlight's `motorDevInit()` can/should be split (device-level setup vs. DMA-stream-claiming) to create an equivalent "before motor DMA claims" insertion point, matching Betaflight's EARLY ordering intent
- Whether to introduce a shared per-MCU-family flag mechanism (matching Betaflight's `target/STM32F405/target.h` pattern) or handle this via `make/mcu/STM32F4.mk`-style build-system selection (EmuFlight already uses this pattern for selecting `bus_spi_stdperiph.c` vs `bus_spi_ll.c` per MCU)
- Full audit of DMA1/DMA2 stream conflicts across all F405/F7X2 targets once ordering is decided, not just the three IMUF9001 targets already checked

## Scope and relationship to other issues

- **Not IMUF9001-specific.** Affects every SPI-DMA-capable target in EmuFlight.
- **#1281** (32kHz stepping/MSP starvation on HELIOSPRING) and **PR #1283** (IMUF9001 driver migration) both encountered this gap, but PR #1283's fix does not depend on it — it uses a dedicated, bus-specific DMA implementation for SPI1 that bypasses `spiInitBusDMA()` entirely, so it is not blocked by this issue.
- User decision: **this codebase-wide gap should be resolved before any future IMUF9001-specific work builds on the generic DMA path**, to avoid layering board-specific fixes on top of an unresolved shared-infrastructure gap.

nerdCopter:
AI Generated pull-request

## Scope vs. issue #1340 — partial resolution, not full closure

Issue #1340's third comment declares two things in scope: (a) generic `spiInitBusDMA()`
activation fleet-wide, explicitly including HELIOSPRING's own SPI2 (Ext SPI) and SPI3
(Flash/OSD), and (b) confirms SPI1 on IMUF9001 boards stays permanently on PR #1283's
dedicated path. This PR delivers (a) only for non-`USE_GYRO_IMUF9001` targets — it
excludes HELIOSPRING/STRIXF10/MODE2FLUX **entirely** (not just their SPI1), deferring their
SPI2/SPI3 to a follow-up despite the issue calling that in scope. H7 is also out of scope
here (no H7 real-aircraft testing done in this PR). This does **not** close #1340 outright;
it resolves the F4/F7-non-IMUF9001 portion. A follow-up PR is needed for IMUF9001-board
SPI2/SPI3 and H7 before #1340 can actually close.

## Issue #1340's own root-cause theory was incorrect

The issue attributes the activation gap to EmuFlight lacking BF's
`motorPostInit()`/`motorEnable()` split (BF's `USE_SPI_DMA_ENABLE_EARLY`/`_LATE`
ordering). Direct comparison shows that's not the real blocker: DSHOT/PWM claims its DMA
inside `motorDevInit()` in both codebases, before either ordering's call site; BF's
EARLY/LATE distinction only matters for DSHOT bitbang, which EmuFlight doesn't implement
at all. The actual reason SPI must claim its DMA streams last: EmuFlight's `dmaInit()` (26
call sites — motors, UART, ADC, LED strip, transponder, SD card) overwrites DMA stream
ownership unconditionally, unlike BF's ownership-checked `dmaAllocate()`. No
`motorDevInit()` split was needed or attempted. The call site is placed after every
`dmaInit()` caller and before `fcTasksInit()` (gyro mode latches on first scheduled read).

## The actual blocker: a second, unrelated, pre-existing bug

Activating `spiInitBusDMA()` exposed a latent bug with nothing to do with motor-init
ordering: `spiInit()` (`bus_spi.c`) was setting `bus->busType = BUS_TYPE_SPI` at boot-time
peripheral bring-up, before any device attaches. `spiSetBusInstance()`'s "first device on
this bus" branch — the only place `bus->initTx`/`initRx` get assigned — gates on
`busType != BUS_TYPE_SPI`, so with `spiInit()` pre-setting it, that branch never fires, for
any SPI bus, on any target. `initTx`/`initRx` have been NULL fleet-wide since the Stage M
refactor; invisible because nothing dereferenced them until this PR's `spiInitBusDMA()`
call did, where it NULL-pointer hard-faults inside `spiInternalResetDescriptors()` (F7 LL
backend).

Confirmed via direct diff this is an EF-only divergence: BF 4.5-maintenance's `spiInit()`
calls only `spiInitDevice()` and never touches `busDevice_t` at all. EF's extra block was
also fully redundant even before being wrong — `spiSetBusInstance()`'s own first-
registration branch already sets the same fields correctly. F7 targets (FOXEERF722V4,
PYRODRONEF7) hard-faulted immediately and reliably on this bug (confirmed via LED-based
hardware bisection, no debugger available). F4 (SKYSTARSF405AIO) silently tolerated the
same NULL-pointer write as undefined behavior without crashing — meaning it was never
actually working correctly either, just not visibly broken. Fixed by removing the
redundant block from `spiInit()`, matching BF exactly (`bus_spi.c`, net -11 lines).

## Test plan
- [x] `make test` — 39/39 unit test suites pass (before and after the `spiInit()` fix)
- [x] Build clean, zero warnings: HELIOSPRING, TUNERCF405, SKYSTARSF405AIO, PYRODRONEF7,
      FOXEERF722V4, FOXEERF405, APEXF7, TMOTORF7, MATEKF411, NBDHMBF4PRO, WORMFC,
      ALIENWHOOPF7, CRAZYFLIE2, STRIXF10, MODE2FLUX
- [x] Link-map cross-reference confirms `spiInitBusDMA()` is dead-code-eliminated on
      HELIOSPRING/STRIXF10/MODE2FLUX and live on all other checked targets
- [x] Bench hardware verification post-fix: FOXEERF722V4, PYRODRONEF7, SKYSTARSF405AIO,
      TMOTORF7 all boot cleanly with correct live gyro/accel data (Configurator sensors
      tab); `dma` CLI output shows expected new `SPI_SDO`/`SPI_SDI` allocations with zero
      collision against existing motor/ADC DMA usage
- [x] Master-vs-branch sensor comparison (same physical board, same orientation): matching
      accelerometer magnitude (~1g) and gyro bias (single-digit deg/s, normal MEMS
      variance) — no scale, offset, or data-corruption difference between the polling and
      DMA-driven read paths
- [x] FOXEERF722V4 real-aircraft hover test flown on this branch (commit `4684ffd7`),
      line-of-sight, non-acrobatic. Blackbox confirms: 0 failsafe events, motor range
      fully utilized with no stuck values, unfiltered gyro (`gyroUnfilt`) shows normal
      sensor behavior with no rail-pinning or repeated/stuck samples, and blackbox
      loop-logging gap rate (1.92%, max gap 205) is identical to an unrelated prior flight
      on the same aircraft — confirming that characteristic predates this PR

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SPI bus initialization and DMA activation during flight controller startup.
  * Enabled late SPI DMA setup on supported STM32F4 and STM32F7 hardware.
  * Added board-specific handling for IMUF9001 targets to prevent conflicting SPI DMA configuration.
  * Improved initialization reliability and compatibility for SPI-connected components across supported flight controller hardware.

### DIFF
--- a/src/main/drivers/bus_spi.c
+++ b/src/main/drivers/bus_spi.c
@@ -142,17 +142,7 @@ bool spiInit(SPIDevice device) {
 #endif
         break;
     }
-    if (ok) {
-        busDevice_t *bus = spiBusByDevice(device);
-        if (bus) {
-            bus->busType = BUS_TYPE_SPI;
-            bus->busType_u.spi.instance = spiDevice[device].dev;
-#if defined(USE_HAL_DRIVER)
-            bus->busType_u.spi.handle = &spiDevice[device].hspi;
-#endif
-            bus->curSegment = (busSegment_t *)BUS_SPI_FREE;
-        }
-    }
+    // busDevice_t bookkeeping belongs to spiSetBusInstance()'s first-registration branch, not here.
     return ok;
 }
 
--- a/src/main/drivers/bus_spi.h
+++ b/src/main/drivers/bus_spi.h
@@ -119,8 +119,8 @@ SPIDevice spiDeviceByInstance(SPI_TypeDef *instance);
 SPI_TypeDef *spiInstanceByDevice(SPIDevice device);
 
 // Bus-abstraction accessor. Returns a pointer to the shared busDevice_t that
-// represents the given SPI peripheral (populated by spiInit()). Returns NULL
-// for invalid devices.
+// represents the given SPI peripheral (populated by spiSetBusInstance() on
+// first device registration). Returns NULL for invalid devices.
 busDevice_t *spiBusByDevice(SPIDevice device);
 
 // Mark an extDevice_t as belonging to an SPI bus, using the 1-based CLI device id.
--- a/src/main/fc/fc_init.c
+++ b/src/main/fc/fc_init.c
@@ -671,6 +671,12 @@ void init(void) {
     latchActiveFeatures();
     pwmEnableMotors();
     setArmingDisabled(ARMING_DISABLED_BOOT_GRACE_TIME);
+#if defined(USE_SPI) && defined(USE_SPI_DMA_ENABLE_LATE)
+    // Must run after every dmaInit() caller above (motors, ADC, LED strip,
+    // transponder, SD card) and before fcTasksInit(), which triggers the first
+    // gyro read that latches GYRO_EXTI_INT_DMA vs GYRO_EXTI_INT (accgyro_mpu.c).
+    spiInitBusDMA();
+#endif
     fcTasksInit();
     systemState |= SYSTEM_STATE_READY;
 }
--- a/src/main/fc/fc_init.c
+++ b/src/main/fc/fc_init.c
@@ -672,9 +672,7 @@ void init(void) {
     pwmEnableMotors();
     setArmingDisabled(ARMING_DISABLED_BOOT_GRACE_TIME);
 #if defined(USE_SPI) && defined(USE_SPI_DMA_ENABLE_LATE)
-    // Must run after every dmaInit() caller above (motors, ADC, LED strip,
-    // transponder, SD card) and before fcTasksInit(), which triggers the first
-    // gyro read that latches GYRO_EXTI_INT_DMA vs GYRO_EXTI_INT (accgyro_mpu.c).
+    // Must run after every dmaInit() caller above and before fcTasksInit()'s first gyro read.
     spiInitBusDMA();
 #endif
     fcTasksInit();
--- a/src/main/target/common_fc_post.h
+++ b/src/main/target/common_fc_post.h
@@ -30,8 +30,7 @@
 #undef USE_VTX_TRAMP
 #endif
 
-// IMUF9001 boards drive SPI1 DMA directly (accgyro_mpu.c); the generic path
-// must not also claim those streams.
+// IMUF9001 boards own SPI1 DMA directly (accgyro_mpu.c), not via the generic path
 #if defined(USE_GYRO_IMUF9001)
 #undef USE_SPI_DMA_ENABLE_LATE
 #endif
--- a/src/main/target/common_fc_post.h
+++ b/src/main/target/common_fc_post.h
@@ -30,6 +30,12 @@
 #undef USE_VTX_TRAMP
 #endif
 
+// IMUF9001 boards drive SPI1 DMA directly (accgyro_mpu.c); the generic path
+// must not also claim those streams.
+#if defined(USE_GYRO_IMUF9001)
+#undef USE_SPI_DMA_ENABLE_LATE
+#endif
+
 #ifndef USE_DSHOT
 #undef USE_ESC_SENSOR
 #endif
--- a/src/main/target/common_fc_pre.h
+++ b/src/main/target/common_fc_pre.h
@@ -59,6 +59,11 @@
 #define USE_ADC_INTERNAL
 #define USE_USB_CDC_HID
 #define USE_USB_MSC
+// SPI DMA must claim its streams after every dmaInit() caller (motors, ADC, LED
+// strip, transponder, SD card) has already run — dmaInit() overwrites ownership
+// unconditionally, unlike dmaAllocate(). common_fc_post.h strips this for
+// USE_GYRO_IMUF9001 boards, which own SPI1 DMA directly.
+#define USE_SPI_DMA_ENABLE_LATE
 
 #if defined(STM32F40_41xxx) || defined(STM32F411xE)
 #define USE_OVERCLOCK
@@ -78,6 +83,7 @@
 #define USE_ADC_INTERNAL
 #define USE_USB_CDC_HID
 #define USE_USB_MSC
+#define USE_SPI_DMA_ENABLE_LATE
 #endif
 
 #ifdef STM32H7
--- a/src/main/target/common_fc_pre.h
+++ b/src/main/target/common_fc_pre.h
@@ -59,10 +59,7 @@
 #define USE_ADC_INTERNAL
 #define USE_USB_CDC_HID
 #define USE_USB_MSC
-// SPI DMA must claim its streams after every dmaInit() caller (motors, ADC, LED
-// strip, transponder, SD card) has already run — dmaInit() overwrites ownership
-// unconditionally, unlike dmaAllocate(). common_fc_post.h strips this for
-// USE_GYRO_IMUF9001 boards, which own SPI1 DMA directly.
+// dmaInit() (used elsewhere in fc_init.c) has no ownership check, unlike dmaAllocate(), so SPI must claim last.
 #define USE_SPI_DMA_ENABLE_LATE
 
 #if defined(STM32F40_41xxx) || defined(STM32F411xE)


nerdCopter:
@CodeRabbitAI please perform a full, thorough analysis of this pull request.

This is safety-critical flight-controller firmware (real aircraft). Two distinct changes
are stacked on this branch:

1. Activating `spiInitBusDMA()` for F4/F7 non-`USE_GYRO_IMUF9001` targets
   (`common_fc_pre.h`, `common_fc_post.h`, `fc_init.c`). Please scrutinize the
   ordering/collision-safety reasoning in the PR description against EmuFlight's
   `dmaInit()` (26 call sites) having no ownership check, unlike the ownership-checked
   `dmaAllocate()` this change relies on — and confirm the `USE_GYRO_IMUF9001` exclusion
   is complete and correctly gated for all three affected boards.

2. A deeper fix in `spiInit()` (`bus_spi.c`) removing a boot-time `busDevice_t`
   pre-marking block, restoring `spiSetBusInstance()`'s "first device on this bus"
   invariant. Please examine whether anything else, anywhere in the tree, relies on
   `spiInit()` having set `busType`/`busType_u.spi.instance`/`curSegment` before any device
   registers on a bus — that invariant is exactly what this fix changes, and I want an
   independent check that nothing else depended on the old (incorrect) behavior.

Please flag anything, however minor, including style, edge cases, or missed BF-parity
details.

nerdCopter:
@CodeRabbitAI review

nerdCopter:
@coderabbitAI , your review feedback will be resolved at earliest convenience. Meanwhile,  are ther any regrssions found in this PR? are there any concerns for unrelated SPI, DMA, UART, or any otherwise resources issues related to this PR-change

nerdCopter:

Addressed findings from review #4836066725:

**Stale doc comment — `src/main/drivers/bus_spi.h:122`** — Actionable
Fixed in commit `dd23de64c`: `spiBusByDevice()` comment said the descriptor is "populated by `spiInit()`"; corrected to name `spiSetBusInstance()`, which does the population since this branch's `spiInit()` change.

**Partial TX DMA allocation retains ownership — `src/main/fc/fc_init.c:674-677` (implementation in `src/main/drivers/bus_spi.c` `spiInitBusDMA()`)** — BF 4.5-maintenance parity
The final `else` branch that clears `bus->dmaTx`/`bus->dmaRx` without releasing the TX `dmaAllocate()` ownership when `USE_TX_IRQ_HANDLER` is undefined is byte-identical to BF 4.5-maintenance's `spiInitBusDMA()` at `src/main/drivers/bus_spi.c:701-705` (commit `b72789ee8`). This PR only wires the `spiInitBusDMA()` call for F4/F7 (`common_fc_pre.h`), and `USE_TX_IRQ_HANDLER` is unconditionally defined for `STM32F4`/`STM32F7` (`bus_spi.c:33-35`), so the leak path is unreachable for any target this PR activates. No code change made — fixing pre-existing shared logic here would diverge from BF without addressing a reachable defect in this PR's scope.

**Regression test coverage — `src/main/drivers/bus_spi.c:145` (x2), covering `spiSetBusInstance()` first/repeat registration and `spiInitBusDMA()` allocation fallback** — Out of scope for this PR
Confirmed there is no existing unit-test infrastructure that compiles `bus_spi.c` or the `dma_stm32f4xx.c`/`dma_stm32f7xx.c` drivers under `UNIT_TEST` — both call into CMSIS register access, RCC/NVIC/EXTI, and `bus_spi_impl.h` internals with no host-side mocks. The one comparable precedent, `dma_bounds_unittest.cc`, deliberately mirrors the guard logic locally rather than linking the real driver file for the same reason. Building a real mock layer for SPI/DMA register access is a separate, heavier-lift infrastructure task, not a same-PR addition. Will open a tracking issue for it if desired — let me know.

Full unit-test suite (`make clean_test && make test`) re-verified green after the doc-comment fix: 40/40 test binaries pass, 0 failures.

nerdCopter:
@CodeRabbitAI review

nerdCopter:
@CodeRabbitAI your formal review from 2026-08-01 (state: CHANGES_REQUESTED) predates this
PR's fixes and is still the current record — replying to it and re-pinging `review` haven't
changed that state, and I understand from your own reply that you don't re-review already-
reviewed commits.

Status of each finding from that review, detailed in
https://github.com/emuflight/EmuFlight/pull/1348#issuecomment-5178849155:
- Actionable (stale doc comment on `spiBusByDevice()`): fixed, commit `dd23de64c`.
- Partial TX-DMA-allocation-leak in `spiInitBusDMA()`'s final `else`: reclassified
  BF-parity — byte-identical to BF 4.5-maintenance `bus_spi.c:701-705`, and unreachable for
  any target this PR activates (`USE_TX_IRQ_HANDLER` is always defined for F4/F7).
- 2 test-coverage nitpicks: out of scope, no host-mock test infra exists for
  `bus_spi.c`/`dma_stm32f4xx.c`/`dma_stm32f7xx.c` under `UNIT_TEST` — tracked as issue
  #1350.

Given that, can you update your review to Approve, or otherwise clear the
`CHANGES_REQUESTED` state? If that's not something you can do from a comment, please say so
directly.

nerdCopter:
@CodeRabbitAI the earlier CHANGES_REQUESTED review has been dismissed (findings addressed,
see the consolidated reply above and https://github.com/emuflight/EmuFlight/pull/1348#issuecomment-5178849155).
Is this PR approvable in its current state? If so, please submit an approving review.

Issue #1373, opened 2026-08-10
H7 UART DMA activation (dmaopt is a real stream-allocation decision, not a safe default)


nerdCopter:
AI Generated issue-ticket

## Summary

PR #1370 (issue #1364) activates the UART DMA PG-config/reqmap machinery for F4/F7, with
every UART defaulting to `DMA_OPT_UNUSED` (off). H7 is deliberately excluded from that
PR's `dma_reqmap_mcu.c` UART entries entirely — this issue tracks completing it as a
follow-up, not a defect in that PR.

## Depends on #1374 — RESOLVED, #1374 merged (PR #1383, 2026-08-13)

~~EF's H7 `serialUART()` doesn't consume the dmaopt/PG-config/reqmap mechanism at all
today~~ — PR #1383 ported the shared mechanism into H7's `uartConfigureDma()`. This
issue's own scope (below) is now unblocked.

## Correction (2026-08-13): step 1 of the proposed fix was based on a false premise

**The original step 1 below is false and already complete.** `git blame` on
`dma_reqmap_mcu.c` shows `REQMAP_DIR(UART, 1..8, TX/RX)` plus the conditional
`UARTDEV_9`/`UARTDEV_10` entries were all added in PR #1194 (`6e50925c84`, 2026-05-31) —
the same commit that introduced the old `uartConfigureDma()` stub. The H7 reqmap table
for UART has been complete since day one of H7 support; it was simply never *consumed*,
because nothing called `dmaGetChannelSpecByPeripheral()` for UART on H7 until PR #1383.

`dmaGetChannelSpecByPeripheral()`'s H7 branch (`dma_reqmap_mcu.c` ~line 228) also doesn't
use `opt` to select between per-peripheral alternates the way F4/F7 do — it uses `opt`
purely as an index into a generic 16-slot DMA-stream pool
(`dmaChannelSpec[MAX_PERIPHERAL_DMA_OPTIONS]`) and calls `dmaSetupRequest()` to bind the
peripheral's DMAMUX request dynamically at whichever stream `opt` names. This mechanism
already works correctly for every `UARTDEV_1`-`10` with no additional reqmap-table work
needed.

**This issue's real remaining scope is only original step 2** — per-target, decide
whether a specific H7 board's UART should activate DMA. See below (unchanged).

## Why H7 needs a different default posture than F4/F7

On F4/F7, `dmaGetChannelSpecByPeripheral()`'s `opt` parameter selects between a small,
[...]

PR #1370, opened 2026-08-10
fix(uart): activate DMA for UART on F4/F7 targets

nerdCopter:
AI Generated pull-request

## Summary

- No F4/F7 target ever had UART DMA compiled in: `USE_UARTx_RX_DMA`/`TX_DMA` were only
  ever defined under `#ifdef STM32F1`, a platform with zero build targets.
- Adds `pg/serial_uart.h`/`.c` — a `serialUartConfig` PG-config array (`txDmaopt`/
  `rxDmaopt` per UART), mirroring BF 4.5-maintenance's model.
- Adds `DMA_PERIPH_UART_TX`/`RX` reqmap entries for F4/F7 in `dma_reqmap_mcu.c`.
- Rewires `serialUART()` in `serial_uart_stm32f4xx.c`/`f7xx.c` to resolve the DMA
  stream/channel via `dmaGetChannelSpecByPeripheral()` at every open, replacing a
  compile-time-only table that nothing had ever activated. `uartDmaClaim()` itself is
  unchanged.
- Retires the now-superseded compile-time DMA stream table (`uartHardware_t`'s
  `DMAChannel`/`rxDMAStream`/`txDMAStream` fields and all associated `#ifdef
  USE_UARTx_RX_DMA`/`TX_DMA` blocks) from the F4/F7 driver files.

**Fleet-wide default is `DMA_OPT_UNUSED` (off) for every F4/F7 UART**, matching BF
4.5-maintenance's own real-world default — zero targets in that tree define any
`UARTn_TX_DMA_OPT` macro. This closes the architectural gap (the PG-config and reqmap
machinery now exists and is correct) while staying dormant by default. Enabling DMA on a
specific UART for a specific target is a follow-up decision requiring a collision check
against that target's SPI (gyro/flash/OSD) DMA usage — not a safe fleet default.

An earlier version of this branch defaulted F4/F7 UARTs to DMA-active fleet-wide,
reasoning that each UART has only one silicon-valid stream option so there's no ambiguity
about which to pick. That reasoning ignored boot-order contention: `mspSerialInit()`/
`rxInit()`/`gpsInit()`/`telemetryInit()` all open UARTs before `spiInitBusDMA()` runs in
`fc_init.c`, so a UART defaulting to active could claim a stream SPI needed first. This
was caught on real hardware (SKYSTARSF405AIO, TMOTORF7 — USB enumerated, MSP timed out
with zero bytes received) and fixed by reverting to `DMA_OPT_UNUSED`, confirmed against
[...]

Issue #1372, opened 2026-08-10
CLI resource dmaopt subsystem for UART/SPI (no infrastructure exists)


nerdCopter:
AI Generated issue-ticket

## Summary

EF has no CLI infrastructure for setting a peripheral's `dmaopt` (DMA stream/channel
option) at runtime. Confirmed while implementing PR #1370 (issue #1364, UART DMA
activation): `interface/cli.c`'s `resource` command (`resourceTable`/
`cliResourceValue_t`, `cliResource()`) only handles IO-tag/pin reassignment. BF's
equivalent mechanism (`cli.c`'s `DEFW_OFS`/`DEFW` macro family, a separate
dmaopt-specific resource table feeding `dmaGetOwner`/`serialUartConfig`-style PG arrays)
does not exist in EF at all — it is not a variant of the existing pin-resource table, a
distinct subsystem.

## Why this matters

PR #1370 ships `serialUartConfig` (per-UART `txDmaopt`/`rxDmaopt` PG-config array) with
every F4/F7 UART defaulting to `DMA_OPT_UNUSED` — correct for a safe fleet default (see
that PR's history: a fleet-wide active default caused a real hardware regression), but
it means there is currently no way for a pilot or target maintainer to opt a specific
UART into DMA on a specific board without editing source and rebuilding. The same gap
applies to any other `dmaopt`-based PG array in EF (e.g. SPI, if it ever moves off its
current auto-probe `spiInitBusDMA()` model to a configurable one).

## Proposed fix

Port BF's `DEFW`/`DEFW_OFS` CLI resource-table macro family (`cli.c`) and wire
`UART_TX`/`UART_RX` entries against `serialUartConfig_t`, matching BF's own
`PG_SERIAL_UART_CONFIG` resource entries. Scope: CLI command only, no MSP handler — per
this project's MSP Freeze Policy (`AGENTS.md`), defer any MSP-side exposure to the
freeze-tracker issue.

## Scope

Independent of PR #1370's own scope (which correctly ships the PG-config storage without
this CLI layer). Not required to close #1364.

Issue #1364, opened 2026-08-05
No F4/F7 target has UART DMA compiled in at all -- port BF 4.5-maintenance's dynamic USE_DMA_SPEC model


nerdCopter:
AI Generated [issue-ticket]

## Summary

No UART on any F4/F7 EmuFlight target has DMA compiled in, for any UART number. Verified
directly: `common_fc_pre.h` defines `USE_UART1_RX_DMA`/`USE_UART1_TX_DMA` only under
`#ifdef STM32F1`; a fleet-wide search confirms zero targets anywhere define any
`USE_UARTx_RX_DMA`/`TX_DMA` flag for F4/F7. `serial_uart_stm32f4xx.c`/`serial_uart_stm32f7xx.c`
fully support UART DMA in their driver logic (confirmed via live debug trace on
FOXEERF722V4 — pin-matching and `uartDevmap[]` population both work correctly), but the
static, compile-time opt-in flags that would populate `hardware->rxDMAStream`/`txDMAStream`
are simply never set anywhere in the target tree.

## Why this isn't a per-target config bug — confirmed against BF's own reference config

Checked BF's current reference config for FOXEERF722V4
(`~/SYNC/nerdCopter-GIT/config/configs/FOXE/FOXEERF722V4/config.h`) directly: it has no
per-target UART DMA opt-in define either. That's because BF 4.5-maintenance doesn't use
static per-target flags for this at all — it resolves UART DMA dynamically via
`USE_DMA_SPEC`: a generic `dmaChannelSpec_t` lookup
(`dmaGetChannelSpecByPeripheral(DMA_PERIPH_UART_TX/RX, device, opt)`) combined with a
persistent, CLI-configurable `txDmaopt`/`rxDmaopt` field per serial port, resolved at
port-open time by `uartConfigureDma()` (`serial_uart.c:385-429` in BF 4.5-maintenance).
EF ported the older static-flag scaffolding during the Stage M `extDevice_t` refactor but
never activated it anywhere, and never followed BF's later move to the dynamic model.

Confirmed this is genuinely in BF **4.5-maintenance** (the project's actual reference
branch), not a master-only or speculative feature — this is an in-scope backport under
this project's evaluation criterion, same reasoning already applied to the `txIrq` port
(#1361).

## Confirmed not a live flight-safety gap, just a closed BF-parity gap

EF has flown on F4/F7 with zero UART DMA for its entire history — this hasn't been
[...]

PR #1383, opened 2026-08-11
fix(uart): port H7 DMA resolution to shared dmaopt mechanism

### LINKED ISSUE #1374: H7 UART DMA resolution is an EF-original stub, not a port of BF's dmaopt mechanism
AI Generated issue-ticket

## Summary

`src/main/drivers/serial_uart_stm32h7xx.c`'s `uartConfigureDma()` was introduced whole
in PR #1194 ("feat(h7): add STM32H7 MCU support", 2026-05-31, the single commit that
added the entire H7 port) as a simplified, EF-original function:

```c
void uartConfigureDma(uartDevice_t *uartdev)
{
    ...
    if (hardware->rxDMAResource) {
        s->rxDMAStream  = (DMA_Stream_TypeDef *)hardware->rxDMAResource;
        s->rxDMAChannel = hardware->rxDMAChannel;
    }
    ...
}
// comment: "Full DMA IRQ wiring deferred -- polled/IRQ UART is sufficient for initial
// bring-up."
```

It only copies static `UARTn_TX/RX_DMA_STREAM`-macro-fed fields from `uartHardware_t`
into the port struct. It never calls `serialUartConfig()` or
`dmaGetChannelSpecByPeripheral()`.

## What BF 4.5-maintenance actually does

Confirmed directly against BF 4.5-maintenance (found while scoping issue #1373, PR
#1370's UART DMA activation work): BF's H7 `serialUART()` calls the **shared**
`uartConfigureDma()` in `drivers/serial_uart.c`, which has an `#ifdef USE_DMA_SPEC`
branch -- `USE_DMA_SPEC` is defined for `STM32H7` in `platform_mcu.h`, so BF's H7
resolves UART DMA via `serialUartConfig(device)->txDmaopt`/`rxDmaopt` +
`dmaGetChannelSpecByPeripheral()`, the *identical* dmaopt/PG-config/reqmap mechanism
F4/F7 use. BF's H7 file also defines the same static `UARTn_TX/RX_DMA_STREAM` macros
EF's does, but they are dead code once `USE_DMA_SPEC` takes over -- the same
retired-static-table pattern PR #1370 found and removed for F4/F7.

EF's H7 never received this port. The gap is architectural, not a missing default or
config value.

## Proposed fix

Port BF's shared `uartConfigureDma()` `USE_DMA_SPEC` branch into EF's H7 code path,
replacing the current static-field-copy stub in `serial_uart_stm32h7xx.c`. This is a
prerequisite for issue #1373 (H7 UART DMA activation) -- that issue's "complete the
reqmap table + decide a default" scope only becomes meaningful once H7 actually
consumes the PG-config/reqmap mechanism it resolves against.

## Scope

Independent of PR #1370 (F4/F7-only, per issue #1364's own title). Blocks issue #1373.
Worth checking whether other H7 driver files from PR #1194 have similar
deliberately-deferred stubs, given the "initial bring-up" framing of the comment above
-- not investigated here, flagging as a possibility only.

nerdCopter:
AI Generated pull-request

## Summary

- `serial_uart_stm32h7xx.c`'s `uartConfigureDma()` was an EF-original stub since
  PR #1194 (H7 bring-up): it only copied static `UARTn_TX/RX_DMA_STREAM`-macro-fed
  fields into the port struct. No H7 target ever defined those macros, so the path
  was provably inert on every target, always.
- Now resolves DMA via `dmaGetChannelSpecByPeripheral()` + `serialUartConfig()`'s
  per-UART dmaopt, mirroring EF's own already-ported F4/F7 implementation
  (PR #1370) — the same `uartDmaClaim()` ownership-check pattern, and
  `dmaSetHandler()` registration for the TX-DMA-complete IRQ.
- `uartDmaIrqHandler()` was a no-op stub, unreachable because `dmaSetHandler()` was
  never called to register it. Now calls `HAL_DMA_IRQHandler()`, matching F7's
  `dmaIRQHandler()`.
- Removed `uartHardware_t`'s now-dead `rxDMAResource`/`txDMAResource`/
  `rxDMAChannel`/`txDMAChannel` fields and their NULL-fallback macros, since
  `uartConfigureDma()` no longer reads them — same retired-static-table cleanup
  already done for F4/F7 in PR #1370.
- Fixed `serialUART()`'s NVIC-enable gate to check both RX and TX DMA fallback
  (`!s->rxDMAStream || !s->txDMAStream`), matching F4/F7's stricter gate. The
  single-condition gate this replaced is BF's own current, shipped H7 behavior too
  (verified directly against BF 4.5-maintenance) — a real latent bug shared with
  BF, not a misread of BF's architecture. Not reachable on any current H7 target
  (dmaopt stays `DMA_OPT_UNUSED` everywhere, so both sides of the new OR are
  already true today) — confirmed by inspection, no re-test needed.

## No default-behavior change

`serialUartConfig()`'s reset function only applies a non-`DMA_OPT_UNUSED` default
overlay under `#if defined(STM32F4) || defined(STM32F7)`. Every H7 UART's
`txDmaopt`/`rxDmaopt` stays at the blanket default (`DMA_OPT_UNUSED`), so
`dmaGetChannelSpecByPeripheral()` short-circuits to `NULL` before touching any
table, exactly as the old stub always resolved to `NULL`. This is an architecture
port only — no H7 UART's dmaopt defaults to active as part of this change (that
decision belongs to a separate follow-up, issue #1373).

## Verification

- `make clean && make STELLARH7DEV FOXEERF722V4 FOXEERF405 SITL`: 4/4 succeeded, 0
  warnings/errors (`CCACHE_DISABLE=1`).
- Full bench sweep (`HELIOSPRING TUNERCF405 SKYSTARSF405AIO PYRODRONEF7
  FOXEERF722V4 FOXEERF405 APEXF7 TMOTORF7 SITL`): 9/9 succeeded, 0
  warnings/errors.
- `make clean_test && make test`: 42/42 binaries pass.
- **Hardware verified on STELLARH7DEV**: `dma`/`status`/`tasks`/`version`/`diff all`
  CLI output byte-identical vs. master (`0d05ac775`). MSP connectivity confirmed via
  the Configurator's Setup-tab-first connection flow on both builds. No flight test
  needed — the touched code path (`opt < 0` short-circuit) doesn't depend on
  RX/failsafe/serial-port configuration.
- Three independent review tools converged: local CodeRabbit (2 rounds — 1
  false-positive rejected, checked against this codebase's actual `dmaChannelSpec_t`
  struct), an independent `/code-review` pass, and opencode — found and fixed 3 real
  issues (the NVIC-gate bug above, and two stale comments claiming H7 doesn't read
  `serialUartConfig()`, now corrected).
- **CI fully green**: all 11 build-matrix groups, `test`×2, `sitl`, Codacy.

## Outstanding

- Found issue #1373's claim that the H7 reqmap table needs completing for
  `UARTDEV_1`-`9` is false (those entries have existed since PR #1194, just never
  consumed until this PR) — flagged separately, not corrected as part of this PR.
- The NVIC-gate bug fixed here is BF's own current, shipped H7 behavior too — worth
  a BF issue ticket, not filed as part of this PR.

Closes #1374.

### DIFF
--- a/src/main/drivers/serial_uart_impl.h
+++ b/src/main/drivers/serial_uart_impl.h
@@ -20,10 +20,6 @@
 
 #pragma once
 
-#if defined(STM32H7)
-#include "drivers/dma.h"
-#endif
-
 // Configuration constants
 
 #if defined(STM32F4)
@@ -133,14 +129,6 @@ typedef struct uartPinDef_s {
 typedef struct uartHardware_s {
     UARTDevice_e device;    // XXX Not required for full allocation
     USART_TypeDef* reg;
-#if defined(STM32H7)
-#ifdef USE_DMA
-    dmaResource_t *txDMAResource;
-    dmaResource_t *rxDMAResource;
-    uint32_t txDMAChannel;   // DMAMUX request ID (DMA_REQUEST_USARTx_TX)
-    uint32_t rxDMAChannel;   // DMAMUX request ID (DMA_REQUEST_USARTx_RX)
-#endif
-#endif
 #if defined(STM32H7)
     uartPinDef_t rxPins[UARTHARDWARE_MAX_PINS];
     uartPinDef_t txPins[UARTHARDWARE_MAX_PINS];
--- a/src/main/drivers/serial_uart_stm32h7xx.c
+++ b/src/main/drivers/serial_uart_stm32h7xx.c
@@ -430,7 +430,9 @@ uartPort_t *serialUART(UARTDevice_e device, uint32_t baudRate, portMode_e mode, 
     }
 
 #ifdef USE_DMA
-    if (!s->rxDMAStream)
+    // Either side can fall back to IRQ-driven mode independently (uartDmaClaim()
+    // conflict); enable the shared NVIC line whenever either does, matching F4/F7.
+    if (!s->rxDMAStream || !s->txDMAStream)
 #endif
     {
         HAL_NVIC_SetPriority(hardware->rxIrq, NVIC_PRIORITY_BASE(hardware->rxPriority), NVIC_PRIORITY_SUB(hardware->rxPriority));
--- a/src/main/drivers/serial_uart_stm32h7xx.c
+++ b/src/main/drivers/serial_uart_stm32h7xx.c
@@ -31,6 +31,7 @@
 
 #include "drivers/system.h"
 #include "drivers/dma.h"
+#include "drivers/dma_reqmap.h"
 #include "drivers/io.h"
 #include "drivers/nvic.h"
 #include "drivers/rcc.h"
@@ -38,6 +39,8 @@
 #include "drivers/serial.h"
 #include "drivers/serial_uart.h"
 #include "drivers/serial_uart_impl.h"
+
+#include "pg/serial_uart.h"
 
 // DMA-accessible UART buffers (H7: must be in AXI SRAM for DMA cache coherency)
 #ifdef USE_UART1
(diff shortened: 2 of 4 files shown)


nerdCopter:
AI Generated comment

Hardware verification complete on STELLARH7DEV (2026-08-11/12): flashed
master (`0d05ac775`, post-PR#1370) and this branch (`006bba942`), captured
`dma`/`status`/`tasks`/`version`/`diff all` CLI output on both, plus a
third capture from the board's prior-flashed firmware (`461bee6`) for
additional context.

Result: byte-identical between master and this branch --
- `dma` table: no UART/SERIAL DMA ownership entries on either (expected --
  dmaopt stays `DMA_OPT_UNUSED` on H7, this PR ports the resolution
  mechanism without changing any default).
- `status`: same CPU clock (480MHz), same GYRO/ACC (ICM42688P), same
  `CPU:1%`, same cycle time (125us), same GYRO rate (8000Hz).
- `tasks`: SERIAL/GYRO-PID timing matches within normal run-to-run
  variance.
- `diff all`: identical config dump (resources, serial, aux, PID/rate
  profiles) -- also matches the prior-flashed baseline capture, confirming
  no config drift across any of the three builds.

MSP connectivity confirmed on both via the Configurator's own connection
flow: the Setup tab (which requires working MSP) is always shown first on
connect, and the CLI tab was reached manually from there on both builds --
MSP was functioning before CLI was used in either case.

No flight test performed or required -- the code path this PR touches
(`dmaGetChannelSpecByPeripheral()` short-circuiting on `opt < 0`) doesn't
depend on flight configuration.

nerdCopter:

Local review found one finding, checked against the actual code and not
actioned -- it suggested using `rxDmaSpec->dmaMuxId` instead of
`rxDmaSpec->channel` for H7. Verified: `dma_reqmap.h`'s
`dmaChannelSpec_t` struct in this codebase has only a `.channel` field --
no `.dmaMuxId` field exists anywhere in EF (that's BF's AT32-specific
union member; EF uses one unified `.channel` field for every MCU).
Applying the suggested fix would be a compile error. For H7 specifically,
that field has always held the DMAMUX request ID (per the struct's own
pre-existing comment `// DMAMUX request ID (DMA_REQUEST_USARTx_RX)`) both
before and after this PR -- this change only altered how the value is
sourced (dynamic reqmap lookup vs. a static macro), not its meaning.

Three things worth an independent check on this PR:

1. **Default-value safety**: this PR is the direct sequel to PR #1370
   (issue #1364), which had a real regression where a fleet-wide
   DMA-active default raced against SPI's own boot-time DMA claim and
   broke MSP on real F4/F7 hardware. Please confirm H7's dmaopt genuinely
   stays `DMA_OPT_UNUSED` by design in the *current diff itself* -- not
   just asserted in the PR description -- and that no path in this change
   causes a UART DMA claim to happen where none existed before.

2. **SITL impact**: PR #1370 separately broke the SITL build (a new
   `pg/serial_uart.c` file lacked an MCU-exclusion entry in
   `make/mcu/SITL.mk`). This PR only touches
   `serial_uart_impl.h`/`serial_uart_stm32h7xx.c` (H7-specific driver
   files), so it shouldn't need any `SITL.mk` change -- please confirm
   directly against the diff rather than assuming the same class of gap
   doesn't apply here.

3. **Orphaned references**: the old static `UARTn_TX/RX_DMA_STREAM`
   macro consumers were removed from these two files. Please check for
   any dangling reference to those macros or the removed struct fields
   elsewhere in the codebase.

Hardware already verified and passing on STELLARH7DEV (see the earlier
comment on this PR) -- this review is for code correctness, not to
re-litigate that.

nerdCopter:
AI Generated comment

Fixed two findings from an independent /code-review pass on `006bba942`, committed `efb5867d7`:

1. **Real bug (not yet reachable, fixed pre-emptively)**: `serialUART()`'s NVIC-enable
   gate checked only `s->rxDMAStream`, not `s->txDMAStream` -- unlike F4/F7's reference
   gate (`!s->rxDMAChannel || !s->txDMAChannel`), which enables the shared IRQ line
   whenever *either* side falls back to IRQ-driven mode. Not reachable today: every H7
   UART's dmaopt stays `DMA_OPT_UNUSED`, so both sides always resolve NULL, making both
   halves of the new OR condition already independently true -- confirmed no behavior
   change for any current configuration. Becomes reachable once a future per-target
   override or CLI support (#1372/#1373) lets RX succeed while TX falls back (or vice
   versa): the old gate would silently leave the shared NVIC line disabled, and TX would
   silently stop working with no build/boot symptom.
2. **Stale comment**: `pg/serial_uart.c` claimed H7 "never reads this PG array for UART"
   -- true before this branch's own change, false after it. Corrected.

Local CodeRabbit's one finding on this branch (suggesting a `.dmaMuxId` field) was
checked and rejected as a false positive -- confirmed via `dma_reqmap.h` that
`dmaChannelSpec_t` has only a single `.channel` field for every MCU in this codebase; no
`.dmaMuxId` field exists (that's BF's AT32-specific union member).

Re-verified: STELLARH7DEV/FOXEERF722V4/SITL all rebuild clean. No hardware re-test
needed for the NVIC-gate fix -- confirmed by inspection it's a no-op under every current
H7 target's configuration, so the STELLARH7DEV verification already posted against
`006bba942` remains valid evidence for `efb5867d7` too.

nerdCopter:

Latest commit `1b83129fd`: comment-only fix. `common_defaults_post.h`
carried the same stale claim already fixed in `pg/serial_uart.c`
(`efb5867d7`) -- "H7 has its own separate, pre-existing
UARTn_{TX,RX}_DMA_STREAM mechanism ... and never reads this PG array for
UART", true before this branch's port, false after it. Found by
opencode's independent second-opinion review, which caught this second
copy of the same now-false claim after `efb5867d7`'s fix addressed only
the first one.

No functional change: confirmed identical STELLARH7DEV `.elf` size
before/after (`389570`/`14568`/`84668` text/data/bss, unchanged).

PR #1384, opened 2026-08-13
feat(cli): add dmaopt resource subsystem for UART DMA stream/channel selection

nerdCopter:
AI Generated pull-request

## Summary

Adds a CLI subsystem for getting/setting a UART's `dmaopt` (DMA stream/channel selection) at runtime, closing #1372. EF previously had no way to opt a UART into a specific DMA resource without editing source and rebuilding — PR #1370/#1383 shipped the `serialUartConfig_t.txDmaopt`/`rxDmaopt` PG fields with every UART defaulting to `DMA_OPT_UNUSED`, but nothing could write them.

Ports BF 4.5-maintenance's `dmaoptEntry_t`/`DEFW` macro family and `cliDmaopt()` (`cli.c` lines ~5295-5714) — a distinct subsystem from the existing pin-based `resource` command, per the issue's own scope note.

New syntax on the existing `dma` command:
- `dma` / `dma show` — unchanged (hardware-verified utilization display, deliberately **not** repointed to match BF's literal bare-arg mapping — that would silently change already flight-verified behavior as a side effect)
- `dma list` — new, dumps every configured dmaopt entry
- `dma <device> <index>` — get
- `dma <device> <index> list` — list available DMA options for that device/index
- `dma <device> <index> <option>|none` — set

Only `UART_TX`/`UART_RX` are wired up (`grep dmaopt` across `pg/*.c` found no SPI/ADC/etc. PG field to port against yet). Guarded to `STM32F4`/`F7`/`H7` — `pg/serial_uart.c` is `MCU_EXCLUDES`'d on SITL.

**MSP deferred → #1250** per this project's MSP freeze policy — CLI + PG storage only, no MSP command/response added.

## Review

Three review rounds, all fully addressed:
- Local CodeRabbit (`coderabbit review --agent`) found and fixed 2 real issues in `0ee510852`: a NULL-deref on `pgFind()`'s result in `dmaoptAddr()`, and an `int8_t` truncation bug where out-of-range CLI input (e.g. `300`) silently wrapped instead of being rejected.
- CodeRabbit's automated GitHub review found and fixed 3 more in `bcfc2afcc`: an exact-match fix for `cliDma()`'s `"show"` parsing (was a prefix match via `strncasecmp`+`len`), `atoi()`→`strtol()` for the option-value parser (rejects non-numeric input instead of silently treating it as 0), and `strcpy`→`memcpy` for writing "NONE" into the fixed-size value buffer (matches BF's own `optToString()` pattern exactly).
- A follow-up CodeRabbit trace correctly caught a gap in my own review of the above: the index parser's `atoi(pch) - 1` was left unchanged on the (mistaken) reasoning that it was "already range-checked" — that only covers out-of-range values, not trailing-garbage input like `"1abc"` (which `atoi` parses as `1`, silently resolving to a valid index). Fixed in `c52fde5cf` with the same `strtol()`+`endptr` pattern as the option-value parser, bounds-checked in `long` arithmetic before the `int` cast to avoid overflow UB.

## Test plan
- [x] `make clean_test && make test` — 42/42 binaries pass, 0 fail (no regression from baseline)
- [x] Bench build incl. SITL: `HELIOSPRING TUNERCF405 SKYSTARSF405AIO PYRODRONEF7 FOXEERF722V4 FOXEERF405 APEXF7 TMOTORF7 SITL STELLARH7DEV` — 10/10 clean, 0 code warnings
- [ ] Hardware bench check recommended (not blocking): confirm `dma UART_TX <n> list` prints real stream options on a target board, then set + save + reboot + confirm UART still initializes

### DIFF
--- a/src/main/interface/cli.c
+++ b/src/main/interface/cli.c
@@ -4027,7 +4027,8 @@ static const dmaoptEntry_t dmaoptEntryTable[] = {
 
 static const char *dmaoptValueToString(dmaoptValue_t opt, char *buf) {
     if (opt == DMA_OPT_UNUSED) {
-        strcpy(buf, "NONE");
+        // fixed-size copy, not strcpy -- matches BF's own optToString() (cli.c:5360, 4.5-maintenance)
+        memcpy(buf, "NONE", DMA_OPT_STRING_BUFSIZE);
     } else {
         tfp_sprintf(buf, "%d", opt);
     }
@@ -4128,12 +4129,16 @@ static void cliDmaopt(char *cmdline) {
     if (strcasecmp(pch, "none") == 0) {
         optval = DMA_OPT_UNUSED;
     } else {
-        optval = atoi(pch);
-        // reject out-of-range input before it silently truncates to the int8_t dmaoptValue_t parameter
-        if (optval < INT8_MIN || optval > INT8_MAX || !dmaGetChannelSpecByPeripheral(entry->peripheral, index, optval)) {
+        char *endptr;
+        long parsed = strtol(pch, &endptr, 10);
+        // reject non-numeric input (atoi would silently parse it as 0, a value indistinguishable
+        // from an explicit "0") and out-of-range input before it truncates to int8_t
+        if (*endptr != '\0' || endptr == pch || parsed < INT8_MIN || parsed > INT8_MAX
+            || !dmaGetChannelSpecByPeripheral(entry->peripheral, index, (int)parsed)) {
             cliPrintErrorLinef("INVALID DMA OPTION FOR %s %d: '%s'", entry->device, index + 1, pch);
             return;
         }
+        optval = (int)parsed;
     }
 
     char optvalString[DMA_OPT_STRING_BUFSIZE];
@@ -4151,8 +4156,7 @@ static void cliDmaopt(char *cmdline) {
 #endif // STM32F4 || STM32F7 || STM32H7
 
 static void cliDma(char* cmdLine) {
-    const int len = strlen(cmdLine);
-    if (len == 0 || strncasecmp(cmdLine, "show", len) == 0) {
+    if (*cmdLine == '\0' || strcasecmp(cmdLine, "show") == 0) {
         printDma();
         return;
     }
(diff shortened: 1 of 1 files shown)


nerdCopter:

Addressed findings from review #4930319339 in commit bcfc2afcc:

- cliDma() nitpick (prefix-match via strncasecmp+len): fixed. Now uses strcasecmp() for an exact "show" match plus a *cmdLine == '\0' check for the empty-input case, closing the prefix-collision gap for any future device name starting with 's'.
- atoi-no-error-detection-c (CWE-20), line 4130 (option-value parse): fixed. Replaced with strtol()+endptr, rejecting non-numeric input instead of silently treating it as 0.
- atoi-no-error-detection-c (CWE-20), line 4100 (index parse): no change. That site already range-checks the parsed value against the device's valid index count immediately after the atoi() call, so garbage input (which atoi parses as 0) is rejected there rather than silently accepted — strtol would add redundant validation with no behavior change.
- dangerous-buffer-functions-c (CWE-120), line 4029 (strcpy(buf, "NONE")): fixed. Replaced with memcpy(buf, "NONE", DMA_OPT_STRING_BUFSIZE) to match BF 4.5-maintenance's optToString() (cli.c:5360) exactly rather than introduce a new pattern.

The two additional-comments items already self-flagged as "Likely an incorrect or invalid review comment" (DEFW macro collision, UART DMA option table loop skip) need no action.

nerdCopter:
@coderabbitAI, are all actionable reviews and nitpicks resolved?

nerdCopter:

Confirmed, the index parser at line 4102 was a real gap I'd dismissed incorrectly — "already range-checked" only covers out-of-range numeric values, not trailing-garbage input like "1abc" (atoi parses the leading digit and ignores the rest).

Fixed in commit c52fde5cf: strtol()+endptr, same pattern as the option-value parser. Rejects when endptr == pch (no digits) or *endptr != '\0' (trailing garbage), with the index-1 subtraction and range check done in long arithmetic before the int cast to avoid overflow UB on out-of-int-range input.

42/42 unit tests and 10/10 bench targets (incl. SITL, STELLARH7DEV) clean against this commit.

PR #1386, opened 2026-08-14
fix(build): stamp-file dependency for version.o REVISION staleness

nerdCopter:
AI Generated pull-request

## Summary
`version.o`'s build rule only depends on `$(SRC)`, not on the `REVISION` value itself. A
dirty-tree build (`REVISION=uncommitted_<timestamp>`) followed by a clean-tree build with no
source-file changes leaves that stale object untouched by Make's own dependency graph, so the
firmware keeps reporting `uncommitted` in CLI `version` even though `HEAD` is clean and
`REVISION` now resolves to the real 7-char commit hash.

This is a regression: #1125 (`e8d439d26`) originally fixed this exact bug by making `version.o`
depend on a `FORCE` phony target. #1238 (`ae9a736d8`) reverted that back to `$(SRC)`, explicitly
to restore no-op incremental-build behavior for parallel builds — a legitimate goal, but it
reintroduced the staleness bug because `$(SRC)` cannot express "REVISION's value changed."

Fix: add `REVISION_STAMP`, a per-target file containing the current `REVISION` string. `version.o`
now depends on both `$(SRC)` and `REVISION_STAMP`, so:
- same `REVISION`, no source change → stamp untouched → no rebuild (preserves #1238's no-op
  incremental-build behavior)
- `REVISION` changes (dirty→clean transition, or a new commit landing with no other source diff)
  → stamp rewritten → `version.o` rebuilds with the correct value

`REVISION_STAMP`'s mkdir/compare logic is an explicit recipe rule (`check_revision_stamp` phony
trigger), not a parse-time `$(shell ...)` call, so it only runs when actually pulled into a build
goal — non-build goals like `make help` never touch it.

## Test plan
- [x] Reproduced the exact regression: clean build → SHA embedded; dirty a tracked non-`$(SRC)`
      file → rebuild → `uncommitted_<timestamp>` embedded; revert to clean, zero further changes →
      rebuild → `version.o` correctly recompiles with the real SHA (this transition was previously
      silently skipped)
- [x] Confirmed no-op case unaffected: rebuild again with zero changes → `version.o` mtime
      unchanged, no unnecessary recompile
- [x] Confirmed non-build goals don't touch the stamp: `make help` leaves `.revision_stamp` absent
- [x] Hardware bench check: flashed HELIOSPRING, TUNERCF405, and PYRODRONEF7, all report `(7da81fa)` — the
      real commit hash, not `uncommitted` — in CLI `version`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved build accuracy by detecting revision changes and recompiling affected components when needed.
  * Preserved incremental build performance when the revision remains unchanged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -304,8 +304,14 @@ CLEAN_ARTIFACTS += $(TARGET_HEX)
 CLEAN_ARTIFACTS += $(TARGET_ELF) $(TARGET_OBJS) $(TARGET_MAP)
 CLEAN_ARTIFACTS += $(TARGET_LST)
 
-# Rebuild version.o whenever any source changes so the embedded timestamp stays current
-$(OBJECT_DIR)/$(TARGET)/build/version.o : $(SRC)
+# Rebuild version.o whenever any source changes, or REVISION itself changes (e.g. a dirty-tree
+# build followed by a clean-tree build with no source diff) so REVISION never goes stale in the
+# embedded firmware string. REVISION_STAMP's mtime only updates when its content actually changes,
+# so a same-REVISION incremental build still skips the recompile.
+REVISION_STAMP  := $(OBJECT_DIR)/$(TARGET)/build/.revision_stamp
+$(shell mkdir -p $(dir $(REVISION_STAMP)))
+$(shell [ "$$(cat $(REVISION_STAMP) 2>/dev/null)" = "$(REVISION)" ] || echo "$(REVISION)" > $(REVISION_STAMP))
+$(OBJECT_DIR)/$(TARGET)/build/version.o : $(SRC) $(REVISION_STAMP)
 
 # All known non-board phony targets.  Used as the single source of truth
 # for both .PHONY and _SUMMARY_SKIP_GOALS — add new phony targets here only.
--- a/Makefile
+++ b/Makefile
@@ -306,12 +306,19 @@ CLEAN_ARTIFACTS += $(TARGET_LST)
 
 # Rebuild version.o whenever any source changes, or REVISION itself changes (e.g. a dirty-tree
 # build followed by a clean-tree build with no source diff) so REVISION never goes stale in the
-# embedded firmware string. REVISION_STAMP's mtime only updates when its content actually changes,
-# so a same-REVISION incremental build still skips the recompile.
+# embedded firmware string. REVISION_STAMP's mtime only advances when its content actually
+# changes, so a same-REVISION incremental build still skips the recompile. check_revision_stamp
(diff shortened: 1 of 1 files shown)


nerdCopter:
@CodeRabbitAI

AI generated comment

Not requesting a full review — asking for analysis on three specific points before this goes further:

1. Is the `REVISION_STAMP` approach (Makefile lines ~307-313) a correct and minimal resolution to the staleness bug described in the PR body? Specifically: does making `version.o` depend on both `$(SRC)` and a stamp file whose content is `$(REVISION)` correctly cover every case where `REVISION`'s resolved value can change without any `$(SRC)` file's mtime changing?

2. Does this adhere to this codebase's existing Makefile design/architecture and prior decisions? In particular, #1238 (`ae9a736d8`) deliberately moved `version.o` off a `FORCE` dependency and onto `$(SRC)` for parallel-build no-op-incremental-build correctness. Does the stamp-file approach preserve that intent while still closing the gap #1238 reopened (originally fixed by #1125 / `e8d439d26`)?

3. Is the implementation OS-agnostic where appropriate? The stamp-file logic uses `$(shell mkdir -p ...)` and a `[ ... ] || echo ... > ...` POSIX-shell comparison inside a `$(shell ...)` call. Is this consistent with the rest of the Makefile's existing shell/OS assumptions (e.g. the pre-existing `git diff --shortstat`, `date`, `grep`/`awk` calls), or does it introduce a new portability constraint (e.g. Windows without MSYS2/WSL/Cygwin) beyond what already exists?

nerdCopter:

Addressed the architectural nit from the analysis above in 7da81fad4:

Moved the REVISION_STAMP mkdir/compare logic out of parse-time `$(shell ...)` calls and into an
explicit recipe rule (`$(REVISION_STAMP): check_revision_stamp`, phony trigger). It now only runs
when REVISION_STAMP is actually pulled in as a prerequisite of a real build goal — confirmed
`make TARGET=SITL help` no longer creates or touches the stamp file. Re-ran the full A/B/C/D
regression sequence (clean->real SHA, dirty->uncommitted, dirty->clean transition with zero
source diff->correct SHA rebuild, true no-op->mtime unchanged) against the updated rule; all pass.

The coarse-mtime-filesystem caveat noted in the analysis is accepted as-is — not applicable on the
project's Linux CI/build environment.

nerdCopter:
post-merge validation of the `fix/version-o-stale-revision` branch.
```
=== Build summary: 459 succeeded, 0 failed ===

real	16m34.127s
user	702m33.150s
sys	70m25.164s

```

Issue #1363, opened 2026-08-05
Fleet-wide dmaInit() -> dmaAllocate() migration for the 13 remaining call sites (finish BF parity)


nerdCopter:
AI Generated [issue-ticket]

## Status (2026-08-18)

Re-scoped 2026-08-15 into 5 staged PRs (one worktree/branch/PR per driver class), after a
fresh count found **16 real call sites**, not 13 (this issue's original count predates
PR#1385's DMA/LL host-test infra work). Staging order: ADC → transponder → LED strip →
motors/DSHOT → SD card.

| Stage | Class | Files | PR | Status |
|---|---|---|---|---|
| 1/5 | ADC | `adc_stm32{f4,f7,h7}xx.c` | #1389 | **MERGED** `1ec15ae40e` |
| 2/5 | Transponder | `transponder_ir_io_{stdperiph,hal}.c` | #1393 | **MERGED** `d80ff20afa` |
| 3/5 | LED strip | `light_ws2811strip_{stdperiph,hal}.c` | #1394 | **MERGED** `a4ca6ad328`, hardware-verified on HELIOSPRING (F4) + PYRODRONEF7 (F7) |
| 4/5 | Motors/DSHOT | `pwm_output_dshot.c`, `pwm_output_dshot_hal.c` | #1397 | **MERGED** `a539314b1` (2026-08-23) — hardware-verified on all 3 real aircraft (HELIOSPRING, FOXEERF722V4, STELLARH7DEV): `dma`/`resource` CLI check byte-identical vs. historical baselines, plus arm+idle-spin+proportional-yaw-response confirmed on all 3 |
| 5/5 | SD card | `sdcard.c`, `sdio_{f4,f7,h7}xx.c` | #1398 | **OPEN DRAFT** — compile-gate only; no `USE_SDCARD`-capable board (flight or bench) exists in inventory at all, confirmed with the user — accepted as a documented, low-severity risk (logging-only, not flight-critical), not pending further code work |

All 5 stages now have a PR. Each mechanical `dmaInit()` → `dmaAllocate()` swap uncovered at
least one real reachability or ordering bug beyond the surface-level call-site change — see
each PR's description and its `CONTEXT_*.md` doc for specifics (a NULL-deref on claim
failure, an always-true init-success flag, a claim-before-hardware-touch ordering issue, and
a double-init/shared-resource idempotency gap were each found and fixed across the 5
stages, not assumed away).

[...]

Issue #1349, opened 2026-08-01
dmaInit() has no ownership check, unlike BF's dmaAllocate() — runtime DMA stream steal risk


nerdCopter:
AI Generated issue-ticket

## Summary

`dmaInit()` (`src/main/drivers/dma_stm32f4xx.c`, and the equivalent in
`dma_stm32f7xx.c`/`dma_stm32h7xx.c`) unconditionally overwrites
`dmaDescriptors[index].owner`, with no check of the stream's current owner:

```c
void dmaInit(dmaIdentifier_e identifier, resourceOwner_e owner, uint8_t resourceIndex) {
    if (!dmaIdentifierIsValid(identifier)) {
        return;
    }
    const int index = DMA_IDENTIFIER_TO_INDEX(identifier);
    RCC_AHB1PeriphClockCmd(DMA_RCC(dmaDescriptors[index].dma), ENABLE);
    dmaDescriptors[index].owner = owner;
    dmaDescriptors[index].resourceIndex = resourceIndex;
}
```

Compare to `dmaAllocate()` (`src/main/drivers/bus_spi.c`'s only caller path,
`src/main/drivers/dma_stm32f4xx.c` definition), which does check:

```c
bool dmaAllocate(dmaIdentifier_e identifier, resourceOwner_e owner, uint8_t resourceIndex) {
    ...
    if (dmaDescriptors[index].owner != OWNER_FREE) {
        return false;
    }
    dmaDescriptors[index].owner = owner;
    ...
}
```

`dmaInit()` has 26 call sites in EmuFlight: motors (`pwm_output_dshot.c`,
`pwm_output_dshot_hal.c`), UART (`serial_uart_stm32f4xx.c`, `_f30x.c`, `_f10x.c`,
`_stm32f7xx.c`), ADC (`adc_stm32f4xx.c`, `_f30x.c`, `_f7xx.c`, `_f10x.c`, `_stm32h7xx.c`),
LED strip (`light_ws2811strip_hal.c`, `_stdperiph.c`), SD card (`sdcard.c`, `sdio_f4xx.c`,
`sdio_f7xx.c`), transponder (`transponder_ir_io_hal.c`, `_stdperiph.c`). Every one of these
can silently steal a DMA stream's ownership from whatever previously held it, with no
error, no warning, and no way for the previous owner to know its stream registers are
about to be reprogrammed out from under it.

## Confirmed by Betaflight comparison

[...]

PR #1389, opened 2026-08-15
fix(adc): migrate dmaInit() to ownership-checked dmaAllocate()

nerdCopter:
AI Generated [pull-request]

## Summary

`adcInit()` (F4/F7/H7) claims its DMA stream via unchecked `dmaInit()`, which
unconditionally overwrites `dmaDescriptors[index].owner` regardless of the
current owner. `bus_spi.c` and the UART drivers already use the
ownership-checked `dmaAllocate()` for this. This PR converts all 3 ADC files
to the same pattern, matching Betaflight 4.5-maintenance's `adc_stm32*xx.c`
exactly: `if (!dmaAllocate(...)) { return; }` followed by `dmaEnable(...)`.

Stage 1 of 5 for #1363 (fleet-wide `dmaInit()` -> `dmaAllocate()` migration).
Remaining stages (transponder, LED strip, motors/DSHOT, SD card) tracked as
separate follow-up PRs against the same issue.

## Why this is safe

- `adcInit()`'s DMA claim is boot-time, single call site, always executed
  before `spiInitBusDMA()` in `fc_init.c` — no other code re-invokes it at
  runtime, so `dmaAllocate()` cannot structurally fail on any current target.
  This mirrors the same collision-free argument already established for the
  other 12 remaining `dmaInit()` sites in #1363's investigation.
- `adcInit()` is `void` in both EF and BF, so the early-return-on-failure
  pattern needed zero signature or caller changes anywhere.
- `dmaEnable()` is required alongside `dmaAllocate()` — `dmaAllocate()` only
  sets ownership bookkeeping, it never enables the DMA controller's
  peripheral clock (found the hard way in PR #1356, same fix applied here
  from the start).

## Verification

- Build: HELIOSPRING (F4), FOXEERF722V4 (F7), STELLARH7DEV (H7) — all real
  aircraft targets — plus TUNERCF405, SKYSTARSF405AIO, PYRODRONEF7,
  FOXEERF405, APEXF7, TMOTORF7, SITL. 10/10 succeeded, no warnings.
- Host unit tests: 47/47 test binaries pass, no change from baseline.
- **Hardware flight-verification: complete.** Live `dma` CLI output on all 3
  real aircraft matches each target's own static DMA-stream macro exactly
  (HELIOSPRING `DMA2 Stream 4`, FOXEERF722V4 `DMA2 Stream 0` on ADC3,
  STELLARH7DEV `DMA2 Stream 1`) — no regression, since the migration only
  changes whether the claim is checked, never which stream gets picked.
  FOXEERF722V4 also showed a correct 22.9V battery reading, confirming the
  DMA transfer itself is functioning end-to-end, not just the ownership
  claim. Clean boot, no fault, on all 3 boards.

## Related

- #1363 — parent issue, fleet-wide migration
- #1349 / PR #1356 — established the `dmaAllocate()`+`dmaEnable()` pattern
  this PR reuses

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ADC initialization across STM32F4, STM32F7, and STM32H7 platforms.
  * Added safer DMA resource allocation and handling when resources are unavailable.
  * Ensured allocated DMA resources are properly enabled during ADC setup.

### DIFF
--- a/src/main/drivers/adc_stm32f4xx.c
+++ b/src/main/drivers/adc_stm32f4xx.c
@@ -245,7 +245,10 @@ void adcInit(const adcConfig_t *config) {
     ADC_DMARequestAfterLastTransferCmd(adc.ADCx, ENABLE);
     ADC_DMACmd(adc.ADCx, ENABLE);
     ADC_Cmd(adc.ADCx, ENABLE);
-    dmaInit(dmaGetIdentifier(adc.DMAy_Streamx), OWNER_ADC, 0);
+    if (!dmaAllocate(dmaGetIdentifier(adc.DMAy_Streamx), OWNER_ADC, 0)) {
+        return;
+    }
+    dmaEnable(dmaGetIdentifier(adc.DMAy_Streamx));
     DMA_DeInit(adc.DMAy_Streamx);
     DMA_InitTypeDef DMA_InitStructure;
     DMA_StructInit(&DMA_InitStructure);
--- a/src/main/drivers/adc_stm32f7xx.c
+++ b/src/main/drivers/adc_stm32f7xx.c
@@ -263,7 +263,10 @@ void adcInit(const adcConfig_t *config) {
             /* Channel Configuration Error */
         }
     }
-    dmaInit(dmaGetIdentifier(adc.DMAy_Streamx), OWNER_ADC, 0);
+    if (!dmaAllocate(dmaGetIdentifier(adc.DMAy_Streamx), OWNER_ADC, 0)) {
+        return;
+    }
+    dmaEnable(dmaGetIdentifier(adc.DMAy_Streamx));
     adc.DmaHandle.Init.Channel = adc.channel;
     adc.DmaHandle.Init.Direction = DMA_PERIPH_TO_MEMORY;
     adc.DmaHandle.Init.PeriphInc = DMA_PINC_DISABLE;
--- a/src/main/drivers/adc_stm32h7xx.c
+++ b/src/main/drivers/adc_stm32h7xx.c
@@ -314,7 +314,10 @@ void adcInit(const adcConfig_t *config)
         }
 
         dmaIdentifier_e dmaId = dmaGetIdentifier((DMA_Stream_TypeDef *)adc.dmaResource);
-        dmaInit(dmaId, OWNER_ADC, 0);
+        if (!dmaAllocate(dmaId, OWNER_ADC, 0)) {
+            return;
+        }
+        dmaEnable(dmaId);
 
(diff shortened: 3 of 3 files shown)


nerdCopter:
@CodeRabbitAI review

Small, focused diff: 3 files (adc_stm32f4xx.c, adc_stm32f7xx.c, adc_stm32h7xx.c), 5 lines
each, converting unchecked dmaInit() to ownership-checked dmaAllocate()+dmaEnable() in
adcInit(). Please analyze for:
- Correctness of the dmaAllocate()/dmaEnable() sequencing vs. bus_spi.c's established
  usage of the same primitives.
- Any early-return-on-failure interaction with surrounding code in each of the 3 files,
  especially adc_stm32h7xx.c, which has multiple existing early-return paths in adcInit().
- Whether the void return type (no signature change, matching both EF's and BF's own
  adcInit()) is handled correctly at the call site.

nerdCopter:

Fixed in 27305cfcf, addressing https://github.com/emuflight/EmuFlight/pull/1389#issuecomment-5304631814 (P2 — H7 internal ADC init coupling).

`dmaAllocate()` failure for the regular ADC DMA stream now falls through
into the successful-allocation branch's scope only, instead of returning
from `adcInit()` entirely — `adcInitInternalInjected()` (VREFINT/temp
sensor) still runs on a regular-ADC DMA conflict, matching F4/F7's
decoupled ordering. Pre-existing `HAL_DMA_Init`/`HAL_ADC_Start_DMA`
failure returns were left untouched (out of this PR's scope, unaffected
by the restructure since a bare `return;` still exits the function from
the new nesting level).

Please verify.

nerdCopter:

Requesting a fresh full-diff analyze on the current HEAD (both commits: the
original ADC migration `39676e255` and the H7 scoping fix `27305cfcf`).
Specifically want a second look at:

1. Whether `adc_stm32h7xx.c`'s restructured `dmaAllocate()` block (commit
   `27305cfcf`) correctly preserves the pre-existing `HAL_DMA_Init`/
   `HAL_ADC_Start_DMA` early-return behavior unchanged, while only
   decoupling the new `dmaAllocate()` failure path from
   `adcInitInternalInjected()`.
2. Whether `adc_stm32f4xx.c`/`adc_stm32f7xx.c`'s unchanged `return;` on
   `dmaAllocate()` failure is safe given their `adcInitInternalInjected()`
   call happens earlier in the function, before the DMA claim (verified
   directly: F4 line 232 vs claim at 248; F7 lines 247/249 vs claim at
   266) — so no equivalent coupling exists there.
3. Any other spot in the 3-file diff with a similar ordering-dependency
   risk not yet caught.

PR #1400, opened 2026-08-18
feat(cli): warn when a UART's configured dmaopt loses its DMA claim

### LINKED ISSUE #1388: feat(dma): no diagnostic when a configured dmaopt loses its boot-time DMA claim (medium-priority)
AI Generated issue-ticket

## Summary

When a peripheral's `dmaopt` resolves to a DMA stream/channel that's already claimed by another
peripheral at boot, the claiming call silently falls back to a non-DMA mode with zero CLI
message, log entry, or other diagnostic. This is the existing, inherited design in both EF and
BF — not a defect introduced by any recent PR — but PR #1384 changes the practical stakes: it
adds a CLI path (`dma <device> <index> <option>`) that lets a pilot freely reassign any UART's
`dmaopt` at runtime. Previously these values were fixed by target maintainers, who presumably
picked defaults that avoid contention; now a user can set a value that collides with another
peripheral, and the CLI will report full success with no indication that the runtime effect
differs from what was requested.

## Where this was found

Investigated during PR #1384's hardware verification (`feat/cli-dmaopt-resource`,
`~/SYNC/nerdCopter-GIT/AI/EF/feat/cli-dmaopt-resource/CONTEXT_cli-dmaopt-resource.md`), chasing
why one bench board (TMOTORF7) showed a UART actually claiming a DMA stream while five others
did not. The actual root cause for that specific case turned out to be board-config-driven (a
`target.h` default, unrelated), not this gap — but tracing it surfaced this real, general
silent-fallback behavior along the way.

## Code paths (confirmed identical mechanism in both projects)

- EF: `src/main/drivers/serial_uart_stm32f4xx.c:192-208` (`serialUART()`) — same pattern in
  `serial_uart_stm32f7xx.c`. On claim failure: `s->txDMAChannel = 0; s->txDMAStream = NULL;`
  with only a source comment (`// stream owned by another peripheral ... fall back to IRQ-driven
  TX`) — no runtime-visible signal.
- BF: `src/main/drivers/serial_uart.c:439-449` (`uartConfigureDma()`) — `dmaAllocate()` return
  value is checked, but on failure the function simply doesn't set up DMA; no error path, no
  log, no CLI-visible state change.

## What's missing

No mechanism exists to tell a user, after setting a `dmaopt` value (or after boot), whether that
specific UART is actually running on DMA or fell back to IRQ. `dma`/`dma show`'s live utilization
dump does show it indirectly (the stream shows a different owner than expected), but only if the
user thinks to cross-check it — nothing surfaces the mismatch proactively, and CLI `dma <device>
<index>` (get) only echoes the *configured* value, not whether it's actually in effect.

## Proposed directions (not a mandated fix — needs design discussion)

1. At CLI `dma <device> <index> <option>` set time, check `dmaGetOwner()` for the resolved
   stream identifier and print a warning if it's currently held by something else (imperfect —
   boot-time claim order can differ from the CLI's live session state, but better than silence).
2. Add a runtime-queryable "is this UART actually DMA-backed right now" indicator, exposed via
   `status` or a new `dma` sub-view, computed from `s->txDMAStream`/`s->rxDMAStream` being
   non-NULL rather than requiring the user to cross-reference the utilization dump manually.
3. At minimum: document the silent-fallback behavior in the `dma` command's CLI help text, so a
   user hitting unexpectedly-slow serial throughput has a documented reason to check for it.

## Scope

Cross-cutting UART DMA driver behavior (EF and BF both affected identically) — not board-
specific, not caused by any single PR, and not blocking PR #1384 (which only adds the CLI setter
for an already-existing value; the fallback behavior itself predates it).

nerdCopter:
AI Generated pull-request

## Summary

`serialUART()` silently falls back to IRQ-driven mode when its DMA claim
(`uartDmaClaim()`/`dmaAllocate()`) loses a stream/channel to another peripheral at boot — no CLI
message, log entry, or other diagnostic exists anywhere (EF or BF, inherited design). PR #1384's
new `dma` CLI setter increased real user exposure to this by making it easy to configure a
`dmaopt` value that collides with another peripheral's claim.

Adds `printDmaoptClaimStatus()` in `src/main/interface/cli.c`, which cross-checks a resolved
`dmaopt` stream against `dmaGetOwner()`/`dmaGetResourceIndex()` — the same live bookkeeping
`dmaAllocate()`/`uartDmaClaim()` already populate — and prints a `# ... WARNING ...` line on
mismatch. Wired into both:
- `printDmaoptEntry()` — the `dma <device> <index>` get path and `dma <device> list` / bare
  `dma list` display path
- `cliDmaopt()` — the set path, so a `dma UART_TX 1 <n>` set gets an immediate proactive warning
  if the target stream is already held by something else

`OWNER_FREE` (stream never claimed by anyone) is treated as a no-op rather than a warning, since
it's indistinguishable from "this UART slot isn't opened as a serial port this boot" — the
overwhelmingly common state for any board with unused UART indices.

Closes #1388.

## Test plan

- [x] `make test` (unit tests) — 47 suites, 0 failures
- [x] `make HELIOSPRING TUNERCF405 SKYSTARSF405AIO PYRODRONEF7 FOXEERF722V4 FOXEERF405 APEXF7 TMOTORF7 STELLARH7DEV SITL` — all succeed, 0 new warnings
- [x] Reviewed via local CodeRabbit CLI (0 findings) and opencode second opinion (1 finding
      investigated, confirmed false positive — cited a SITL-only file never compiled where this
      feature runs)
- [x] Hardware-verified on TMOTORF7 (build `dd9d2b8`, EmuFlight 0.4.3, MSP API 1.54), all three
      code paths through `printDmaoptClaimStatus()` exercised live via CLI:
  - Real conflict: `dma UART_TX 3 0` (only option, resolves to DMA1 Stream 3, already held by
    `SPI_SDI 2`) → `# UART_TX 3: WARNING stream held by SPI_SDI 2, not this UART -- expect
    IRQ-driven fallback` on both set and get
  - `OWNER_FREE` (non-`NONE` value resolving to a genuinely-unclaimed stream): `dma UART_TX 2 0`
    (resolves to DMA1 Stream 6, `FREE`) → no warning
  - Explicit `NONE`: `dma UART_TX 3 none` → no warning
  - No `save` run at any point; flash-persisted defaults untouched

## Summary by CodeRabbit

* **Bug Fixes**
  * Added diagnostics for UART DMA configuration mismatches.
  * The CLI now identifies when a selected DMA stream is assigned to an unexpected resource and warns that operation may fall back to interrupt-driven transfers.
  * Free DMA streams no longer generate unnecessary warnings.

### DIFF
--- a/src/main/interface/cli.c
+++ b/src/main/interface/cli.c
@@ -4053,6 +4053,28 @@ static dmaoptValue_t *dmaoptAddr(const dmaoptEntry_t *entry, int index) {
     return (dmaoptValue_t *)(base + entry->stride * index + entry->offset);
 }
 
+// Surfaces serialUART()'s otherwise-silent IRQ-driven fallback via dmaAllocate()'s live ownership state.
+static void printDmaoptClaimStatus(const dmaoptEntry_t *entry, int index, const dmaChannelSpec_t *dmaChannelSpec) {
+    const dmaIdentifier_e identifier = dmaGetIdentifier((DMA_Stream_TypeDef *)dmaChannelSpec->ref);
+    const resourceOwner_e expectedOwner = (entry->peripheral == DMA_PERIPH_UART_TX) ? OWNER_SERIAL_TX : OWNER_SERIAL_RX;
+    const resourceOwner_e actualOwner = dmaGetOwner(identifier);
+    const uint8_t actualIndex = dmaGetResourceIndex(identifier);
+    if (actualOwner == expectedOwner && actualIndex == RESOURCE_INDEX(index)) {
+        return;
+    }
+    // OWNER_FREE is indistinguishable from "this UART not opened this boot" (unassigned serial function) -- skip to avoid false positives on every unused UART slot.
+    if (actualOwner == OWNER_FREE) {
+        return;
+    }
+    if (actualIndex > 0) {
+        cliPrintLinef("# %s %d: WARNING stream held by %s %d, not this UART -- expect IRQ-driven fallback",
+                      entry->device, index + 1, ownerNames[actualOwner], actualIndex);
+    } else {
+        cliPrintLinef("# %s %d: WARNING stream held by %s, not this UART -- expect IRQ-driven fallback",
+                      entry->device, index + 1, ownerNames[actualOwner]);
+    }
+}
+
 static void printDmaoptEntry(const dmaoptEntry_t *entry, int index) {
     const dmaoptValue_t *addr = dmaoptAddr(entry, index);
     if (!addr) {
@@ -4066,6 +4088,7 @@ static void printDmaoptEntry(const dmaoptEntry_t *entry, int index) {
         if (dmaChannelSpec) {
             cliPrintLinef("# %s %d: " DMASPEC_FORMAT_STRING, entry->device, index + 1,
                           DMA_CODE_CONTROLLER(dmaChannelSpec->code), DMA_CODE_STREAM(dmaChannelSpec->code), DMA_CODE_CHANNEL(dmaChannelSpec->code));
+            printDmaoptClaimStatus(entry, index, dmaChannelSpec);
         }
     } else {
         cliPrintLinef("dma %s %d NONE", entry->device, index + 1);
@@ -4161,6 +4184,13 @@ static void cliDmaopt(char *cmdline) {
         cliPrintLinef("# dma %s %d: changed from %s to %s", entry->device, index + 1, orgvalString, optvalString);
     } else {
         cliPrintLinef("# dma %s %d: no change: %s", entry->device, index + 1, orgvalString);
+    }
+
+    if (optval != DMA_OPT_UNUSED) {
+        const dmaChannelSpec_t *dmaChannelSpec = dmaGetChannelSpecByPeripheral(entry->peripheral, index, optval);
+        if (dmaChannelSpec) {
+            printDmaoptClaimStatus(entry, index, dmaChannelSpec);
+        }
     }
 }
 #endif // STM32F4 || STM32F7 || STM32H7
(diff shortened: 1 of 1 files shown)


nerdCopter:

Requesting analysis (not a full review) on two specific points in `printDmaoptClaimStatus()`
(`src/main/interface/cli.c`):

1. The function early-returns without printing a warning when `dmaGetOwner(identifier) ==
   OWNER_FREE`, on the reasoning that `OWNER_FREE` is indistinguishable from "this UART device
   slot was never opened as a serial port this boot" (the common case for unused UART indices)
   versus a genuine unexpected claim failure. Is there a real code path where a UART *is*
   actually opened via `serialUART()`, its DMA claim legitimately fails for a reason distinct
   from "another peripheral already holds this stream" (i.e. leaves the stream `OWNER_FREE`
   rather than owned by a rival), and where suppressing the warning here would hide an
   actionable signal rather than avoid a false positive?

2. `dmaGetIdentifier()` returns `0` (an invalid identifier) when the passed stream pointer isn't
   found in `dmaDescriptors[]`. Confirm whether `dmaGetOwner()`/`dmaGetResourceIndex()` in the
   actual compiled implementations for this feature's build targets —
   `src/main/drivers/dma_stm32f4xx.c`, `dma_stm32f7xx.c`, `dma_stm32h7xx.c` (NOT
   `src/main/drivers/dma.c`, which is SITL-only per `make/mcu/SITL.mk` and never compiled where
   this CLI code runs) — guard against identifier `0` via `dmaIdentifierIsValid()` before
   indexing `dmaDescriptors[]`, i.e. that no out-of-bounds read is reachable through this new
   function on any real target.

Issue #1381, opened 2026-08-11
chore(STRIXF10): remove stale commented-out defines (USE_QUAD_MIXER_ONLY, DEFAULT_PIDS_*)


nerdCopter:
AI Generated issue-ticket

## Summary

`src/main/target/STRIXF10/target.h` has 4 stale, commented-out defines with no effect:

```c
//#define DEFAULT_PIDS_ROLL   {45, 70, 20, 0}
//#define DEFAULT_PIDS_PITCH  {45, 70, 20, 0}
//#define DEFAULT_PIDS_YAW    {45, 70, 5, 0}

// #define USE_QUAD_MIXER_ONLY
```

## `USE_QUAD_MIXER_ONLY` is not dead code

`USE_QUAD_MIXER_ONLY` is a compile-time feature flag, not dead code, and must not be removed
from the 12 guarded call sites that use it. No tracked target defines it by default, but
EmuFlight's `Makefile` composes `CFLAGS` from `OPTIONS`/`EXTRA_FLAGS`
(`make TARGET OPTIONS=USE_QUAD_MIXER_ONLY` or `EXTRA_FLAGS=-DUSE_QUAD_MIXER_ONLY`), so the flag
remains settable from the command line and the guarded branches remain reachable. Confirmed via
`Makefile:220` (`$(addprefix -D,$(OPTIONS))`) and `Makefile:249` (`$(EXTRA_FLAGS)`), both
composed into `CFLAGS`.

The 12 sites — `mixer.c`, `cli.c`, `msp.c`, `config.c`, `fc_core.c`, `pwm_output_counts.h` — stay
untouched. The commented-out definer in `STRIXF10/target.h` is the only genuinely dead artifact:
a comment cannot be restored via `OPTIONS`/`EXTRA_FLAGS`, unlike the macro itself.

## Fix

Remove the 4 stale commented-out lines from `STRIXF10/target.h` only. No other file changes.

nerdCopter:
@coderabbitAI , confirm the correctness of this issue-ticket. is it really dead-code, or is it designed to be a compile time flag/directive?

nerdCopter:
AI Generated comment

Confirmed and corrected — title and description above updated to drop the "remove dead code"
framing and the incorrect 12-site removal claim. `USE_QUAD_MIXER_ONLY` stays as a compile-time
`OPTIONS`/`EXTRA_FLAGS` feature flag; only the 4 stale commented-out lines in
[...]

PR #1417, opened 2026-08-28
fix: guard cli.c/servos.c against USE_QUAD_MIXER_ONLY

### LINKED ISSUE #1416: bug: USE_QUAD_MIXER_ONLY fails to build (cli.c) and fails to link (servos.c)
AI Generated issue-ticket

## Summary

`USE_QUAD_MIXER_ONLY` is a real, still-supported compile-time feature flag (confirmed:
emuflight/EmuFlight#1381, comment
https://github.com/emuflight/EmuFlight/issues/1381#issuecomment-5444847924), settable via
`make TARGET OPTIONS=USE_QUAD_MIXER_ONLY` or `EXTRA_FLAGS=-DUSE_QUAD_MIXER_ONLY`. No tracked
target defines it by default, so this combination has not been build-tested and has bit-rotted.
Confirmed by building `FOXEERF722V4` (which defines `USE_SERVOS` by default via
`target/common_fc_pre.h`) with `EXTRA_FLAGS="-DUSE_QUAD_MIXER_ONLY"` against current master
(`ffda4cfcc`).

## Bug 1 — compile error: `src/main/interface/cli.c`

`cliServoMix()`'s `smix load <name>` subcommand (line 1759, gated only by `#ifdef USE_SERVOS`)
references `mixerNames[]` unconditionally:

```c
for (uint32_t i = 0; ; i++) {
    if (mixerNames[i] == NULL) {
```

`mixerNames[]` is defined under `#ifndef USE_QUAD_MIXER_ONLY` (`cli.c:215-225`). With both flags
set, this fails to compile:

```
./src/main/interface/cli.c:1759:21: error: 'mixerNames' undeclared (first use in this function); did you mean 'ownerNames'?
```

## Bug 2 — link error: `src/main/flight/servos.c`

`servosInit()` (line 217) references `mixers[currentMixerMode].useServo` unconditionally:

```c
useServo = mixers[currentMixerMode].useServo;
```

`mixers[]` is defined under `#ifndef USE_QUAD_MIXER_ONLY` (`mixer.c:150-330`). The reference
compiles clean — only an `extern const mixer_t mixers[];` declaration (`mixer.h:111`) is checked
at compile time — but fails at link time, since `mixer.o` built with `USE_QUAD_MIXER_ONLY`
exports no `mixers` symbol. Confirmed via `arm-none-eabi-nm -u servos.o`, showing `U mixers`
with no matching definition in `mixer.o`.

Bug 2 only surfaces once bug 1 is bypassed (verified via a temporary, uncommitted `#if 0` stub
around the bug-1 site — no source change from this).

## Scope

Both fixes need a guard against `USE_QUAD_MIXER_ONLY` at the two call sites above — matching the
pattern the rest of the codebase already uses for `mixerNames[]`/`mixers[]` consumers (see
emuflight/EmuFlight#1381's site list). Not scoped/implemented here — sizing and exact guard
placement (e.g. whether `smix load` should degrade gracefully vs. compile out entirely under
`USE_QUAD_MIXER_ONLY`) is a design call for whoever picks this up.

nerdCopter:
AI Generated pull-request

## Summary

`USE_QUAD_MIXER_ONLY` is a real, still-supported compile-time feature flag (settable via
`OPTIONS=`/`EXTRA_FLAGS=-D`, confirmed on #1381), but no tracked target exercises it, so this
combination bit-rotted. Two bugs surfaced when building `FOXEERF722V4` (default `USE_SERVOS`)
with `EXTRA_FLAGS="-DUSE_QUAD_MIXER_ONLY"`:

1. `cli.c` `cliServoMix()`'s `smix load <name>` subcommand (guarded only by `USE_SERVOS`)
   referenced `mixerNames[]` unconditionally — that array only exists when
   `USE_QUAD_MIXER_ONLY` is *not* defined. **Compile error.**
2. `servos.c` `servosInit()` referenced `mixers[currentMixerMode].useServo` unconditionally —
   same story, but only an `extern` declaration is checked at compile time, so this compiled
   clean and failed at **link** time instead (confirmed via `nm -u servos.o`).

## Fix

Guard both against `USE_QUAD_MIXER_ONLY`, matching the pattern already used elsewhere in these
same files (`cliMotorMix()` in `cli.c`, `validateAndFixConfig()` in `config.c`):

- `smix load <name>` reports "Invalid name" under the flag (same message as any unmatched name)
  — there's no mixer-name table to look up.
- `servosInit()` falls back to `useServo = 0`, matching the real `mixers[]` table's own
  `MIXER_QUADX` entry (`useServo = false`) — the only mixer mode that can exist under this flag.
  `FEATURE_SERVO_TILT`/`FEATURE_CHANNEL_FORWARDING` still override this independently right
  after, unaffected.

No behavior change on any current build — both `#ifndef` arms are the pre-existing code,
byte-for-byte identical.

## Verification

- The actual reproduction: `make FOXEERF722V4 EXTRA_FLAGS="-DUSE_QUAD_MIXER_ONLY"` now builds
  clean end-to-end (compile + link), where it previously failed at compile, then would have
  failed at link.
- `make clean_test && make test`: 48/48 unit tests pass.
- Normal build (no flag): `HELIOSPRING`/`FOXEERF722V4` clean, byte-identical `.elf` size vs.
  pre-fix.
- Not on AGENTS.md's hardware-verification list (no flash/SPI/gyro/DMA/USB touched); no owned
  target builds with this flag anyway — compile-gate-only, same precedent as #1381/#1415.

Closes #1416.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved behavior for quad-mixer-only configurations by disabling servo use where unsupported.
  * Prevented mixer loading through the CLI in quad-mixer-only builds, returning an appropriate validation message instead.


nerdCopter:
AI generated comment

@CodeRabbitAI analyze

Requesting a full analysis of this branch. Both guarded call sites are described in the PR body
and #1416. Flag anything analysis-worthy — correctness, coverage, maintainability, or otherwise.

nerdCopter:
AI generated comment

@CodeRabbitAI Confirmed both.

P1: verified the full chain (`fc_init.c:322` → `mixerInit()` → `servoConfigureOutput()`'s
`servoMixers[currentMixerMode]`, `config.c`'s reset skipped under the flag). Fixed at the
source: `mixerInit()` now forces `currentMixerMode = MIXER_QUADX` under
`USE_QUAD_MIXER_ONLY`, ignoring the persisted value, so every downstream consumer sees a
consistent mode.

P3: fixed. Note — embedding the `#ifndef`/`#endif` directly inside the `CLI_COMMAND_DEF(...)`
macro's argument list (the obvious first approach) produces its own new compiler warning
("embedding a directive within macro arguments is not portable"), confirmed via build. Used two
full `CLI_COMMAND_DEF` entries instead.

Commit e2c7e678c. 48/48 unit tests pass, both configurations build with zero warnings, normal
build `.elf` size unchanged.

PR #1330, opened 2026-07-24
build: expand SIZE_OPTIMISED_SRC to cover init/config/telemetry paths

nerdCopter:
AI Generated pull-request

Closes #1221

## Change

Adds 41 files to `SIZE_OPTIMISED_SRC` in `make/source.mk`, forcing `-Os` instead of default
optimization on init/config/telemetry paths not on the gyro/PID control loop. Build-flag-only
change — no source logic modified in any listed file.

- Flash drivers (5): `drivers/flash.c`, `drivers/flash_m25p16.c`, `drivers/flash_w25m.c`, `drivers/flash_w25n.c`, `drivers/flash_w25q128fv.c`
- Telemetry (12): `telemetry/telemetry.c`, `telemetry/crsf.c`, `telemetry/ghst.c`, `telemetry/srxl.c`, `telemetry/frsky_hub.c`, `telemetry/hott.c`, `telemetry/jetiexbus.c`, `telemetry/smartport.c`, `telemetry/ltm.c`, `telemetry/mavlink.c`, `telemetry/msp_shared.c`, `telemetry/ibus.c`, `telemetry/ibus_shared.c`
- GPS/navigation (3): `io/gps.c`, `flight/gps_rescue.c`, `flight/position.c`
- Rangefinder (3): `drivers/rangefinder/rangefinder_hcsr04.c`, `drivers/rangefinder/rangefinder_lidartf.c`, `sensors/rangefinder.c`
- Sensor init (3): `sensors/initialisation.c`, `sensors/barometer.c`, `sensors/compass.c`
- Display/IO (10): `io/ledstrip.c`, `io/pidaudio.c`, `io/rcdevice.c`, `io/rcdevice_cam.c`, `io/displayport_max7456.c`, `io/displayport_msp.c`, `io/displayport_hdzero_osd.c`, `io/displayport_oled.c`, `io/displayport_srxl.c`, `io/displayport_crsf.c`
- CMS menus (2): `cms/cms_menu_failsafe.c`, `cms/cms_menu_gps_rescue.c`
- Other (3): `sensors/esc_sensor.c`, `drivers/camera_control.c`, `io/vtx_beesign.c`

All 41 paths verified present in `upstream/master` @ `7651c15373` by direct filesystem check —
not taken from the issue's original file list, since master may have moved files since filing.

## Why these are safe for `-Os`

[...]

PR #1432, opened 2026-09-03
fix(sensors): gate host-side lowpass filters off USE_GYRO_IMUF9001

### LINKED ISSUE #1345: IMUF9001 targets run redundant host-side lowpass filtering; consider gating but keep notch filters
AI Generated issue-ticket

## Summary
`src/main/sensors/gyro_filter_impl.h`'s `GYRO_FILTER_FUNCTION_NAME()` runs the static host-side lowpass/notch filter chain unconditionally for every gyro type, including `USE_GYRO_IMUF9001`. The IMUF9001/IMU-F coprocessor already performs its own onboard lowpass filtering (`imuf_pitch_lpf_cutoff_hz`/`imuf_roll_lpf_cutoff_hz`/`imuf_yaw_lpf_cutoff_hz`, `settings.c:537-539`), making the host-side lowpass stage redundant on these targets.

## Verified facts

- `gyro_filter_impl.h:34-36` and `:62-64` already gate the Kalman filter calls behind `#ifndef USE_GYRO_IMUF9001` — precedent for per-target gating exists in this exact function.
- `gyro_filter_impl.h:40,42-44` do **not** have that gate: `lowpass2FilterApplyFn` (line 40, under `USE_GYRO_LPF2`), `lowpassFilterApplyFn` (line 42), `notchFilter1ApplyFn`/`notchFilter2ApplyFn` (lines 43-44) run on every gyro type, IMUF9001 included.
- `settings.c:508-526` — `gyro_lowpass_hz_roll/pitch/yaw`, `gyro_lowpass2_hz_roll/pitch/yaw`, `gyro_notch1_hz`, `gyro_notch1_cutoff`, `gyro_notch2_hz`, `gyro_notch2_cutoff` have no `USE_GYRO_IMUF9001` compile guard — all are CLI/MSP-settable and functional on HELIOSPRING/MODE2FLUX/STRIXF10 today. A nonzero `gyro_lowpass_hz` on an IMUF9001 target actually filters; it is not already a no-op.

## Proposal

Gate out only the **lowpass** path (`lowpassFilterApplyFn`, `lowpass2FilterApplyFn`) for `USE_GYRO_IMUF9001` targets, since the coprocessor already performs equivalent lowpass filtering onboard. Keep the two **notch** filters (`notchFilter1ApplyFn`/`notchFilter2ApplyFn`) compiled and callable for IMUF9001 — these serve a different purpose (fixed-frequency noise rejection, e.g. prop-wash/motor-harmonic notches) that the coprocessor's lowpass does not replicate, and should remain available regardless of gyro type.

## Efficiency impact

Removes 1-2 function-pointer-indirect calls per axis per GYRO/PID loop cycle (lowpass + optional lowpass2), i.e. up to 6 calls/loop across 3 axes. Most of these currently resolve to `nullFilterApply` (a no-op) at the IMUF9001 default of 0Hz cutoff, so the runtime saving is real but modest — this is a code-correctness/intent cleanup as much as a performance change.

## Backward-compatibility question (needs a decision before implementation)

Because `gyro_lowpass_hz_*`/`gyro_lowpass2_hz_*` are currently functional (not dead) on IMUF9001 if set nonzero, gating the calls out is a behavior change for any existing IMUF9001 config with those set. Two implementation options, unresolved:
1. Leave the CLI/MSP fields settable but make them inert on IMUF9001 (silent no-op) — simpler, but a user could set a value that appears accepted and does nothing, with no CLI feedback.
2. Also gate the CLI/MSP fields themselves out for IMUF9001 (matching how `imuf_*` fields are already conditionally compiled at `settings.c:531-547`) — cleaner, but changes what a config-diff/dump looks like for these targets and needs Configurator awareness.

## Configurator dependency (unverified, flagged not assumed)

EmuConfigurator is a separate repository not available in this session, so its current gating behavior could not be checked directly. The firmware already reports gyro hardware name via MSP (the `"GYRO_IMUF9001"` string in `settings.c:118`), which is a pre-existing detection channel Configurator could key off to hide the lowpass sliders for detected IMUF9001 boards — but whether Configurator already uses this, or whether a new MSP semver would actually be needed, is unconfirmed and should be checked against Configurator's own source before implementation, not assumed from the firmware side alone.

## Relationship to PR #1283 (IMUF9001 DMA migration refactor)

Independent of the DMA-migration branch's own scope (SPI transfer mechanism, not filter architecture). No ordering dependency identified either direction — can land as a predecessor or a follow-up.

## Scope

Analysis only; no implementation performed as part of filing this issue.

nerdCopter:
AI Generated pull-request

Resolves #1345

## Summary

`GYRO_FILTER_FUNCTION_NAME()` (`src/main/sensors/gyro_filter_impl.h`) unconditionally called the host-side static lowpass filters (`lowpassFilterApplyFn`/`lowpass2FilterApplyFn`) for every gyro type, including `USE_GYRO_IMUF9001`. The IMUF9001 coprocessor already performs its own onboard lowpass filtering, making the host-side lowpass stage redundant on these targets.

Four commits, in order:
1. Gate the two lowpass filter-apply calls behind `#ifndef USE_GYRO_IMUF9001`. Notch filters (`notchFilter1ApplyFn`/`notchFilter2ApplyFn`), dynamic notch, and the Smith Predictor are unaffected — they serve fixed-frequency noise rejection and prop-wash/motor-harmonic filtering the coprocessor's onboard lowpass does not replicate.
2. Resolve the backward-compatibility question from #1345 as Option 1: `validateAndFixGyroConfig()` (`src/main/fc/config.c`) forces `gyro_lowpass_hz[]`/`gyro_lowpass2_hz[]` to 0 on `USE_GYRO_IMUF9001` targets on every config-validation pass, so `dump`/MSP always reads 0 there regardless of what a user previously set or what remains in EEPROM/PG.
3. Remove the now-dead filter storage and setup code entirely for IMUF9001: `gyroSensor_t`'s `lowpassFilterApplyFn`/`lowpassFilter[]`/`lowpass2FilterApplyFn`/`lowpass2Filter[]` fields and `gyroInitLowpassFilterLpf()` (declaration, definition, both call sites in `gyroInitSensorFilters()`) are now gated the same way. Since the value is already forced to 0, this is a build-time-only change — the filter was already inert; this just stops allocating and initializing it.
4. Cosmetic, no functional/size change: gate `gyroLowpassFilter_t` (now unused on IMUF9001) the same way, and add `static` to `gyroInitLowpassFilterLpf()`'s definition to match its forward declaration. Both from CodeRabbit analysis on commit 3.

No MSP protocol change: no field is removed, no response format changed. The fields stay present and settable in CLI/MSP for IMUF9001 targets. `MSP` label added per `AGENTS.md` § MSP Freeze Policy since forcing a user-setting's effective value has Configurator-visibility implications, even though it isn't itself a protocol change.

Option 2 (removing the fields from CLI/MSP entirely for IMUF9001) remains a separate, deliberately deferred follow-up — tracked against the MSP freeze ticket #1250, needs a dedicated pre-release MSP PR once that freeze lifts. A Configurator-side companion question (whether to disable or hide the now-inert lowpass sliders on IMUF9001 boards) is tracked on emuflight/EmuConfigurator#654 — preferred approach is disable/gray-out rather than hide, itself gated on a new MSP capability signal not yet implemented, with hide/unhide as a fallback if disable/enable proves impractical.

## Measured impact — RAM/flash

Clean-tree builds at matching commits (both sides use the short-SHA `REVISION` string, and the same gcc-13.3.1 toolchain — this branch has since merged `upstream/master`, which picked up the project's existing toolchain default):

| Target | Flash = text+data (before → after) | RAM = data+bss (before → after) |
|---|---|---|
| HELIOSPRING (IMUF9001) | 386799 → 385951 (**-848 B**) | 112752 → 112528 (**-224 B**) |
| FOXEERF722V4 (non-IMUF9001) | 391064 → 391064 (unchanged) | 98880 → 98880 (unchanged) |

FOXEERF722V4 is byte-for-byte identical before and after — confirms zero effect on non-IMUF9001 targets. The RAM savings (224 B) come from no longer allocating `lowpassFilter[3]`/`lowpass2Filter[3]` (a union sized for the largest of pt1/biquad/svf/ptn filter state) on IMUF9001 targets; the flash savings come from removing `gyroInitLowpassFilterLpf()` and its two call sites entirely for those targets, on top of the earlier filter-apply-call removal.

(Earlier revisions of this PR description cited much smaller numbers — 16 to 24 bytes flash, 0 RAM — measured before the struct/init-level gating in commit 3 above existed, and before this branch merged in the upstream toolchain update. Both prior numbers are stale; the table above is current.)

## Findings on #1281 (32kHz CPU starvation / blocky-slow-saturation)

Not expected to help, and confirmed not related after reviewing #1281's own root-cause analysis: that issue's cause is a **blocking ~15-20µs SPI transfer** in `imufSpiGyroRead()` (`accgyro_imuf9001.c:208-214`, synchronous `spiSequence()` call), consuming ~50-64% of the GYROPID task's CPU budget at a 32kHz loop rate (31.25µs period). This PR removes at most a couple of function-pointer-indirect calls per axis per loop plus some now-unneeded setup work — negligible compared to the SPI bottleneck. #1281's own fix path (replacing the blocking SPI call with true non-blocking DMA) is unrelated to anything touched here.

## Hardware verification

Confirmed on real HELIOSPRING: after flashing this PR's firmware (commit 2), EmuConfigurator save + refresh shows both gyro-lowpass sliders as OFF/unusable, as expected from the Option 1 zeroing (no Configurator-side code change needed — it already renders a 0 value as OFF).

## Test plan

- [x] `make test`: 49/49 PASS, no regressions, after every commit in this PR (no unit tests reference this code path)
- [x] Build passes: HELIOSPRING, FOXEERF722V4 (representative IMUF9001 / non-IMUF9001 targets), after every commit
- [x] `/ef-build-and-report` full cycle (bench/compile/unique/SITL target sets + unit tests): PASS
- [x] Hardware verification: real HELIOSPRING, EmuConfigurator save+refresh confirms both lowpass sliders OFF/unusable as expected


nerdCopter:

Analysis requested on this single-file change:

1. Confirm the `#ifndef USE_GYRO_IMUF9001` gate placement in `GYRO_FILTER_FUNCTION_NAME()` doesn't leave `gyroADCf` in an inconsistent state on any code path (notch filters, dynamic notch, and the Smith Predictor stage all still receive the correct, non-lowpassed value on IMUF9001 targets, matching the intended behavior).
2. Confirm no compiler warning (e.g. unused-variable, unreachable-code) is introduced on any `USE_GYRO_IMUF9001` target given the surrounding `#ifdef USE_GYRO_LPF2`/`#ifdef USE_GYRO_DATA_ANALYSE` nesting.
3. Flag anything else structurally risky about this narrow conditional-compilation change.

nerdCopter:

Addressed the documentation-drift observation from the analysis above:
- Stale comment ("apply static notch filters and software lowpass filters" on the lowpass-only block): fixed in commit dc0bf6f90 — reworded to describe only the lowpass calls it now precedes, and to state the IMUF9001 skip reason inline instead of a separate line.

No functional change.

nerdCopter:
AI Generated comment

Cross-referencing #1250 (MSP freeze tracker) — this PR gates the host-side static lowpass filters off `USE_GYRO_IMUF9001` at the filter-apply level (coprocessor already filters onboard), and (as of commit `2c49022a8`, Option 1) forces `gyro_lowpass_hz[]`/`gyro_lowpass2_hz[]` to 0 in `validateAndFixGyroConfig()` for these targets — so CLI/MSP `dump` always reads 0 there too, regardless of what a user previously set or what remains in EEPROM/PG storage.

No MSP protocol change: no field is removed, no response format changed, no freeze conflict. The fields stay present and settable in CLI/MSP for IMUF9001 targets; they're just forced back to 0 on every config-validation pass.

Configurator-hiding of these now-always-0 fields for IMUF9001 (if wanted) is a separate, still-undecided follow-up — Option 2 (removing the fields from CLI/MSP entirely) is tracked against this freeze ticket, not this PR.

nerdCopter:
AI Generated comment

## Merge-readiness summary (for future reference — no re-discovery needed)

**Final commit**: `2c49022a8` (both the filter gate and the Option 1 zeroing).

**RAM/flash — final, corrected numbers** (clean-tree builds, matching commits, same-length `REVISION` string on both sides to avoid measurement noise — see below):

| Target | Flash = text+data (before → after) | RAM = data+bss (before → after) |
|---|---|---|
| HELIOSPRING (IMUF9001) | 386799 → 386783 (**-16 B**) | 112752 → 112752 (unchanged) |
| FOXEERF722V4 (non-IMUF9001) | 386610 → 386610 (unchanged) | 96604 → 96604 (unchanged) |

The filter-gate commit alone measured -24 B; the Option 1 zeroing commit (6 forced-zero assignments in `validateAndFixGyroConfig()`) cost back 8 B, netting -16 B. FOXEERF722V4 is byte-identical across every commit in this PR — confirmed zero effect on non-IMUF9001 targets. (Earlier comments/PR-body revisions in this thread cite -24 B — that was measured before the Option 1 commit existed; -16 B is the correct final number.)

**#1281 (32kHz blocky/slow/saturation) — not related.** #1281's root cause is a blocking ~15-20µs SPI transfer in `imufSpiGyroRead()` (`accgyro_imuf9001.c:208-214`), consuming ~50-64% of the GYROPID task's CPU budget at 32kHz. This PR removes at most a couple of function-pointer-indirect calls per axis per loop — roughly two orders of magnitude smaller. #1281's own fix path (non-blocking DMA) is untouched by this PR.

**Hardware verification: done.** Real HELIOSPRING, user-confirmed 2026-09-03 — after flashing, EmuConfigurator save + refresh shows both gyro-lowpass sliders as OFF/unusable, as expected. No Configurator-side code change was needed (it already renders a 0 value as OFF).

**CI status** as of this comment: Codacy pass, CodeRabbit pass, sitl pass, both `test` jobs pass, 11/12 `build` target-groups pass, 1 (`targets-group-10`) still pending.

**Deliberately deferred, not blocking this PR**: Option 2 (removing `gyro_lowpass_hz*`/`gyro_lowpass2_hz*` from CLI/MSP entirely for IMUF9001, rather than just forcing them to 0) — tracked against #1250, needs that MSP freeze lifted and EmuConfigurator source access to verify current gating behavior first. No branch/work started on it.

**Verified, no action needed**: all 3 IMUF9001 targets (HELIOSPRING/STRIXF10/MODE2FLUX) need zero per-target `target.h`/`config.c` changes — the shared `USE_GYRO_IMUF9001` macro gate in `fc/config.c` covers all three; none of their `targetConfiguration()` functions touch the affected fields, and call-order (`targetConfiguration()` only on EEPROM reset vs. `validateAndFixGyroConfig()` on every boot) can't create a conflict either way.

nerdCopter:

Analysis requested on the newest commit (699cc545b), which fully gates the `gyroSensor_t` lowpass filter storage (`lowpassFilterApplyFn`/`lowpassFilter[]`/`lowpass2FilterApplyFn`/`lowpass2Filter[]`) and `gyroInitLowpassFilterLpf()` (declaration, definition, both call sites) behind `#ifndef USE_GYRO_IMUF9001`, on top of the earlier filter-apply-call gate:

1. Confirm no other file references the now-conditionally-compiled struct fields or function for an `USE_GYRO_IMUF9001` build (I grepped the full repo and found none outside `gyro.c`/`gyro_filter_impl.h`, but a second pass is worth having).
2. Confirm `gyroInitLowpassFilterLpf()`'s forward declaration and definition guards are consistent (no unused-function warning risk, no ODR issue) across all 3 build combinations (IMUF9001 defined; IMUF9001 undefined + LPF2 defined; IMUF9001 undefined + LPF2 undefined).
3. Flag anything else structurally risky about removing this struct storage/init entirely rather than just the runtime call.

nerdCopter:

Addressed both non-blocking observations from the analysis above, commit eadea8d54:
- gyroLowpassFilter_t: gated under #ifndef USE_GYRO_IMUF9001, matching the struct fields that use it.
- gyroInitLowpassFilterLpf(): added static to the definition, matching the forward declaration.

No functional or size change.

nerdCopter:
AI Generated comment

Flight-tested 2026-09-05 via a mashup of this branch (commit `699cc545b`+`eadea8d54`) with PR #1330 (`build/size-optimised-src-expand`), commit `6201ba2ad`. Two long flights each on HELIOSPRING (8k/4k) and FOXEERF722V4 (8k/8k). All 4 logs analyzed: no motor oscillation, smooth stick input, normal step response on every flight. Firmware revision `6201ba2` confirmed on all 4 logs.

HELIOSPRING's blackbox reports show `LPF1: —, LPF2: —` on every axis — the host-side lowpass fields this PR forces to 0 are recorded as empty/absent in real flight data, not just a CLI dump or Configurator UI state. `IMUF_lowpass_roll/pitch/yaw=90` in the header is the onboard-coprocessor filter, a separate field untouched by this PR.

FOXEERF722V4 still shows its normal PT1 lowpass filters active (120/120/100 Hz) in both logs — confirms this PR's zero-effect result on non-IMUF9001 targets in real flight, not just the earlier static build-size comparison.

PR #1433, opened 2026-09-03
feat(target): enable USE_ESCSERIAL fleet-wide via common_fc_pre.h

### LINKED ISSUE #1430: target: enable USE_ESCSERIAL across all supported boards via common_pre/post
AI Generated issue-ticket

Found while implementing IT #1426/PR #1429 (`serial_escserial.c` TX-pin fallback fix). Scope here is target-config plumbing, not a code change — no logic in `drivers/serial_escserial.c` needs to change for this.

**Current state.** 228 of 465 EF `target.h` files define `USE_ESCSERIAL` (49%). This is not driven by flash budget: 31 of 59 F411 targets (the smallest common flash class in the fleet) already enable it, while several H743/H725 targets with abundant flash (`AIRBRAINH743`, `AXISFLYINGH743PRO`, `BETAFPVH725`, `BETAFPVH743`) do not. The feature has no unusual peripheral dependency — any board with motor outputs has the pins/timers `USE_ESCSERIAL` needs. The inconsistency looks like incomplete/unaudited per-target headers, not a deliberate exclusion.

**BF comparison.** BF gates `USE_ESCSERIAL` per MCU-family header (~20 shared files under `src/platform/<family>/target/<MCU>/target.h`, e.g. `STM32F405`, `STM32H743`), not per individual board. Since BF's ~629 real board configs inherit from these shared family headers rather than redefining the flag, `USE_ESCSERIAL` is effectively fleet-wide across every family BF still supports.

**Motivation.** `PROTOCOL_KISS` (used by open/KISS-protocol ESCs) only configures the TX side of `openEscSerial()` — it never touches the RX side of the shared timer channel, so it already works correctly on any board with `USE_ESCSERIAL` enabled, with or without a dedicated `ESCSERIAL_TIMER_TX_PIN`. (The TX-pin-fallback bug fixed in PR #1429 only affects the legacy SIMONK/BLHELI/CASTLE protocols, not KISS.) Enabling `USE_ESCSERIAL` fleet-wide would deliver working open/KISS-ESC compatibility on every board immediately, with no further code changes needed beyond PR #1429.

**Scope (corrected — see comment thread below).** EF has a direct equivalent of BF's shared-header mechanism: `src/main/target/common_fc_pre.h`, included before every `target.h` (`src/main/platform.h` processes `common_fc_pre.h` → `target.h` → `common_fc_post.h`, confirmed by CodeRabbit analysis). The correct technique is one `#define USE_ESCSERIAL` added there, not 237 individual per-target additions:

1. Add `#define USE_ESCSERIAL` once to `common_fc_pre.h`.
2. Remove the now-redundant per-target `#define USE_ESCSERIAL` lines from the 228 `target.h` files that already had them.
3. For boards that cannot safely take this on (flash budget, conflicting peripheral use), add an explicit `#undef USE_ESCSERIAL` in that board's own `target.h`, with a documented reason — verified case by case, not left silently unaudited.

The original scope here proposed the opposite of this — per-target `#define` additions across ~237 individual `target.h` files — which directly contradicts this issue's own BF-comparison paragraph above (BF avoids exactly that fragmentation via shared family headers). Caught and corrected before any implementation started; no target files were edited under the original scope.

**Relationship to IT #1395 (fleet-wide `USE_TIMER_MGMT`).** Independent, not blocked either direction. This issue's enablement doesn't add exposure to IT #1426's motor-timer ownership-conflict bug — that bug is only reachable where a real ownership check exists (currently `NERO`/`STM32F7X2` only), and enabling `USE_ESCSERIAL` elsewhere doesn't change that. If IT #1395 eventually lands fleet-wide, boards enabled here would start surfacing that already-documented, cleanly-failing limitation more widely — an acceptable future consequence, not a reason to sequence this behind it.

Related: #1426(escserial-defects), #1429(TX-pin-fix), #1144(target-unify), #1395(USE_TIMER_MGMT).

nerdCopter:
AI Generated pull-request

## Summary
`USE_ESCSERIAL` was defined per-board in 228 of 465 target.h files, inconsistently and unrelated to flash budget (31/59 F411 boards already had it while several H743/H725 boards with more flash did not). Adds a single `#define USE_ESCSERIAL` to `common_fc_pre.h` (fleet-wide default, mirrors how BF gates the same feature per shared MCU-family header rather than per board) and removes the now-redundant per-target defines from those 228 files.

`SITL` excludes `drivers/serial_escserial.c` from its build (`make/mcu/SITL.mk` `MCU_EXCLUDES` — no real timer/serial hardware to pass through), but `cli.c`/`msp.c` call the driver's `escEnablePassthrough()` under the same `USE_ESCSERIAL` guard. The fleet-wide define broke the SITL link (`undefined reference to escEnablePassthrough`); `SITL/target.h` now explicitly `#undef`s it back off, with a one-line reason. No other MCU family excludes the driver file.

No logic change in `drivers/serial_escserial.c` — this only extends which boards compile in the already-existing, already-reviewed driver. `PROTOCOL_KISS`/`KISSALL` ESCs only use the TX side and work correctly with or without a dedicated TX pin; the SIMONK/BLHELI/CASTLE TX-pin fallback path is covered by the fix in PR #1429 (merged).

Measured cost (clean build, isolated from any newly-enabled board): ~4.2–4.65 KB flash (text) and ~2.25 KB RAM (bss) — the driver's two 1024-byte TX/RX buffers plus code (TUNERCF405 +4197/+2256 B, FOXEERF722V4 +4454/+2276 B, SPRACINGH7EF +4646/+2264 B; measured with both build sides on a clean, committed tree — EmuFlight's Makefile embeds a dirty-tree REVISION string of different length that otherwise skews .text by a constant +32 B). Checked the tightest-flash board in the fleet, `SPRACINGH7EF` (128K internal flash, EXST/external-OctoSPI): usage goes from 35.83% to ~36.3% of its 1 MB external-flash region, no risk of overflow. `FOXEERF722V4` (real aircraft) internal flash goes from 79.17% to ~80.1%.

CodeRabbit analysis found 6 boards (`AOCODAF722MINI`, `GEPRCF722BT`, `GEPRC_F722_AIO`, `JHEF405PRO`, `TMOTORF7V2`, `TMVELOX`) where a `//PPM` comment — documenting that `ESCSERIAL_TIMER_TX_PIN` shares a timer/pin with PPM input — was attached to the deleted `USE_ESCSERIAL` line and got dropped along with it. Fixed by relocating the comment onto the `ESCSERIAL_TIMER_TX_PIN` line it actually documents; no behavior change.

Closes #1430

## Test plan
- [x] `make clean_test && make test` — 49/49 unit test binaries pass, 0 failures
- [x] Bench matrix: HELIOSPRING TUNERCF405 SKYSTARSF405AIO PYRODRONEF7 FOXEERF722V4 FOXEERF405 APEXF7 TMOTORF7 SITL — 9/9 succeed
- [x] Compile-only matrix: MATEKF411 NBDHMBF4PRO WORMFC ALIENWHOOPF7 CRAZYFLIE2 — 5/5 succeed
- [x] Rebuild of the 6 `//PPM`-comment-fix boards — 6/6 succeed
- [x] Full CI build-all-targets matrix (all ~465 targets) — SUCCESS, all 11 build groups + rest green, no other board needed an exclusion
- [x] Hardware verification — TUNERCF405 (F405, newly enabled by this PR, no dedicated ESCSERIAL_TIMER_TX_PIN): Bluejay ESC flash via 4-way interface succeeds (no motor/timer regression); `escprog bl 0` cleanly refuses per PR #1429's existing guard, `status` clean post-attempt


nerdCopter:

Analysis requested on 3 specific points:

1. Confirm `common_fc_pre.h` is the correct fleet-wide location (included via `platform.h` before every `target.h`) and that no other MCU-family header also touches `USE_ESCSERIAL`.
2. Confirm the SITL `#undef` in `SITL/target.h` is the only exclusion needed — check whether any other `make/mcu/*.mk` `MCU_EXCLUDES` list omits `drivers/serial_escserial.c` while its callers (`cli.c`/`msp.c`) remain compiled in.
3. Confirm none of the 228 removed per-target `#define USE_ESCSERIAL` lines carried a board-specific comment/config (e.g. a non-default `ESCSERIAL_TIMER_TX_PIN`) that got silently dropped along with the define.

nerdCopter:

Addressed finding from the analysis above:
- //PPM comment loss on 6 boards (AOCODAF722MINI, GEPRCF722BT, GEPRC_F722_AIO, JHEF405PRO, TMOTORF7V2, TMVELOX): fixed in commit 03449ff86 — relocated the comment from the removed USE_ESCSERIAL line onto the ESCSERIAL_TIMER_TX_PIN line it actually documents (shared timer/pin with PPM input), preserving the resource-relationship note.

224-vs-228 count and SITL-exclusion confirmations: no action needed, both consistent with the PR description (228 = 224 active + 4 already-commented-out USE_ESCSERIAL lines removed as dead cruft).

nerdCopter:
AI Generated comment

Flash/RAM cost analysis — no action needed, recorded here so it doesn't need re-deriving at merge time.

**Measured cost per newly-enabled board**: TUNERCF405 +4197 B flash / +2256 B RAM, FOXEERF722V4 +4454 B / +2276 B, SPRACINGH7EF +4646 B / +2264 B. Both build sides were on a clean, fully-committed tree (dirty-tree builds embed a longer `REVISION` string that otherwise skews `.text` by a constant +32 B — verified and excluded).

**Conclusion: not a problem, no follow-up optimization ticket warranted.**
1. Scale: 4.2–4.65 KB flash and 2.25 KB RAM is under 1-4% of flash and under 2% of RAM on every board checked, including the fleet's tightest-flash target (`SPRACINGH7EF`, 128K internal flash, but EXST-mode with a 1 MB external-flash region — usage goes 35.83%→~36.3%).
2. The buffer size is not an EF inefficiency: BF 4.5-maintenance's own `serial_escserial.h`/`.c` use the identical `ESCSERIAL_BUFFER_SIZE 1024` / `MAX_ESCSERIAL_PORTS 1` — a long-standing upstream value, not something EF added or oversized. Shrinking it would increase the EF/BF delta rather than reduce it (this project's stated evaluation criterion), for a cost that isn't actually a problem, and risks breaking the ESC-flashing protocol if 1024 B is sized to a real packet/page requirement rather than arbitrary slop.

**Remaining gate before merge**: the full CI `build.yml` matrix (~465 targets) is the authoritative check for any board-specific flash overflow this analysis doesn't cover — still running as of this comment. If it fails on a specific board, the correct fix is a documented per-board `#undef USE_ESCSERIAL` in that board's `target.h` (the same mechanism already used for `SITL`), not a driver-level buffer change.

nerdCopter:
AI Generated comment

Real-hardware confirmation from user, TUNERCF405 (F405) @ commit 03449ff86:

```
EmuFlight / TUNERCF405 (S405) 0.4.3 Sep  4 2026 / 07:37:34 (03449ff) MSP API: 1.54
```

1. Bluejay ESC firmware successfully flashed to the onboard 8-bit ESCs — this path goes through the 4-way interface (`USE_SERIAL_4WAY_BLHELI_BOOTLOADER`, fleet default, unrelated to this PR's driver), confirming no motor-pin/timer regression from enabling `USE_ESCSERIAL` on this board.
2. `escprog bl 0` (CLI passthrough, `PROTOCOL_BLHELI`) returned a clean, immediate `###ERROR### Error starting ESC connection` — expected: TUNERCF405 has no `ESCSERIAL_TIMER_TX_PIN` defined, so this hits PR #1429's already-merged guard (refuse before touching hardware for SIMONK/BLHELI/CASTLE when no dedicated TX pin exists, rather than hang). Matches the same signal confirmed on HELIOSPRING/STM32F7X2/STELLARH7DEV during PR #1429's own hardware verification.
3. `status` output clean post-attempt — normal uptime, no crash, arming-disable flags as expected for a bench setup, resource state uncorrupted.

This is the first hardware confirmation of this PR's change on a board that was previously `USE_ESCSERIAL`-OFF and has no dedicated `ESCSERIAL_TIMER_TX_PIN` (unlike the boards PR #1429 was hardware-verified on, which all had `USE_ESCSERIAL` already enabled). No new findings — behaves exactly as designed.

nerdCopter:
AI Generated comment

Full CI build-all-targets matrix completed: SUCCESS. All 11 build groups + targets-group-rest (the entire ~465-target fleet), both `test` jobs, `sitl`, Codacy, and CodeRabbit all green. No other board needed a `USE_ESCSERIAL` exclusion beyond the already-documented `SITL` one — confirms this PR's local sampling (14 targets) correctly predicted the full-fleet result.

Combined with the TUNERCF405 real-hardware confirmation above, this PR has no known outstanding findings. Staying DRAFT per standing instruction — ready for you to review at your own pace.

Issue #1455, opened 2026-09-17
target: ELLE0 ADC1/TIM8-CH3 DMA-stream collision at DMA2_Stream4, live by default


nerdCopter:
AI Generated issue-ticket

## Board affected

ELLE0 (STM32F405).

## Root cause

`target.c`'s `timerHardware[]` sets TIM8 CH3 (PC8, motor 3) to `dmaopt 1`, resolving to
DMA2_Stream4 (per its own comment: `// SERVO3 DMA2_ST4`). `target.h` has no `ADC_INSTANCE`
or `ADC1_DMA_STREAM` override, so ADC1 defaults to `DMA2_Stream4` too (`drivers/adc.h`'s
`#ifndef ADC1_DMA_STREAM` default). Both resolve to the same physical stream.

## Failure mode

`adcInit()` claims its configured stream via a single-shot `dmaAllocate()` call with no
retry/fallback. ADC init runs before motor/PWM init, so ADC1 wins the stream first; the
motor driver's later `dmaAllocate()` call for TIM8 CH3 then fails silently, and motor 3 gets
no DShot DMA configured.

## Reachability

Live, not latent. This target does not define `USE_DSHOT_DMAR`, so burst DShot (which uses a
separate, non-colliding DMA claim path) is compiled out entirely for this board -- there is
no pilot setting that avoids this collision.

## Fix

No `USE_SPI_DEVICE_1` on this target (single gyro on SPI2 only), so ADC1's other stream
option (Stream0) is free -- confirmed against this board's full DMA2 usage (TIM8
CH1/CH2/CH4, no SPI1, no UART DMA opt-in, no LED_STRIP/SDCARD/transponder). Add an explicit
override:

    +#define ADC1_DMA_STREAM DMA2_Stream0

No `ADC_INSTANCE` change needed (stays on ADC1).

nerdCopter:
AI Generated comment

Duplicate of IT #1452 (same ADC1/TIM8-CH3 DMA-stream collision, different board). Consolidated tracking into IT #1452, fix merged into PR #1454.

Issue #1210, opened 2026-05-28
fix(h7/usb): USB MSC not mounting on Linux with W25M02G NAND flash active


nerdCopter:
AI Generated [issue-ticket]

> [!IMPORTANT]
> **Do not close this issue until USB MSC mode is explicitly tested and passing on current master with STELLARH7DEV + W25M02G NAND.** Successful BBL recording and MSP download do NOT satisfy this requirement — they use a different code path (USB CDC serial / MSP protocol).

> [!CAUTION]
> **TEST FAILED 2026-06-02 on `cb9bbfa0c4` (master).** Configurator showed: *"Error: This operation is not supported by your hardware."* Root cause: `mscCheckFilesystemReady()` returned false at `msp.c:1473` → MSP reboot response byte = 0 → Configurator error. Investigation and firmware fix required.

---

## Summary

On STELLARH7DEV (W25M02G NAND, SPI4), USB mass storage does not mount on Linux. The device enumerates as USB MSC but the SCSI layer fails. Configurator returns "not supported by your hardware" before even attempting to reboot into MSC mode.

---

## Prerequisites — RESOLVED

| Prerequisite | Status |
|---|---|
| #1209 — NAND erase blocking scheduler | ✅ FIXED in `37c07ca880` (PR #1231, merged 2026-06-02) |
| BBL recording to NAND (W25M02G, SPI4) | ✅ Verified — 4.3 MB / 87s, gap-free |
| BBL download via Configurator (MSP path) | ✅ Verified — data intact |
| NAND erase via Configurator | ✅ Verified — completes asynchronously, no FC disconnect |

---

## MSC Test Result — FAILED 2026-06-02

| Test | Status |
|---|---|
| Configurator → USB Mass Storage button on STELLARH7DEV + W25M02G | ❌ "This operation is not supported by your hardware" |

**Failure path:** `MSP_REBOOT` with `rebootMode = MSP_REBOOT_MSC` → `mscCheckFilesystemReady()` returned `false` → MSP wrote `0` to response → Configurator showed error.

**`mscCheckFilesystemReady()` logic** (`src/main/io/usb_msc.c`):
```c
return (blackboxConfig()->device == BLACKBOX_DEVICE_FLASH && flashfsGetSize() > 0);
```

Both conditions should be true given working BBL recording. Investigation needed into why this returns false at MSC request time — timing, NAND-specific init sequence, or BF parity gap in H7 MSC code path.

---

[...]

PR #1456, opened 2026-09-17
fix(target): resolve ADC1/TIM8-CH3 DMA-stream collision on ELLE0

nerdCopter:
AI Generated pull-request

## Summary

Fixes IT #1455: TIM8 CH3 (motor 3, PC8, `dmaopt 1` -> DMA2_Stream4) and ADC1 (no override,
defaults to DMA2_Stream4 per `adc.h`) collide on `ELLE0`. `adcInit()` claims the stream first
at boot, so the motor's later DMA claim silently fails -- no DShot output on that motor, no
error surfaced.

Live, not latent: this target does not define `USE_DSHOT_DMAR`, so burst DShot's separate,
non-colliding DMA claim path is compiled out entirely. There is no pilot setting that avoids
this collision on this board.

## Fix

No `USE_SPI_DEVICE_1` on this target (single gyro on SPI2 only), so ADC1's other stream
option (Stream0) is free -- checked against this board's full DMA2 usage (TIM8
CH1/CH2/CH4, no SPI1, no UART DMA opt-in, no LED_STRIP/SDCARD/transponder). An explicit
`ADC1_DMA_STREAM` override is sufficient; no `ADC_INSTANCE` change needed, unlike other
boards in this bug family that do have an SPI1 gyro bus contending for the same stream.

## Test plan

- [x] `ELLE0` builds clean with `EXTRA_FLAGS=-Werror`
- [x] `SITL` builds clean
- [x] `make test`: 49/49 unit test binaries pass
- [ ] No owned hardware for this board -- compile/CLI-level verification only

Closes IT #1455.


nerdCopter:
AI Generated comment

Consolidated into PR #1454 (same commit, cherry-picked onto that branch) to match how the equivalent betaflight/config fix (PR #1210) already bundles every affected board in one PR, instead of splitting by board.

Issue #1395, opened 2026-08-18
feat: timerAllocate() ownership tracking — BF fleet-wide default via platform_mcu.h + config submodule, EF has neither


nerdCopter:
AI Generated issue-ticket

## Summary

EF has no equivalent of BF's `timerAllocate()`/`timerOwners[]`/`timerGetOwner()` ownership-tracking mechanism for timer resources. Every EF driver claims a timer via `timerGetByTag(ioTag)`, which always returns the timer's hardware descriptor with no conflict check at all.

## Second correction to this issue's scope (first correction understated the gap)

This issue was edited once already this session to note that BF's `timerAllocate()` ownership check is gated behind `USE_TIMER_MGMT`, and concluded the real check was rare (2 of EF's 465 targets have the equivalent gate). That conclusion was wrong — it read `timer_common.c` in isolation without checking BF's platform-default headers or its real per-board fleet (the `betaflight/config` submodule, ~629 boards, not `src/main/target/`, which only holds shared MCU-family headers post-unified-target).

Corrected finding:

- `drivers/stm32/platform_mcu.h` unconditionally `#define`s `USE_TIMER_MGMT` for STM32/AT32 (line 253), gated off only if `USE_DMA_SPEC` is undefined. `USE_DMA_SPEC` is itself `#define`d for every STM32 family BF supports — F4, F7, H7, G4 (lines 126/148/166/184) — so `USE_TIMER_MGMT` is never undef'd on any real STM32 board.
- The runtime pinmap `timerAllocate()` walks (`timerIOConfig()`, populated by `pgResetFn_timerIOConfig()`) is pre-filled at boot from a per-board `TIMER_PIN_MAPPING` macro, not left empty pending manual CLI configuration. Checked the config submodule directly: 616 of 629 real board configs define `TIMER_PIN_MAPPING`.
- Net result: BF's real ownership check is fleet-wide default on BF's actual shipped fleet, not a 2-board curiosity. The original filing's framing (a genuine, fleet-wide flight-safety gap) was closer to correct than this issue's first edit gave it credit for.

[...]

PR #1459, opened 2026-09-18
feat(drivers): wire fullTimerHardware[]/TIMER_PIN_MAPPING into the timer resolution layer

nerdCopter:
AI Generated pull-request

## Summary

IT #1395 Stage 2: connects the `fullTimerHardware[]`/`TIMER_PIN_MAPPING` architecture ported in prior work (previously dead code — zero consumers, zero boards with pin-mapping data) to the driver layer that actually resolves timer hardware.

- Adds `TIMER_HARDWARE`/`TIMER_CHANNEL_COUNT` indirection macros to `timer.h`, gated on `USE_TIMER_MGMT` (transitional seam over today's per-board `timerHardware[]`/`USABLE_TIMER_CHANNEL_COUNT` path — unconverted boards are unaffected).
- Ports a resolver split into `timer_common.c` matching the reference architecture: strict occurrence-index resolution under `USE_TIMER_MGMT`, first-match legacy behavior otherwise.
- Rewires the remaining driver-layer consumers (RCC clock-enable, IRQ dispatch, channel-info sizing, `timer <pin> list` CLI enumeration) in `timer.c`/`timer_hal.c`/`cli.c`.
- Fixes a latent pointer-arithmetic bug this rewire exposed (`timerChInit()` computing an array index against the wrong array once resolution can return a pointer into a different table) — currently dead code (zero call sites), fixed to match the reference implementation regardless.
- Gives 4 boards (`FOXEERF722V4`, `HELIOSPRING`, `NERO`, `STELLARH7DEV`) real `TIMER_PIN_MAPPING` data, derived mechanically and cross-validated against real reference config data where available. `STM32F7X2` is deliberately excluded — its runtime-configurable pin count exceeds the fixed-size pin-mapping array's current cap, with no reference architecture to converge toward; tracked as follow-up work.

Local review (CodeRabbit + opencode) caught and fixed 2 real occurrence-index bugs on the two boards without independent ground-truth validation before this PR opened; both confirmed correct after the fix, re-verified against real hardware CLI output.

Class A (usageFlags-driven pin discovery) is deliberately untouched — needs fleet-wide named-pin-define conversion first, deferred to a later stage.

## Test plan

- [x] `make clean_test && make test`: 49/49 unit test binaries pass
- [x] Build clean (0 warnings): `HELIOSPRING`, `FOXEERF722V4`, `STELLARH7DEV`, `NERO`, `STM32F7X2`, plus a legacy-board spot-check for fleet-wide regression
- [x] Real hardware: `timer list` CLI output confirmed exact match against committed pin-mapping data on `HELIOSPRING` and `FOXEERF722V4`
- [ ] Real hardware: pre-branch baseline `dma`/`resource` capture for a true byte-identical diff (post-branch only captured so far)
- [ ] Real hardware: `STELLARH7DEV` CLI verification
- [x] Real hardware: props-off motor spin test on all 3 boards
- [ ] Full fleet CI (this PR's own check run)

## Summary by CodeRabbit

* **New Features**
  * Added timer management with per-pin timer mappings for supported targets.
  * DShot motor DMA now uses the board-configured DMA request and matching interrupt stream.
  * Added timer CLI examples for listing all timers or timers assigned to a specific pin.
  * Added indexed timer lookup support for configured timer mappings.

* **Bug Fixes**
  * Improved timer and DMA selection to respect board-specific configurations.


nerdCopter:

Analysis requested on this diff (IT #1395 Stage 2, timer resolution wireup):

1. `timer_common.c`'s `timerGetByTagAndIndex()`/`timerGetByTag()` split — confirm the `USE_TIMER_MGMT` branch's strict `!timerIndex` -> NULL semantics is correctly scoped to only that branch, and the legacy `#else` branch preserves first-match behavior byte-for-byte for boards without `USE_TIMER_MGMT`.
2. `timer.c`/`timer_hal.c`'s `timerChInit()` pointer-arithmetic fix (`channel = timHw - TIMER_HARDWARE`) — confirm this is dead code today (zero call sites) as claimed, and that the fix is still correct against the array it now resolves through.
3. Per-board `TIMER_PIN_MAPPING` occurrence values on `FOXEERF722V4`, `HELIOSPRING`, `NERO`, `STELLARH7DEV` — independently recompute occurrence indices against each family's `fullTimerHardware[]` table and flag any mismatch. Note: complementary-channel entries (`CHxN`) count as real occurrences in the runtime resolution loop; a prior derivation bug that dropped them from counting was caught and fixed pre-PR (commit history has the detail).
4. `STM32F7X2`'s reverted `USE_TIMER_MGMT` + newly-added per-board `USED_TIMERS` — confirm this correctly restores pre-existing legacy-path behavior with no gap.

nerdCopter:

Analysis requested on the latest commits (`076597831`, `6842e68d6`):

1. `dma_reqmap_mcu.c`: `dmaoptByTag()` (H7 and F4/F7 blocks) now reads the board's real `TIMER_PIN_MAPPING` dmaopt by ioTag, with a fallback to the prior reverse-derivation (`dmaGetOptionByTimer()`) when no pinmap entry covers a pin. Confirm the fallback correctly preserves byte-identical resolution for every board without `USE_TIMER_MGMT`, and that it doesn't mask a genuine misconfiguration on a converted board where a motor pin is missing from `TIMER_PIN_MAPPING`.
2. `pwm_output_dshot_hal.c` / `pwm_output_dshot.c`: `dmaIrqHandler` is now re-derived via `dmaGetIdentifier()` on the resolved stream instead of the stale baked field. Confirm this is applied consistently at every use site in both files (`dmaAllocate`, `dmaEnable`, `dmaSetHandler`) and that burst-mode (`useBurstDshot`) code paths are correctly left untouched.
3. `cli.c`: the `timer` command's example/hint text now advertises `list`/`<pin> list` — confirm no other command in the same file has an equivalent working-but-undocumented subcommand introduced by this branch specifically (not pre-existing gaps in unrelated commands).

Real-hardware verification: all 3 converted boards with real ownership of the fix (HELIOSPRING, FOXEERF722V4, STELLARH7DEV) have been reflashed to the latest commit and hover-tested successfully, with `dma`/`resource` CLI output confirmed byte-identical or correctly changed as expected against pre-branch baselines.

nerdCopter:

Addressed:

1. **`dmaoptByTag()` fallback**: confirmed correct per your analysis — `timerAllocate()` gates motor setup before `pwmDshotMotorHardwareConfig()` runs, so a converted board's missing pinmap entry is caught upstream, not masked by this fallback.
2. **`dmaIrqIdentifier`**: confirmed correct in both files. The separate issue you found in `pwm_output_dshot.c` (resolved `dmaRef` never stored for the runtime write/IRQ paths, which kept reading the stale `timerHardware->dmaRef`) was real — fixed in commit `8b421012a`, matching the pattern `pwm_output_dshot_hal.c` already uses (`motor->dmaRef`).
3. **CLI hint text**: confirmed, no other undocumented subcommand introduced by this branch.

All 3 real-hardware boards with this fix (HELIOSPRING, FOXEERF722V4, STELLARH7DEV) were hover-tested against the prior commit; this new fix touches only `pwm_output_dshot.c` (F4/STDPERIPH), used by HELIOSPRING, which currently runs with `dshot_burst=ON` and so does not exercise the affected code path today — flagging for re-verification if burst is ever disabled on that board.

